PR #37, opened 2026-04-27
[AAASM-30] ✨ (aa-runtime/ipc): Implement Unix domain socket IPC server

Chisanan232:
## Summary

- Adds `ApprovalDecision` proto message to `proto/event.proto` (SDK→runtime approval resolution)
- Extends `RuntimeConfig` with `agent_id` (required, fail-fast) and `ipc_max_connections` (default 64); `from_env()` now returns `Result`
- Implements the full UDS IPC server in `aa-runtime/src/ipc/`: message types (`IpcFrame`, `IpcResponse`), length-framed binary codec (1-byte tag + prost varint + prost payload), and `IpcServer` with accept loop, semaphore connection limiting (max 64), per-connection reader/writer tasks, stale socket cleanup, and 0600 permissions
- Wires `IpcServer` into `runtime::run()` via `TaskTracker` + `CancellationToken`; bind failure is non-fatal

**Scope:** Transport-only. The server accepts connections, decodes frames onto an `mpsc::Sender<IpcFrame>`, and writes responses from a per-connection `mpsc::Receiver<IpcResponse>`. Gateway forwarding and response routing are out of scope and will be implemented in a follow-on ticket.

## Test Plan

- [ ] `cargo test -p aa-runtime -- --test-threads=1` — 24 tests pass, 1 ignored
- [ ] `cargo clippy -p aa-runtime --all-targets --all-features -- -D warnings` — clean
- [ ] `cargo fmt -p aa-runtime --check` — clean
- [ ] Integration tests in `ipc::server::tests` cover: Heartbeat, PolicyQuery, EventReport frame round-trips over a real UDS, and 5 concurrent connections
- [ ] Codec unit tests cover all 4 inbound frame types and all 3 outbound response types (encode + decode round-trip)
- [ ] Config tests cover `AA_AGENT_ID` required/blank/path-separator validation, `AA_IPC_MAX_CONNECTIONS` zero guard, and all default fallbacks

Closes #AAASM-30

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,7 @@ version = "0.0.1"
 dependencies = [
  "aa-core",
  "aa-proto",
+ "prost",
  "tokio",
  "tokio-util",
  "tracing",
--- a/aa-runtime/Cargo.toml
+++ b/aa-runtime/Cargo.toml
@@ -13,6 +13,7 @@ tokio                = { version = "1", features = ["full"] }
 tokio-util           = { version = "0.7", features = ["rt"] }
 tracing              = "0.1"
 tracing-subscriber   = { version = "0.3", features = ["json", "env-filter"] }
+prost                = "0.13"
 
 [lints]
 workspace = true
--- a/aa-runtime/src/config.rs
+++ b/aa-runtime/src/config.rs
@@ -50,6 +50,10 @@ impl RuntimeConfig {
             return Err("AA_AGENT_ID must not be blank or empty".to_string());
         }
 
+        if agent_id.contains('/') || agent_id.contains("..") {
+            return Err("AA_AGENT_ID must not contain path separators ('/' or '..')".to_string());
+        }
+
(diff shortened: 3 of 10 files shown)


Chisanan232:
## codecov/patch — known gap, intentionally left open

The `codecov/patch` check is failing but the overall `Coverage` check passes. This is expected and acceptable.

**What's not covered:** The uncovered new lines are async infrastructure paths — the `IpcServer::run` accept loop body, the socket bind/cleanup paths (which depend on real OS filesystem state), and the `main()` binary entrypoint. These are exercised by integration tests at runtime but may not be instrumented by the coverage tool in all configurations.

**Why we're not adding tests for them:** We have 24 tests covering the real behavior (codec round-trips, server integration, config validation, runtime lifecycle). Adding tests purely to satisfy the patch coverage metric would mean shimming OS signals or socket state — testing the framework, not the feature. That violates our policy of testing behavior, not implementation.

**Decision:** Merge as-is. All 11 required CI checks pass including `Coverage`, `Test`, `Build`, `Clippy`, and `Format`.

PR #1073, opened 2026-06-16
[AAASM-2975] 🔧 (release): Ship aa-gateway in the release tarball so aasm can launch a core

Chisanan232:
## Description

The GitHub Release shipped only the `aasm` binary (`aasm-<triple>.tar.gz`), but `aasm gateway start` spawns a **separate `aa-gateway` binary** via `Command::new(...)`. Because `aa-gateway` was never in the release artifact, a user installing from a Release or via Homebrew could not launch a core — `aasm gateway start` failed with `aa-gateway binary not found`.

Per the owner decision, this PR **packages `aa-gateway` in the release tarball** alongside `aasm` (no in-process serve mode). It also hardens `aasm`'s binary resolution so the packaged sibling is actually found.

### What changed

**`.github/workflows/release.yml`**
- Build step now compiles both binaries: `cargo build --release --target <triple> -p aa-cli -p aa-gateway` (was `-p aa-cli` only). `aa-gateway` is a normal host binary — no eBPF / cross-target quirks — so it builds for every matrix target.
- Added a lightweight `Verify aa-gateway binary built` step (size + `file` type; it is not executed to avoid binding a listener).
- Archive step now tars both binaries into the **same** `aasm-<triple>.tar.gz`: `tar -czf … -C target/<triple>/release aasm aa-gateway`. The tarball **name is unchanged**, so `SHA256SUMS`, cosign keyless signing, the Homebrew formula sha256 rewrite, and the GitHub Release upload all continue to work untouched.

**`aa-cli/src/commands/gateway/start.rs`**
- `resolve_binary()` now checks the directory of the running `aasm` executable (`std::env::current_exe()`) **first**, before `$PATH` → `~/.cargo/bin` → `./target/{release,debug}`. This is required because a tarball can be unpacked to an arbitrary directory that is not on `$PATH`; the new lookup finds the co-packaged `aa-gateway` regardless. New `sibling_binary()` helper + updated "not found" message.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-2975

## Testing

- [x] No tests required for the workflow change (explain why)

How verified:
- **release.yml** only runs on a tag push, so it was validated **statically**: `actionlint -shellcheck=` clean (avoids the known SC2016 false positive) and `python3 -c 'import yaml; yaml.safe_load(...)'` parses. Confirmed by reading the final archive + upload steps that **both** `aasm` and `aa-gateway` are built, verified, packaged, and uploaded for every matrix target, with the artifact name unchanged.
- **start.rs**: 3 new unit tests for `sibling_binary` (found alongside `aasm`, absent, present-but-not-executable). `cargo nextest run -p aa-cli` → 588 passed, 0 failed. `cargo fmt --all` clean, `cargo clippy -p aa-cli --all-targets -- -D warnings` exit 0.

How resolution now works: a Release/Homebrew install unpacks `aasm` and `aa-gateway` side-by-side; when the user runs `aasm gateway start`, `resolve_binary()` finds `aa-gateway` next to the running `aasm` executable (first lookup) and spawns it — no `$PATH` configuration required.

> Note (out of scope, follow-up): the dormant macOS codesign/notarize step signs only `aasm`. When Apple signing is enabled (`MACOS_SIGNING_ENABLED`), `aa-gateway` would also need signing. Left unchanged here to keep the fix scoped.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,8 +64,14 @@ jobs:
         run: |
           pnpm install --frozen-lockfile
           pnpm build
-      - name: Build aasm release binary
-        run: cargo build --release --target ${{ matrix.target }} -p aa-cli
+      - name: Build aasm + aa-gateway release binaries
+        # aasm is the operator CLI; aa-gateway is the core daemon that
+        # `aasm gateway start` spawns. Both must ship in the release tarball
+        # (AAASM-2975) so a release / Homebrew install can launch a core —
+        # without aa-gateway, `aasm gateway start` fails with
+        # "aa-gateway binary not found". aa-gateway is a normal host binary
+        # (no eBPF / cross-target quirks), so it builds for every matrix target.
+        run: cargo build --release --target ${{ matrix.target }} -p aa-cli -p aa-gateway
 
       - name: Install Rosetta 2 (x86_64-apple-darwin smoke test)
         if: matrix.target == 'x86_64-apple-darwin'
@@ -78,6 +84,17 @@ jobs:
           ls -lh "$BIN"
           file "$BIN"
           "$BIN" --version
+
+      - name: Verify aa-gateway binary built (size + type)
+        # aa-gateway is the daemon `aasm gateway start` spawns; it ships in the
+        # same tarball (AAASM-2975). We don't execute it here (it would bind a
+        # listener); a presence + Mach-O/ELF type check is enough — the archive
+        # step hard-fails anyway if the binary is missing.
+        shell: bash
+        run: |
+          BIN="target/${{ matrix.target }}/release/aa-gateway"
+          ls -lh "$BIN"
+          file "$BIN"
 
       - name: Codesign + notarize aasm (macOS — gated on Apple Developer creds)
         # Dormant until the operator sets the MACOS_SIGNING_ENABLED repo variable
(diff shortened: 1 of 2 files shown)


Chisanan232:
## 🔎 Code Review — Claude Code (record)

**Verdict: ✅ All CI green, scope complete — ready to approve & merge.**

### CI status — all green ✅
All 19 non-skipped checks pass: Build, Clippy, Format, Rustdoc, Rust conformance, Dependency checks (`cargo deny`), Compatibility matrix, no_std targets, Analyze (JS-TS + Python), **Test**, **Integration tests (macos-latest)**. No reds.

### Acceptance-criteria review (AAASM-2975 — "GitHub Release ships only aasm, not aa-gateway")
| Acceptance criterion | Done? | Evidence |
|---|---|---|
| A release-only install can run a working local gateway | ✅ | `release.yml` now builds `-p aa-cli -p aa-gateway` and tars **both** binaries into the same `aasm-<triple>.tar.gz`; `aasm gateway start` resolves `aa-gateway` next to the running exe |
| Chosen approach: package aa-gateway in the tarball | ✅ | Archive step `tar … -C target/<triple>/release aasm aa-gateway` (not an in-process serve mode, per the decision) |
| `aasm gateway start` finds the co-packaged binary | ✅ | `start.rs::resolve_binary()` checks `current_exe()`'s dir first (`sibling_binary` helper) before PATH / `~/.cargo/bin` / `./target` |
| No release-pipeline regressions | ✅ | tarball name unchanged → SHA256SUMS, cosign signing, Homebrew formula sha256 rewrite, and upload all untouched; added a presence/type verify step (doesn't execute the daemon) |
| Tests | ✅ | 3 unit tests (`sibling_binary` resolves / absent / not-executable); `nextest -p aa-cli` 588 passed |
| Static validation (release.yml runs on tag only) | ✅ | `actionlint -shellcheck=` clean, YAML parses |

### Scope confirmation — complete
Both the packaging (release.yml) and the runtime resolution (start.rs) are delivered, so the ticket's two outcomes hold: a release/Homebrew user can launch a core, and release-mode SDK⇄core tests can stand one up from published artifacts. 3 granular commits, base `master`, pushed to canonical `remote`, no hook bypass.

**Out-of-scope follow-up noted in the PR:** the dormant macOS codesign/notarize step signs only `aasm`; if Apple signing is ever enabled, `aa-gateway` would need signing too — left unchanged to keep the fix scoped.

Approving as ready to merge.

PR #1120, opened 2026-06-18
[AAASM-3353] 🐛 (aa-gateway): Accrue LLM spend + recover audit seq on restart

Chisanan232:
## Description

Two QA-round bug fixes that both live in `aa-gateway/src/service/policy_service.rs`, so they ship as one PR.

**D — AAASM-3353 (budget accrual):** The live `CheckAction` gRPC handler called only `engine.evaluate`, which *reads* accumulated spend at Stage 7 but never *records* it. `PolicyEngine::record_spend` / `BudgetTracker::check_and_decrement` had no caller outside tests, so daily/monthly budget limits never fired. This PR closes the loop: after a non-deny `LLM_CALL`, the call is priced from the model name and accrued via the existing `engine.record_spend` budget path (in both `check_action` and `batch_check`), so a later call is denied once the limit is exceeded.

**G — AAASM-3356 (audit seq recovery):** The service seeded `seq: AtomicU64::new(0)` and, on restart, recovered the hash-chain head (`read_last_hash`) but **not** the seq, so sequence numbers restarted at `0` and produced duplicate `seq` values in the WORM log. This PR adds `AuditWriter::read_last_seq`, a `PolicyServiceImpl::with_initial_seq` builder, and seeds the seq atomic from `last_persisted_seq + 1` at startup — mirroring the existing hash recovery.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

`setup_audit` (a private fn in `server.rs`) gained a fourth return value (`initial_seq`); no public API changed. `with_initial_seq` is an additive builder.

## Assumptions & scope notes

- **Provider is not in the proto.** Per spec, `proto/` was not changed. The `LlmCallContext` proto carries only the model name string (`model`), a `prompt_tokens` estimate, and `contains_pii` — no provider, no output tokens. The provider is therefore **inferred from the model name** via a new `Model::infer_from_name` helper (case-insensitive substring match, most-specific-first). NOTE: the spec said to put the helper "in aa-core", but the `Provider`/`Model` enums live in `aa-gateway::budget::types` (not aa-core), so the helper lives next to those enums. Moving the enums/helper into aa-core is a larger refactor and is left as a follow-up.
- **Output tokens are treated as 0** when pricing — the pre-execution check has no completion yet, so only input (prompt) tokens are charged. A post-execution true-up (e.g. via `ToolResult`) is a possible follow-up.
- **Unknown model → $0.00** (no spend accrued) rather than guessing a price.
- **Denied calls are not charged** (a hard `Deny` means the call did not run).
- **AuditService's independent seq.** This PR fixes only the `PolicyService` seq per the ticket scope. `AuditServiceImpl` keeps its own independent seq counter (also seeded from `initial_hash` only); that is a pre-existing, separate concern not regressed here and can be addressed in a follow-up if the two writers' sequence spaces need to be unified.

## Related Issues

- Closes AAASM-3353
- Closes AAASM-3356

## Testing

- `tests/budget_accrual_test.rs` — first `LLM_CALL` allowed and accrues spend; a second call is denied once accrued spend exceeds the daily limit; an unknown model accrues nothing.
- `tests/audit_seq_recovery_test.rs` — seq continues monotonically across a simulated restart (no duplicate of the prior run's seqs); a control case shows a fresh service without recovery duplicates `seq 0`.
- `tests/audit_test.rs` — `read_last_seq` returns the last entry's seq and `None` for a missing file.
- Full crate suite: `cargo nextest run -p aa-gateway` → **1121 passed, 0 failed**.
- `cargo fmt -p aa-gateway`, `cargo clippy -p aa-gateway --all-targets` clean.

QA evidence: see `agent-assembly-integration-tests/verification-reports/base-branch-2026-06/`.

## Checklist

- [x] Documentation updated if behaviour changed (doc comments on new APIs)
- [ ] All CI checks passing

### DIFF
--- a/aa-gateway/src/audit.rs
+++ b/aa-gateway/src/audit.rs
@@ -216,6 +216,42 @@ impl AuditWriter {
         }
         Ok(last_hash)
     }
+
+    /// Read the `seq` of the last entry in a JSONL file.
+    ///
+    /// AAASM-3356 — on restart the service recovers the hash chain head via
+    /// [`read_last_hash`](Self::read_last_hash) but previously re-seeded the
+    /// `seq` counter at `0`, producing duplicate sequence numbers after a
+    /// restart. Pairing this with `read_last_hash` lets the service seed the
+    /// `seq` atomic from `last_seq + 1` so sequence numbers stay monotonic and
+    /// unique across process restarts.
+    ///
+    /// Returns `None` if the file does not exist or is empty.
+    /// Skips blank or incomplete trailing lines (standard JSONL recovery).
+    pub async fn read_last_seq(path: &Path) -> io::Result<Option<u64>> {
+        let file = match tokio::fs::File::open(path).await {
+            Ok(f) => f,
+            Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(None),
+            Err(e) => return Err(e),
+        };
+        let reader = tokio::io::BufReader::new(file);
+        let mut lines = reader.lines();
+        let mut last_seq: Option<u64> = None;
+
+        while let Some(line) = lines.next_line().await? {
+            if line.trim().is_empty() {
+                continue;
+            }
+            match serde_json::from_str::<AuditEntry>(&line) {
+                Ok(entry) => last_seq = Some(entry.seq()),
+                Err(_) => {
+                    // Incomplete trailing line from a crash — skip it.
+                    continue;
+                }
+            }
+        }
+        Ok(last_seq)
+    }
 }
 
 /// Result of a hash-chain verification.
--- a/aa-gateway/src/budget/tracker.rs
+++ b/aa-gateway/src/budget/tracker.rs
@@ -343,6 +343,15 @@ impl BudgetTracker {
         self.monthly_limit_usd
     }
 
+    /// Borrow the tracker's [`PricingTable`].
+    ///
+    /// AAASM-3353 — lets the service layer price an LLM call from the same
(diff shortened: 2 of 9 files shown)


Chisanan232:
**Claude Code — PR review**

**CI:** ✅ all required checks green (21 success / 14 skipped). 1121 tests pass locally, incl. 6 new.

**Scope — covers two tickets (shared file `service/policy_service.rs`):**
- **AAASM-3353 (budget):** the live `CheckAction` path now accrues LLM-call cost (via a pricing helper + provider-inferred-from-model), so per-agent/team/daily/monthly limits actually fire — the QA repro (12× LLM_CALL all ALLOW) is resolved. Regression test asserts spend accrues and a later call is denied once the limit is exceeded.
- **AAASM-3356 (audit seq):** adds `AuditWriter::read_last_seq` and seeds the PolicyService seq from the last persisted entry on startup, so seq continues monotonically after restart (no duplicates). Tests cover recovery + post-restart monotonicity.

**Documented assumptions (reasonable):** provider inference lives in `aa-gateway::budget::types` rather than `aa-core` (relocation tracked as **AAASM-3362**); output tokens priced 0 at pre-execution; unknown model → $0 (no accrual); denied calls not charged; only PolicyService's seq is recovered — AuditService keeps an independent seq still to be fixed (**AAASM-3363**).

**Verdict:** ✅ Ready to approve & merge. Closes AAASM-3353 and AAASM-3356. (The two documented gaps are tracked as follow-ups, not blockers.)

PR #1147, opened 2026-06-19
[AAASM-3387] 🐛 (aa-sdk-client): Emit did:key agent identity for gateway registration

Chisanan232:
## Description

`AgentLifecycleService.Register` now **rejects** any `agent_id` that is not a
syntactically-valid `did:key` DID (`InvalidArgument: agent_id is not a did:key
DID (missing "did:key:" prefix)`). The language SDKs (python/node/go) and the
shared `aa-sdk-client` configure agents with plain human-readable identifiers,
so SDK-originated registration now fails against the live core — the root of the
broken example → live-core path (related to AAASM-3380).

This PR fixes the identity layer in the shared `aa-sdk-client` crate (which all
per-language FFI shims wrap), so the fix is inherited everywhere:

- Adds `aa_sdk_client::agent_id_to_did_key(identity)` — derives a deterministic,
  conformant Ed25519 `did:key` (`did:key:z6Mk…`, multicodec `0xed01` + base58btc)
  from a plain agent identifier. An identifier that is already a `did:key` is
  passed through unchanged.
- Adds `AssemblyConfig::registration_did()` so FFI shims get the registration DID
  directly; the socket-path / event-tag `agent_id` is intentionally left as-is.

### Why derive (and not require a keypair)

The task premise was to derive the DID from an Ed25519 public key the SDK already
generates. In this codebase **no SDK layer provisions an Ed25519 keypair** today
(`aa-sdk-client` and `aa-runtime` have none; the `public_key` field on
`RegisterRequest` is populated by the language SDKs, none of which generate a
key). The gateway's `validate_proto_agent_id` only checks the DID **syntax**
(`did:key:` prefix, `z` base58btc multibase, non-empty decoded bytes) — it does
**not** verify the multicodec key type or any signature. So the minimal,
self-contained fix is a deterministic SHA-256-derived Ed25519-form `did:key`,
which is byte-for-byte conformant and stable across restarts. When the SDKs grow
a real keypair, `agent_id_to_did_key` can be replaced with derivation from the
real public key without changing callers. No `proto/` change is required.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

(The socket-path / event-tag `agent_id` contract is unchanged; only the
registration identity is now a did:key.)

## Related Issues

- Related Jira ticket: AAASM-3387
- Closes AAASM-3387

## Testing

- [x] Unit tests added / updated — `aa-sdk-client` `identity` module tests
  (conformance, determinism, distinctness, canonical `z6Mk` prefix, multicodec
  payload, did:key pass-through). `cargo nextest run -p aa-sdk-client`: 34 passed.
- [x] Integration tests added / updated — **end-to-end proof**:
  `aa-gateway/tests/lifecycle_service_test.rs::register_with_sdk_derived_did_key_is_accepted`
  spins the real `AgentLifecycleService` gRPC server, registers with an agent_id
  produced by `aa_sdk_client::agent_id_to_did_key("my-agent-001")`, and asserts
  Register **accepts** it (returns a credential token, no `did:key` InvalidArgument).
  Full `lifecycle_service_test` suite: 19 passed.
- `cargo fmt` + `cargo clippy -p aa-sdk-client --all-targets`: clean.

### How to verify

```bash
export RUSTUP_TOOLCHAIN=stable
cargo nextest run -p aa-sdk-client
cargo test -p aa-gateway --test lifecycle_service_test register_with_sdk_derived_did_key_is_accepted
```

## Follow-ups (out of scope for this PR — shared-layer fix only)

The per-language SDKs must call `registration_did()` / `agent_id_to_did_key`
when building their `RegisterRequest.agent_id`:

- **python-sdk**: `GatewayClient.register_agent` posts to
  `/agents/{agent_id}/register` with the plain id — switch to the derived DID.
- **node-sdk** / **go-sdk**: thread the derived DID into their register event /
  `RegisterRequest`.
- **examples**: update any that hardcode plain agent ids.

These are tracked as separate SDK-component tickets.

## Checklist

- [x] Documentation updated if behaviour changed (module + fn docs added)
- [ ] All CI checks passing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,6 +257,7 @@ dependencies = [
  "aa-core",
  "aa-proto",
  "aa-runtime",
+ "aa-sdk-client",
  "aa-security",
  "aa-storage-postgres",
  "ahash 0.8.12",
@@ -461,7 +462,9 @@ version = "0.0.1-beta.2"
 dependencies = [
  "aa-proto",
  "aa-security",
+ "bs58",
  "prost",
+ "sha2 0.11.0",
  "tokio",
  "tracing",
 ]
--- a/aa-gateway/Cargo.toml
+++ b/aa-gateway/Cargo.toml
@@ -67,6 +67,10 @@ aa-storage-postgres = { path = "../aa-storage-postgres", optional = true }
 # driver's own embedded MIGRATOR so the gateway can never drift from the
 # canonical MVP schema (AAASM-2389).
 aa-storage-postgres = { path = "../aa-storage-postgres" }
+# End-to-end proof that an SDK-derived did:key identity is accepted by the
+# live AgentLifecycleService.Register RPC (AAASM-3387). Default features off
+# to avoid pulling the advisory-preflight (aa-security) dependency.
+aa-sdk-client = { path = "../aa-sdk-client", default-features = false }
 criterion    = { workspace = true, features = ["async_tokio"] }
 proptest     = { workspace = true }
 tempfile     = { workspace = true }
--- a/aa-gateway/tests/lifecycle_service_test.rs
+++ b/aa-gateway/tests/lifecycle_service_test.rs
@@ -221,6 +221,46 @@ async fn register_with_non_did_agent_id_returns_invalid_argument() {
     );
 }
 
+/// AAASM-3387 regression: a plain agent identifier run through the shared SDK
+/// client's `did:key` derivation is ACCEPTED by the live Register RPC. This is
+/// the end-to-end proof that SDK-originated registration succeeds against the
+/// gateway's did:key validation (the broken example→live-core path).
+#[tokio::test]
+async fn register_with_sdk_derived_did_key_is_accepted() {
+    let (addr, _registry) = start_server().await;
+    let mut client = AgentLifecycleServiceClient::connect(format!("http://{addr}"))
+        .await
+        .unwrap();
+
+    // A human-readable agent_id of the kind SDKs configure today — which the
+    // gateway rejects verbatim — converted via the shared SDK derivation.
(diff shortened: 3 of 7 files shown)


Chisanan232:
**Claude Code — PR review**

**CI:** ✅ all green. aa-sdk-client 34 / aa-gateway lifecycle 19 pass.

**Scope vs AAASM-3387 — FOUNDATIONAL (capability only; AC not yet fully met):** Adds `aa_sdk_client::agent_id_to_did_key()` (deterministic Ed25519 `did:key:z6Mk…`) + `AssemblyConfig::registration_did()`, and **proves the gateway Register accepts an SDK-derived did:key** (real-server E2E test). Solid building block, no proto change.

**But the investigation is the headline — the SDK live-core path is broken at multiple layers, so this PR alone does NOT make SDK/example registration work:**
- `aa-sdk-client` doesn't even perform gateway Register (it's a UDS client to aa-runtime; treats agent_id as a config string; no keypair).
- No SDK generates an Ed25519 keypair; all SDKs emit **plain** agent_ids.
- python SDK `GatewayClient` POSTs to **REST** `/agents/{id}/register` + `/agents/{id}/policy/check` — non-existent endpoints (the 3380 404s), contradicting ADR 0004 (should be `aa-sdk-client → gRPC`).

**Verdict:** ✅ Merge the capability — but **keep AAASM-3387 open** (its AC "SDK/example registration succeeds against current core" isn't met by this PR). The remaining work is per-SDK: call `registration_did()` AND migrate the SDK enforcement path to `aa-sdk-client → gRPC` per ADR 0004 — i.e. this folds into the **3021 → 3380** chain. Recommend per-SDK follow-up tickets.

PR #1162, opened 2026-06-19
[AAASM-3442] 🐛 (aa-cli): Resolve active policy for 'policy get' with no version

Chisanan232:
## Description

After a successful `aasm policy apply` (which reports `Active: true`) and with `aasm policy list` showing the active version, `aasm policy get` with no `--version` returned **"No policy versions found."**

Root cause: `apply` and `list` both talk to the **gateway** over HTTP (`/api/v1/policies`), where `active` is a server-side property. But `policy get` (no arg) read the **local** `FsHistoryStore` — a different source that `apply` never populates — so it was always empty for the typical apply-then-get flow.

Fix: `policy get` with no `--version` now resolves the currently active policy from the gateway via `GET /api/v1/policies/active` (the same source `apply`/`list` use) and prints its `policy_yaml`. This also makes the command's own doc comment ("Show the currently active policy") accurate. The `--version <id>` path continues to read the local history store, since the gateway exposes no by-version retrieval endpoint.

Design choice: **active-resolve** (preferred UX), not a required-arg error — the gateway already exposes the active policy, and `get.rs`'s documented intent was always "show the currently active (or a specific) policy".

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

`policy get` with no `--version` previously failed for the apply-then-get flow; it now succeeds. The CLI surface (flags) is unchanged.

## Related Issues

- Related Jira ticket: AAASM-3442

## Testing

`get.rs` tests now cover the `--version` (local-store) success and unknown-version failure paths against a real `FsHistoryStore`. The no-arg active path is an HTTP call to the gateway (covered by the existing API `get_active_policy` route). Validated with `cargo nextest run -p aa-cli commands::policy` (28 passed), `cargo clippy -p aa-cli --all-targets`, `cargo fmt`.

## Checklist

Closes AAASM-3442

### DIFF
--- a/aa-cli/src/commands/policy/get.rs
+++ b/aa-cli/src/commands/policy/get.rs
@@ -4,68 +4,69 @@ use std::process::ExitCode;
 
 use aa_gateway::policy::history::{FsHistoryStore, HistoryConfig, PolicyHistoryStore};
 use clap::Args;
+use serde::Deserialize;
+
+use crate::config::ResolvedContext;
 
 /// Arguments for `aasm policy get`.
 #[derive(Args)]
 pub struct GetArgs {
     /// Version identifier (SHA-256 prefix) to retrieve.
-    /// Shows the latest active policy when omitted.
+    /// Shows the currently active policy when omitted.
     #[arg(long)]
     pub version: Option<String>,
 }
 
-/// Execute the `aasm policy get` command.
-///
-/// When `--version` is provided, retrieves that specific policy version.
-/// Otherwise, retrieves the most recently applied (active) policy.
-pub fn run(args: GetArgs) -> ExitCode {
-    run_with_config(args, HistoryConfig::default_config())
+/// Response from `GET /api/v1/policies/active`. Only the YAML is consumed here.
+#[derive(Debug, Deserialize)]
+struct ActivePolicyResponse {
+    /// Raw YAML content of the currently active governance policy.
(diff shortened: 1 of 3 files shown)


Chisanan232 (commented):
## Claude Code — PR review (AAASM-3442) — now green

**✅ Ready to merge.** The earlier red was a real regression (now fixed), not a flake — good that it was caught before merge. All gates green: `Test`, `Integration tests (macos-latest)`, `CI Success` all SUCCESS.

### Fix
The behavior change (no-arg `policy get` resolves the ACTIVE policy from the gateway instead of the local history store) invalidated the integration test `policy_get_with_empty_data_dir_exits_failure`, which assumed no-arg reads local history. Updated it to the new contract (`policy_get_no_version_resolves_active_policy`: seed a policy → no-arg get succeeds + returns the active YAML); `policy_get_unknown_version_exits_failure` (`--version`, still reads local history) left as-is and passing. Validated against the real `aa-integration-tests cli_policy` suite this time (not just `-p aa-cli` unit tests).

### Design choice (sound)
Active-resolve over the gateway is the design-consistent fix — `get.rs`'s doc already promised "show the currently active policy," `apply`/`list` use the gateway, and `GET /api/v1/policies/active` exists. `--version <id>` still reads local history (no gateway by-version endpoint). Good to merge.

PR #1230, opened 2026-06-24
[AAASM-3702] 🐛 (docs): Null-guard mermaid-init theme-toggle listeners

Chisanan232:
## Description

The published mdBook site threw `TypeError: Cannot read properties of null (reading 'addEventListener')` on every page. The cause is the vendored `docs/mermaid-init.js` (referenced from `book.toml` `additional-js`), which iterates the theme ids (`ayu`/`navy`/`coal`/`light`/`rust`) and calls `document.getElementById(id).addEventListener('click', …)`. When the rendered theme picker does not contain every one of those ids, `getElementById` returns `null` and the chained `addEventListener` throws, aborting the rest of the inline script.

Fix: add optional chaining (`?.`) before `addEventListener` on both loop bodies so a missing theme-toggle element is skipped instead of throwing. Minimal, behaviour-preserving change.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-3702

## Testing

Validated `docs/mermaid-init.js` parses cleanly with `node --check`. The guard is a standard optional-chaining null check; when the element exists the listener is attached exactly as before.

## Checklist

### DIFF
--- a/docs/mermaid-init.js
+++ b/docs/mermaid-init.js
@@ -22,7 +22,7 @@
     // Simplest way to make mermaid re-render the diagrams in the new theme is via refreshing the page
 
     for (const darkTheme of darkThemes) {
-        document.getElementById(darkTheme).addEventListener('click', () => {
+        document.getElementById(darkTheme)?.addEventListener('click', () => {
             if (lastThemeWasLight) {
                 window.location.reload();
             }
@@ -30,7 +30,7 @@
     }
 
     for (const lightTheme of lightThemes) {
-        document.getElementById(lightTheme).addEventListener('click', () => {
+        document.getElementById(lightTheme)?.addEventListener('click', () => {
             if (!lastThemeWasLight) {
                 window.location.reload();
             }


Chisanan232:
## ✅ Claude Code review — AAASM-3702

**CI:** all green — `CodeQL` ✅, `Analyze (rust/go/python/javascript-typescript)` ✅, `Build mdBook` ✅.

**Scope vs ticket:** `docs/mermaid-init.js` (vendored, referenced by `book.toml` `additional-js`) — added `?.` before `addEventListener` in both theme-toggle loops, so a missing toggle element is skipped instead of throwing `TypeError: Cannot read properties of null (reading 'addEventListener')` on every page. Confirmed it's a real in-repo source file (not a generate-only artifact), so the guard lands at the right layer; `node --check docs/mermaid-init.js` passes. Minimal, targeted fix. AC fully met.

**Verdict:** ✅ Ready to approve & merge.

PR #1264, opened 2026-06-26
[AAASM-3787] 🐛 (aa-gateway): Include action args in decision cache key

Chisanan232:
## Description

The cascade decision cache (`aa-gateway/src/engine/cache.rs`) built its key for a
`ToolCall` from the agent id, policy epoch, and only the tool **name** — the action
**args** were discarded (`ToolCall { name, .. }`). The same gap applied to
`ToolResult` (keyed on `tool_name` only, dropping the response body).

Because stage-5 `requires_approval_if` predicates are evaluated over the payload
(e.g. `args.amount > 1000`, or response-side `tool_result.<key>`), a benign call's
cached `Allow` could be served — within the 60-second TTL — for a *different* call
that used the same tool name but a dangerous payload. That is an **approval bypass**:
a high-value action that should return `RequiresApproval` is silently allowed.

The fix hashes the full payload into `action_hash`:
- `ToolCall` now hashes `args`.
- `ToolResult` now hashes `result`.

So distinct payloads produce distinct cache keys and are re-evaluated through the
full stage-5 path instead of colliding on a cached verdict.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

Cache keys change shape, but the cache is an internal, derive-on-the-fly structure
with no persisted form and no public API surface — there is no compatibility concern.

## Related Issues

- Related Jira ticket: AAASM-3787
- Related GitHub issues: #N/A

Closes AAASM-3787

## Testing

Added regression tests:
- `cache.rs::different_tool_args_produce_different_keys` — same tool, different args → distinct keys.
- `cache.rs::different_tool_results_produce_different_keys` — same tool, different result bodies → distinct keys.
- `engine::tests::cached_allow_not_served_for_dangerous_args_through_cascade` — the end-to-end
  regression: a benign `transfer {amount:1}` (Allow, cached) followed within the TTL by
  `transfer {amount:1000000}` re-evaluates and yields `RequiresApproval` rather than the cached Allow.

Local validation (per-crate; the full-workspace `--all-features` pre-commit hook includes
the Linux-only `aa-ebpf` crate and cannot run on this macOS host — CI runs the full gate):

```
cargo build -p aa-gateway          # ok
cargo nextest run -p aa-gateway    # ok
cargo fmt --all -- --check         # ok
cargo clippy -p aa-gateway --all-targets -- -D warnings   # ok
```

## Checklist

- [x] All CI checks passing

### DIFF
--- a/aa-gateway/src/engine/cache.rs
+++ b/aa-gateway/src/engine/cache.rs
@@ -22,7 +22,11 @@ pub struct CacheKey {
     /// Policy epoch at the time of evaluation. Stale entries are invalidated
     /// when the epoch advances (via `load_policy` or `apply_yaml`).
     pub policy_epoch: u64,
-    /// FNV-1a hash of a canonical `"{action_kind}:{action_discriminant}"` string.
+    /// Hash of the action's kind tag **and** its full evaluated payload (tool
+    /// name + args, network URL + method, file path + mode, …). The payload is
+    /// part of the key because stage-5 `requires_approval_if` predicates evaluate
+    /// over it — a key that ignored args would serve a benign payload's cached
+    /// `Allow` for a dangerous one (AAASM-3787). See [`action_discriminant`].
     pub action_hash: u64,
 }
 
@@ -44,9 +48,14 @@ fn action_discriminant(action: &aa_core::GovernanceAction) -> u64 {
 
     let mut h = AHasher::default();
     match action {
-        aa_core::GovernanceAction::ToolCall { name, .. } => {
+        aa_core::GovernanceAction::ToolCall { name, args } => {
             "tool".hash(&mut h);
             name.hash(&mut h);
+            // Hash the full args payload: stage-5 `requires_approval_if`
+            // predicates (e.g. `args.amount > 1000`) are evaluated against the
+            // args, so two calls to the same tool with different args must not
+            // collide on the cache key (AAASM-3787).
+            args.hash(&mut h);
         }
         aa_core::GovernanceAction::ToolResult { tool_name, .. } => {
             "tool_result".hash(&mut h);
--- a/aa-gateway/src/engine/cache.rs
+++ b/aa-gateway/src/engine/cache.rs
@@ -209,4 +209,21 @@ mod tests {
         let key_e2 = CacheKey::new(&[1u8; 16], 2, &action);
         assert_ne!(key_e1, key_e2);
     }
+
+    #[test]
+    fn different_tool_args_produce_different_keys() {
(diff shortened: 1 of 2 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review (AAASM-3787)

Independent senior-Rust review. I did not write this PR.

### CI status
All required checks **green** (Build, Test, Clippy, Format, Rustdoc, Integration macOS, conformance, CodeQL). Coverage / SonarCloud / Dashboard jobs are `skipping` (no relevant paths touched) — acceptance-deferred, not failures. No red checks.

### Scope vs ticket
| Ticket sub-point | Status |
|---|---|
| `ToolCall` cache key must include `args` | ✅ `args.hash(&mut h)` |
| `ToolResult` cache key must include `result` (response-side mirror) | ✅ `result.hash(&mut h)` |
| Regression: benign→malicious same-tool re-evaluates approval | ✅ `cached_allow_not_served_for_dangerous_args_through_cascade` (engine-level, end-to-end through the cascade) |
| Key-collision unit coverage | ✅ `different_tool_args_produce_different_keys`, `different_tool_results_produce_different_keys` |
| "Fold-in" schedule/budget-state caching | ⚠️ Explicitly scoped out as a follow-up. Justification is sound: that residual is TTL-bounded (60 s) and the budget direction fails **safe** (over-deny). Reasonable to defer; recommend a tracking ticket so it is not lost. |

### Side-effect / regression assessment
- **Correctness for non-approval flows: unaffected.** Adding the payload to `action_hash` only makes the key *more specific*. The only behavioural consequence is a lower cache hit-rate (more re-evaluations) — it can never produce a wrong `Allow`/`Deny`. Safe direction.
- **No `None`/serialization hazard.** `ToolCall.args` and `ToolResult.result` are `String` (not `Option`), so the `..` removal is total and there is no missing-field/`None` branch to mishandle. `AHasher` over a `String` cannot panic.
- **Hashing the raw JSON string** means two semantically-equal but textually-different payloads (whitespace/key-order) get distinct keys → a cache miss, never a mis-cache. Safe.
- **No public/persisted surface.** `CacheKey` is derived on the fly; no schema/wire change.
- Nit (non-blocking): the e2e test's predicate is `args.amount > 100` while the PR description's narrative uses `> 1000`. Test data only — does not affect the assertion's validity.

### Verdict: **APPROVE-READY**
Tightly-scoped, correct, fail-safe, well-tested fix. Only ask: open a follow-up ticket for the deferred schedule/budget-state caching residual.

PR #1266, opened 2026-06-26
[AAASM-3789] 🐛 (aa-api): Guard webhook /test against SSRF

Chisanan232:
## Description

Closes the authenticated SSRF in the alert-webhook destination test-fire
(`POST /api/v1/alerts/destinations/{id}/test`). Previously a caller with Write
scope could point a webhook destination at `http://169.254.169.254/…` (cloud
metadata), loopback, or any RFC1918 host, fire the connector, and read the
upstream response — the validation was scheme-only, the HTTP client followed
redirects, and the upstream body was reflected back to the caller.

Three layers of defense are added:

1. **Egress guard** (`destinations/validate.rs`) — new `validate_webhook_egress`
   resolves the URL host and rejects loopback / RFC1918 private / link-local
   (`169.254.0.0/16`, incl. the IMDS address) / IPv6 ULA (`fc00::/7`) /
   link-local (`fe80::/10`) / unspecified addresses. Literal-IP hosts are checked
   directly; hostnames are resolved and **every** returned address must be
   allowed (defeats DNS-rebinding to an internal target). It is enforced in the
   `test_destination` handler **before dispatch** (run on the blocking pool since
   it does a DNS lookup) and returns `400` with detail
   `webhook.url resolves to a disallowed internal address`.
2. **No redirects** (`connectors/mod.rs`) — the shared `reqwest::Client` now uses
   `redirect(Policy::none())`, so an allowlisted target cannot 3xx-bounce inward;
   a redirect is surfaced as a non-2xx outcome instead of being chased.
3. **No body reflection** (`connectors/webhook.rs`) — the webhook connector no
   longer captures or returns the upstream response body on either the success
   or HTTP-error path, so internal/origin content can never leak to the caller.

A documented, **secure-by-default** opt-out (`AA_ALLOW_PRIVATE_WEBHOOK_EGRESS`)
lets a *limited-function self-hosted* deployment legitimately reach a webhook on
its own private network — the configurable egress allowlist the ticket suggests.
Unset (the SaaS default) keeps the guard active.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [ ] No
- [x] Yes (describe below)

Behavioural change (no schema change): a webhook test-fire to an internal
address now returns `400` instead of dispatching, and the
`connector_response_body` / `connector_body` fields are now always empty for the
webhook kind (the upstream body is no longer reflected). The response DTO shapes
(`TestDestinationResponse`, `ConnectorFailedBody`) are unchanged, so the OpenAPI
contract and dashboard generated types are unaffected.

## Related Issues

- Related Jira ticket: AAASM-3789
- Related GitHub issues: #XX

Closes AAASM-3789

## Testing

New / updated coverage (all green via `cargo nextest run -p aa-api`):
- `validate::tests::egress_rejects_loopback_metadata_and_private` — `127.0.0.1`,
  `169.254.169.254`, `10.0.0.1`, `192.168.1.1`, `172.16.0.1`, `[::1]` rejected.
- `validate::tests::egress_allows_public_ip` — `8.8.8.8` / `1.1.1.1` allowed.
- `validate::tests::egress_env_escape_allows_private` — opt-out works.
- `connectors::webhook_dispatch_never_reflects_response_body` — body withheld.
- `connectors::webhook_dispatch_does_not_follow_redirects` — 308 not chased.
- `destinations::test_webhook_to_internal_host_is_rejected` — route returns 400
  for metadata/loopback/RFC1918 targets.
- Existing webhook route tests updated to opt into private egress (loopback
  httpmock) and to assert the upstream body is no longer reflected.

How to verify manually:
```
# default (guard on): rejected
curl -XPOST .../alerts/destinations/{id}/test   # id is a 169.254.169.254 webhook -> 400
# self-host opt-out:
AA_ALLOW_PRIVATE_WEBHOOK_EGRESS=1 <serve>        # private target allowed
```

## Checklist

- [ ] All CI checks passing

### DIFF
--- a/aa-api/src/destinations/connectors/mod.rs
+++ b/aa-api/src/destinations/connectors/mod.rs
@@ -100,6 +100,12 @@ pub(crate) fn shared_client() -> &'static Client {
         Client::builder()
             .connect_timeout(Duration::from_secs(8))
             .timeout(Duration::from_secs(15))
+            // AAASM-3789: never follow redirects. A redirect can bounce an
+            // otherwise-allowlisted destination to an internal address (cloud
+            // metadata, loopback, RFC1918), re-opening the SSRF the egress
+            // guard closes. A 3xx is surfaced to the caller as a non-2xx
+            // outcome instead of being chased inward.
+            .redirect(reqwest::redirect::Policy::none())
             .build()
             .expect("reqwest client construction failed")
     })
--- a/aa-api/src/destinations/connectors/webhook.rs
+++ b/aa-api/src/destinations/connectors/webhook.rs
@@ -8,7 +8,7 @@
 use chrono::Utc;
 
 use crate::destinations::connectors::{
-    shared_client, truncate_body, ConnectorError, DispatchOutcome, DispatchRequest, NotificationConnector,
+    shared_client, ConnectorError, DispatchOutcome, DispatchRequest, NotificationConnector,
 };
 use crate::destinations::types::{Destination, DestinationConfig};
 
@@ -45,17 +45,23 @@ impl NotificationConnector for WebhookConnector {
             .await
             .map_err(|e| ConnectorError::Transport(e.to_string()))?;
         let status = resp.status().as_u16();
-        let body = resp.text().await.unwrap_or_default();
-        let body = truncate_body(body);
+        // AAASM-3789: do NOT capture or return the upstream response body. A
+        // webhook test-fire must not reflect origin/internal response content
+        // back to the caller (SSRF data-exfiltration vector). Drain and discard
+        // the body so only the observed status is surfaced.
+        drop(resp.bytes().await);
 
         if (200..300).contains(&status) {
             Ok(DispatchOutcome {
                 delivered_at: Utc::now().to_rfc3339(),
                 connector_response_status: status,
-                connector_response_body: body,
+                connector_response_body: String::new(),
             })
         } else {
-            Err(ConnectorError::Http { status, body })
+            Err(ConnectorError::Http {
+                status,
+                body: String::new(),
+            })
         }
     }
 }
--- a/aa-api/src/destinations/validate.rs
+++ b/aa-api/src/destinations/validate.rs
@@ -4,6 +4,8 @@
 //! connector relies on at dispatch time (URL parse, https-only for Slack,
 //! non-empty PagerDuty routing key, OpsGenie key + team).
 
+use std::net::{IpAddr, ToSocketAddrs};
+
 use crate::destinations::types::{Destination, DestinationConfig};
 
 /// Reason a destination payload failed validation. The `'static str`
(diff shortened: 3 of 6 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review (AAASM-3789)

Independent senior-Rust review. I did not write this PR.

### CI status
All required checks **green** (Build, Test, Clippy, Format, Rustdoc, Integration macOS, OpenAPI drift + Spectral, conformance, CodeQL). Coverage / SonarCloud / Dashboard jobs `skipping` — acceptance-deferred, not failures. No red checks.

### Scope vs ticket
| Ticket sub-point | Status |
|---|---|
| Resolve host, reject loopback/link-local/private/ULA/metadata before dispatch | ✅ `validate_webhook_egress` + `ip_is_internal` (v4 loopback/private/link-local/unspecified/broadcast; v6 loopback/unspecified/`fc00::/7`/`fe80::/10`; v4-mapped-v6 unwrapped) |
| Reject on every redirect hop | ✅ Handled more strongly via `redirect(Policy::none())` (no hop to re-check) |
| Resolve hostnames; every resolved IP must pass (defeat rebind) | ✅ `to_socket_addrs` loop, all-must-pass, empty-resolution rejected |
| Do not return upstream body to caller | ✅ webhook connector drains+discards; `connector_response_body`/error `body` = `""` |
| Configurable egress allowlist | ✅ `AA_ALLOW_PRIVATE_WEBHOOK_EGRESS` opt-out, secure-by-default |
| Regression tests (169.254.169.254, 127.0.0.1, 10.0.0.0/8, redirect-to-internal, body-not-reflected) | ✅ all present (unit + connector + route-level 400) |

### Side-effect / regression assessment
- **Legitimate external webhooks: not blocked.** Public literal IPs and public-resolving hostnames pass. Good.
- **DNS error → false denial (availability):** a transient resolver failure turns a legitimate webhook `/test` into `400 "host could not be resolved"`. Acceptable trade-off, but worth documenting as expected behaviour.
- **TOCTOU residual (note, not blocking):** the guard resolves DNS, then `reqwest` resolves again at connect time. The multi-answer rebind is defeated (all answers must pass), but a *time-based* rebind (different answer on the second lookup) is not fully closed. This is the standard hard limit of resolve-then-connect guards; `redirect(none)` narrows it. Consider a pinned-IP dial or a resolving HTTP connector as hardening follow-up.
- **`redirect(Policy::none())` is global to the shared client** (Slack/PagerDuty/OpsGenie share it). Slack incoming-webhooks and the PD/OpsGenie REST APIs return 2xx directly and do not rely on 3xx, so no legitimate provider breaks. Low risk; flagging because the blast radius is wider than webhook-only.
- **Body reflection removed only for webhook.** Slack still reflects its body — but Slack is host-locked to `*.slack.com` in production (`validate_config`, `#[cfg(not(test))]`), and PD/OpsGenie use fixed API hosts, so the SSRF-read vector is webhook-specific. Scope is correct.
- **Dispatch surface coverage is complete.** I verified the alerts module does not yet fire connectors — `/test` (`routes/destinations.rs:431`) is the *only* connector-dispatch path, and it is guarded. Forward-looking note: when real alert delivery is wired, `validate_webhook_egress` must be applied on that path too.
- **Guard is at `/test`, not at create/update.** Creating an internal-target destination still returns 201 (test asserts this); only the fire is blocked. Fine given no other dispatch path exists today.
- **Test env-var mutation** (`set_var`/`remove_var`) is safe under `nextest` (process-per-test) but would race under plain `cargo test`. CI uses nextest. Minor.

### Verdict: **APPROVE-READY**
Defense-in-depth is correct and well-tested; scope matches the live dispatch surface. Notes above (TOCTOU residual, global redirect policy, apply-guard-to-future-delivery-path) are hardening follow-ups, none blocking.

PR #1294, opened 2026-06-27
[AAASM-3864] 🔒 (aa-proxy): Close egress enforcement bypasses

Chisanan232:
## Description

Closes three egress-enforcement bypasses in the `aa-proxy` layer-2 interceptor (all on code paths that are **not** behind any default-off config):

**(a) Keep-alive request pipelining.** `handle_llm_mitm` and the non-MCP branch of `handle_non_llm_with_gateway` inspected only the *first* request on a connection and then dropped into an unconditional bidirectional `tokio::io::copy`. On a persistent HTTP/1.1 tunnel, requests #2..N skipped the credential scan and MCP CheckAction. **Fix chosen: force `Connection: close` + single-exchange relay.** The request serializer now drops any inbound `Connection` header and emits exactly one `Connection: close`, so the upstream tears the connection down after one response. The data path relays **only** upstream→client and never copies further client bytes upstream — so a pipelined second request can never reach upstream un-inspected. I chose this over a per-request inspection loop because it is materially simpler, has no realistic regression surface (the proxy already opens one upstream dial per client connection, so upstream keep-alive was never reused across requests anyway), and is unambiguously fail-closed.

**(b) Plain-HTTP allowlist/denylist bypass.** `handle_plain_http` only re-validated resolved IPs (SSRF) and never called `connect_deny_reason`, so an `http://` scheme-downgrade bypassed `denied_hosts` and the network allowlist that the CONNECT path enforces. It now calls `connect_deny_reason` on the parsed host (port stripped) before any upstream dial and returns `403` on deny, mirroring the CONNECT path.

**(c) Chunked-body inspection gap.** `read_http_request` reads bodies by `Content-Length` only; a `Transfer-Encoding: chunked` request yielded an empty parsed body, so the credential scanner and `parse_mcp_request` saw nothing while the real payload was forwarded un-inspected. The parser now **fails closed**: it rejects chunked requests with a `ProxyError::Config` error, dropping the connection rather than forwarding an un-scanned body.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

Forwarded requests now always carry `Connection: close` and chunked request bodies are refused; both are intended fail-closed enforcement behaviours, not public-API changes.

## Related Issues

- Related Jira ticket: AAASM-3864

## Testing

Regression coverage (all pass via `cargo nextest run -p aa-proxy`, 171 tests):
- (a) `serialize_forces_connection_close_and_strips_inbound_connection` (unit) and `attacker::keep_alive_second_request_is_not_forwarded_uninspected` (integration — pipelines a 2nd request on a MitM tunnel, asserts the counting upstream sees exactly one request).
- (b) `plain_http_request_to_denied_host_is_rejected_with_403` (integration) and the updated loopback-literal SSRF unit test now asserting the explicit 403.
- (c) `rejects_chunked_transfer_encoding_request` (unit).

Validation gate: `cargo build -p aa-proxy`, `cargo nextest run -p aa-proxy` (171 passed), `cargo fmt --all`, `cargo clippy -p aa-proxy --all-targets -- -D warnings` — all clean.

## Checklist

- [x] All CI checks passing

### DIFF
--- a/aa-proxy/src/proxy/http.rs
+++ b/aa-proxy/src/proxy/http.rs
@@ -489,6 +489,24 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn rejects_chunked_transfer_encoding_request() {
+        // AAASM-3864: a request advertising a chunked body cannot be inspected
+        // by this Content-Length-only parser, so it must fail closed rather than
+        // yield an empty body that forwards an un-scanned payload upstream.
+        let raw = b"POST /v1/chat/completions HTTP/1.1\r\n\
+                    Host: api.openai.com\r\n\
+                    Transfer-Encoding: chunked\r\n\
+                    \r\n\
+                    5\r\nhello\r\n0\r\n\r\n";
+        let mut reader = make_reader(raw);
+        let err = read_http_request(&mut reader).await.unwrap_err();
+        assert!(
+            matches!(&err, ProxyError::Config(msg) if msg.contains("chunked")),
+            "expected fail-closed chunked rejection, got: {err:?}"
+        );
+    }
+
+    #[tokio::test]
     async fn unexpected_eof_reading_request_headers_is_error() {
         // Stream ends after the request line, before the header-terminating
         // blank line — this is a truncated request, not a clean inter-request
--- a/aa-proxy/src/proxy/http.rs
+++ b/aa-proxy/src/proxy/http.rs
@@ -7,9 +7,10 @@
 //!
 //! Scope is intentionally small: only requests with a literal
 //! `Content-Length` header are fully supported. Requests using
-//! `Transfer-Encoding: chunked` parse the head but leave the body empty —
-//! that variant is rare for LLM POSTs and can be added when needed without
-//! breaking this API.
+//! `Transfer-Encoding: chunked` cannot be inspected by this parser, so the
+//! request side rejects them (fail-closed, AAASM-3864) rather than forwarding
+//! an un-scanned body. The response side still parses the head and leaves a
+//! chunked body empty (the MCP path falls back to a transparent relay).
 
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, BufReader};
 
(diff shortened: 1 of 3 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: ✅ Ready to merge**

### Scope coverage (AAASM-3864)
All three egress bypasses from the ticket are addressed:
- **(a) keep-alive pipelining** — `serialize_http_request_with_auth` strips any inbound `Connection` header and forces a single `Connection: close`; both the LLM-MitM and non-LLM gateway relays were changed from bidirectional `select!` copy to a one-direction upstream→client `copy` + `shutdown`. A pipelined request #2 can no longer reach upstream un-inspected.
- **(b) plain-HTTP scheme-downgrade** — `handle_plain_http` now calls `connect_deny_reason(deny_host)` (port stripped) before any dial and returns 403, mirroring the CONNECT path. Closes the `http://` denylist/allowlist bypass.
- **(c) chunked request bodies** — `read_http_request` fails closed on `Transfer-Encoding: chunked` rather than forwarding an un-scanned body.

- **Streaming responses preserved**: the new upstream→client `copy` relays the full upstream byte stream (SSE/chunked responses included) until upstream EOFs; `Connection: close` bounds it. No regression for normal single request/response or streamed responses.
- **Chunked rejection not over-broad**: match is `transfer-encoding` header containing `chunked`, so `gzip, chunked` etc. are caught; plain Content-Length requests unaffected. LLM request bodies use Content-Length, so legitimate impact is low (fail-closed by design).
- **Denylist match correct**: `host.split(':').next()` strips the port before `connect_deny_reason`.
- **Behavioral tradeoff (acceptable)**: HTTP/1.1 keep-alive request *pipelining* over a MitM tunnel no longer works — each request needs a fresh exchange. Intended. Truly bidirectional protocols over the inspected tunnel (not the LLM/HTTP use case) would not survive, but that path isn't in scope.
- No remaining un-inspected request path found; the MCP response-parse-failure fallback also relays upstream→client only.

### CI state
All required checks green (Build, Test, Clippy, conformance, bypass-permissions integration). Only coverage/SonarCloud/dashboard jobs skipping — acceptable. New regression tests cover all three sub-fixes incl. an attacker pipelining test asserting upstream sees exactly one request.

**Caveat:** none blocking — just note the keep-alive-pipelining and chunked-request behavior changes are intentional fail-closed tradeoffs.

PR #1296, opened 2026-06-27
[AAASM-3873] 🔒 (aa-runtime): Bind op-control kill switch to server identity

Chisanan232:
## Description

`aa-runtime`'s op-control kill switch (`handle_policy_query`) skipped the op-control consult entirely when the agent-supplied `trace_id` was empty, and otherwise matched only the op id `"{trace_id}:{span_id}"` — both fields are attacker-controlled on `CheckActionRequest`. A non-cooperative SDK could omit or alter `trace_id` to dodge an operator terminate/pause.

**Binding approach chosen:** bind op-control matching to a **server-side identity** rather than the agent-supplied `trace_id`.

- New reserved halt keys in `aa-runtime/src/op_control.rs`:
  - `GLOBAL_HALT_OP_ID` (`"*"`) — a global kill switch.
  - `agent_halt_op_id(agent_id)` (`"agent:{agent_id}"`) — an agent-scoped halt keyed by the agent identity the runtime itself knows (`PipelineConfig::agent_id`). Both are disjoint from the colon-bearing per-op id namespace.
- `handle_policy_query` now **always** consults the global + agent-scoped keys on every request, regardless of `trace_id`. The op-specific `"{trace_id}:{span_id}"` key is added only as an *additional, finer-grained* match when a `trace_id` is actually supplied, so it can never reduce coverage.
- `op_control_halts` scans all bound keys; a terminate on any key fast-fails with `OpTerminatedError`, taking priority over a pause; a pause on any key blocks until resume.

Result: an empty/absent or forged `trace_id` no longer bypasses an active **agent-level or global** halt. As the ticket notes, defeating a halt scoped to a *specific* op by claiming a *different* op remains inherent to the advisory SDK fast-path (an unlabelled request cannot be attributed to that op), but the operator's agent-wide and global kill switches are now unbypassable. The runtime stays authoritative; the gateway can address an agent/global halt by publishing under these reserved op ids (operator-facing emission is the natural follow-up).

## Type of Change

- [x] 🐛 Bug fix (security)

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-3873

Closes AAASM-3873

## Testing

Validation (worktree, `aa-runtime`):
- `cargo build -p aa-runtime` — ok
- `cargo nextest run -p aa-runtime` — 361 passed, 2 skipped
- `cargo fmt --all` — clean
- `cargo clippy -p aa-runtime --all-targets -- -D warnings` — clean

New regression tests:
- `op_control_agent_terminate_halts_with_empty_trace_id` — agent terminate enforced with empty `trace_id`.
- `op_control_agent_terminate_halts_with_altered_trace_id` — agent terminate enforced with a forged `trace_id`.
- `op_control_agent_pause_blocks_then_resume_with_empty_trace_id` — agent pause blocks an empty-trace check, resume releases it (normal flow still works).
- `op_control_other_op_terminate_does_not_block_normal_request` — no false positives: an unrelated op terminate does not block a normal request.
- `server_side_halt_keys_are_disjoint_from_per_op_ids` (unit) — key namespaces don't collide.

## Checklist

### DIFF
--- a/aa-runtime/src/op_control.rs
+++ b/aa-runtime/src/op_control.rs
@@ -41,6 +41,27 @@ const INITIAL_BACKOFF: Duration = Duration::from_secs(1);
 /// Upper bound on the reconnect delay (1s → 2 → 4 → … → 32s cap).
 const MAX_BACKOFF: Duration = Duration::from_secs(32);
 
+/// Reserved `op_id` addressing **every** op for the subscribed agent — a global
+/// kill switch.
+///
+/// A signal the gateway publishes under this id halts the agent regardless of
+/// which op (if any) a request claims, so it cannot be evaded by an absent or
+/// altered agent-supplied `trace_id` (AAASM-3873). It never collides with a
+/// per-op id, which always has the colon-bearing `"{trace_id}:{span_id}"` form.
+pub const GLOBAL_HALT_OP_ID: &str = "*";
+
+/// Reserved `op_id` addressing the whole agent identified by `agent_id`.
+///
+/// Op-control matching binds to this **server-side** identity — the agent id the
+/// runtime itself knows — rather than the attacker-controlled `trace_id` /
+/// `span_id` on a `CheckActionRequest`. An agent-level terminate/pause published
+/// under this id applies to every request from the agent, including ones that
+/// omit or forge a `trace_id` (AAASM-3873). The `agent:` prefix keeps it disjoint
+/// from per-op ids of the form `"{trace_id}:{span_id}"`.
+pub fn agent_halt_op_id(agent_id: &str) -> String {
+    format!("agent:{agent_id}")
+}
+
 /// The runtime-observed lifecycle state of a single op.
 ///
 /// Derived from the most recent [`OpControlSignal`] the gateway pushed for the
--- a/aa-runtime/src/op_control.rs
+++ b/aa-runtime/src/op_control.rs
@@ -271,6 +271,20 @@ mod tests {
     }
 
     #[test]
+    fn server_side_halt_keys_are_disjoint_from_per_op_ids() {
+        // Per-op ids always carry a colon ("{trace_id}:{span_id}"); the
+        // server-side halt keys must not collide with that namespace.
+        assert!(!GLOBAL_HALT_OP_ID.contains(':'));
+        assert_eq!(agent_halt_op_id("svc-agent"), "agent:svc-agent");
+
+        // An agent-level halt is addressable independently of any op halt.
+        let store = OpControlStore::new();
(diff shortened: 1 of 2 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: ⚠️ Ready with caveat** — runtime-side binding is correct; full end-to-end kill switch depends on a disclosed gateway follow-up.

### Scope coverage (AAASM-3873)
`handle_policy_query` now **always** consults op-control via two server-side keys — `GLOBAL_HALT_OP_ID` (`*`) and `agent_halt_op_id(agent_id)` (`agent:{id}`, keyed by `config.agent_id` the runtime itself knows) — and only *adds* the `{trace_id}:{span_id}` op-specific key when a trace_id is supplied. The `if !trace_id.is_empty()` early-return that caused the no-op bypass is gone. `op_control_halts` now scans all bound keys with terminate-over-pause priority.

- **Empty/forged trace_id no longer bypasses op-control consultation** ✅ (tests: `..._with_empty_trace_id`, `..._with_altered_trace_id`).
- **Global + agent keys always consulted regardless of trace_id** ✅; `agent_id` comes from server-side `PipelineConfig`, not the request.
- **Normal trace_id flow intact / no over-block** ✅: an unrelated op terminate does not block a normal request (`op_control_other_op_terminate_does_not_block_normal_request`).
- **Namespace overlap (harmless):** `agent:{id}` *can* collide with `{trace_id}:{span_id}` if an attacker sets trace_id=`agent`/span_id=`<id>`. This can only *add* a halt match for the attacker's own request (self-DoS at worst), never remove one — no escape. `*` cannot collide (per-op ids always contain a colon).

### ⚠️ Scope caveat — flag for maintainer
This PR is **aa-runtime only**. It makes the runtime *capable* of honoring agent-wide/global halts, but the operator's agent-level terminate/pause is **inert end-to-end until the gateway emits `OpControlSignal` under the reserved `*` / `agent:{id}` keys** — which the author explicitly notes as the "natural follow-up." Today the gateway keys signals by op (trace:span). So:
- The **empty-trace_id no-op bug is fully fixed** (op-control is now always consulted).
- The **agent-wide / global kill switch only becomes operative once the gateway-side emission lands.** Defeating an *op-scoped* halt by claiming a different op remains inherent to the advisory SDK fast-path (acknowledged).

Mergeable as the runtime half; ensure the gateway-emission follow-up ticket exists so the kill switch is actually wired end-to-end.

### CI state
All required checks green (Build incl. sidecar + 9-image language matrix + smokes, Test, Integration tests, Clippy, conformance). coverage/SonarCloud skipping — acceptable.

PR #1299, opened 2026-06-27
[AAASM-3869] 🔒 (aa-gateway): Bind audit events to verified caller

Chisanan232:
## Description

The gRPC `AuditService` RPCs (`report_events`, `stream_events`, `ingest_event`) run behind the fail-closed `auth_interceptor` (which injects a `VerifiedCaller` into request extensions) but never read it. `event.agent_id` is fully attacker-supplied yet drives WORM/hash-chained audit **attribution + lineage**, so any registered agent could submit events with `agent_id = victim` and forge audit history under another agent/tenant (AAASM-3869).

This binds every ingested event to the authenticated caller. The handlers now read the `VerifiedCaller` from the request extensions **before** `into_inner()` and reject any event whose composite `agent_id` does not resolve to the caller's own identity, mirroring how sibling services (`topology`/`approval`/`secrets`, AAASM-3845/3846) bind the caller.

**Binding predicate — per-agent (not per-tenant).** The event's composite `agent_id` is hashed to the same 16-byte registry key the interceptor derived for the caller (`registry::convert::proto_agent_id_to_key` ↔ `VerifiedCaller::agent_key`) and compared for equality. This is **stricter** than the sibling per-tenant reads, deliberately: audit attribution is inherently *self-reported* — an agent writes WORM entries only about its own activity — so per-tenant binding would still permit cross-agent forgery within a tenant, which corrupts the hash-chained log. Per-agent binding closes that gap entirely.

Failure modes (fail closed):
- Missing `VerifiedCaller` → `unauthenticated`.
- `agent_id` belongs to another agent, or is absent → `permission_denied`.
- `report_events` validates the whole batch before ingesting any event, so a poisoned batch writes nothing.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

A correctly-behaving SDK already reports events under its own authenticated identity, so legitimate self-attribution is unaffected.

## Related Issues

- Related Jira ticket: AAASM-3869

## Testing

New `audit_service` unit tests: cross-agent forgery → `permission_denied`; self-attributed event → ok; missing caller → `unauthenticated`; absent `agent_id` → denied; mixed batch rejected atomically. The seq-recovery and lineage integration tests now inject the `VerifiedCaller`, and the e2e server is fronted with the `auth_interceptor` so report/stream exercise the binding end to end with credential tokens.

Validation (gRPC only — no openapi/dashboard regen):
- `cargo build -p aa-gateway` ✅
- `cargo nextest run -p aa-gateway` ✅ (1250 passed; `policy_latency_test` flaky-passed on retry, known macOS flake)
- `cargo fmt --all` ✅
- `cargo clippy -p aa-gateway --all-targets -- -D warnings` ✅

## Checklist

- [ ] All CI checks passing

Closes AAASM-3869

### DIFF
--- a/aa-gateway/src/service/audit_service.rs
+++ b/aa-gateway/src/service/audit_service.rs
@@ -14,6 +14,7 @@ use aa_proto::assembly::audit::v1::audit_service_server::AuditService;
 use aa_proto::assembly::audit::v1::{AuditEvent, ReportEventsRequest, ReportEventsResponse, StreamEventsResponse};
 use aa_proto::assembly::common::v1::Decision;
 
+use crate::iam::VerifiedCaller;
 use crate::registry::{convert as registry_convert, AgentRegistry};
 use crate::service::convert;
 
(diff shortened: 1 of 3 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: ✅ Ready to merge**

### Scope coverage (AAASM-3869)
`report_events` and `stream_events` now read the `VerifiedCaller` injected by the fail-closed auth interceptor and reject any event whose `agent_id` does not resolve to the caller's own registry key (`proto_agent_id_to_key` ↔ `VerifiedCaller::agent_key`). Binding is **per-agent** (stricter than the sibling per-tenant reads), which is the right call — audit attribution is self-reported, so per-tenant would still allow intra-tenant forgery.

- **Per-agent binding correct** ✅: equality compares the caller's interceptor-derived key (from the *attacker's* registration) to the event's key — an attacker cannot make their own key equal the victim's, regardless of what `agent_id` they put in the event.
- **Batch atomicity** ✅: `report_events` validates the entire batch in a pre-loop and returns `permission_denied` before *any* `ingest_event`, so a poisoned batch writes nothing (test `report_events_rejects_mixed_batch_atomically`).
- **Fail-closed on missing caller** ✅: absent `VerifiedCaller` → `unauthenticated`; absent/None `agent_id` → `permission_denied` (`is_some_and` over `None` is false).
- **Legit self-reporting preserved** ✅: an agent reporting its own `agent_id` with its credential token passes; lineage-population test updated accordingly. Integration tests now register callers and front the service with the real auth interceptor.
- **Minor note (not blocking):** `stream_events` ingests incrementally, so a stream of [self, forged] writes the self event before rejecting+terminating on the forged one — i.e. not all-or-nothing like the batch path. This is inherent to streaming and **does not permit forgery** (the forged event is never written); only the batch path needs atomicity and has it.

### CI state
All required checks green (Build, Test, Integration tests, TimescaleDB, Migration drift, Clippy). coverage/SonarCloud skipping — acceptable. Regression tests cover cross-agent forgery, self-attribution, missing caller, absent agent_id, and mixed-batch rejection.

PR #1301, opened 2026-06-27
[AAASM-3874] 🔒 (aa-security): Reject unknown policy keys / fail-closed parse

Chisanan232:
## Description

Fixes a fail-open in the canonical policy parser (`aa-security/src/policy/parse.rs`). `#[serde(flatten)]` on the envelope precludes `deny_unknown_fields`, so a misspelled security key (`dney:` for `deny:`, `allow_list:` for `allowlist:`) or a misspelled section (`capabilties:`) was silently dropped — yielding an empty, permissive policy that still parsed `Ok`. A forgotten/misspelled tool `allow:` key also silently permitted the tool (default was allow-by-default).

Two-part fix:

1. **Reject unknown keys.** Parse to a raw `serde_yaml::Value` first and validate it against the known schema before typed deserialization. Validation is strict on the cross-layer security dimensions this crate owns — `network` (`allowlist`), `capabilities` (`allow`/`deny`), `syscalls` (`allow`), and each `tools.<name>` (`allow`/`requires_approval_if`/`limit_per_hour`) — plus the top-level and `spec` key sets. L7-only sections (`budget`/`schedule`/`data`) are accepted but **not** descended into, so this crate does not couple to the gateway-owned inner schema. Unknown keys now raise the new `PolicyParseError::UnknownKey { path, key }`.
2. **Deny-by-default tools.** A tool entry that omits `allow:` now defaults to **deny** instead of allow.

Malformed YAML and unknown capability/syscall tokens continue to fail closed (unchanged).

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes (describe below)

**Intentional behaviour changes (security hardening):**
- Policies with unknown/misspelled keys in the security sections now **error** instead of parsing into a silently-weakened policy.
- A `tools.<name>` entry without an explicit `allow:` now defaults to **deny** (previously allow). All on-disk `policy-examples/*.yaml` already set `allow:` explicitly, so the example corpus is unaffected.

Egress semantics are deliberately unchanged: an empty/absent `allowlist` still means "no restriction" (documented cross-layer contract). What is fixed is the *typo* vector — a misspelled `allowlist` key now errors rather than collapsing to an empty allow-all list.

## Related Issues

- Related Jira ticket: AAASM-3874
- Closes AAASM-3874

## Testing

`cargo build -p aa-security` (with and without `serde`), `cargo nextest run -p aa-security --features serde` (117 passed, incl. the `policy-examples/` corpus test), `cargo fmt --all`, `cargo clippy -p aa-security --all-targets --features serde -- -D warnings`. The pre-commit hook additionally fmt+clippy-checked the full workspace (all crates compile, no downstream breakage).

New regression tests: misspelled `deny`/`allowlist`/`allow` keys error; misspelled section errors; unknown top-level key errors; tool without `allow:` defaults to deny; valid L7-only spec sections still parse; malformed tokens still fail closed.

## Checklist

- [x] All CI checks passing

### DIFF
--- a/aa-security/src/policy/error.rs
+++ b/aa-security/src/policy/error.rs
@@ -21,6 +21,19 @@ pub enum PolicyParseError {
         /// Why it was rejected.
         reason: String,
     },
+    /// A structural key was not part of the known policy schema.
+    ///
+    /// Raised when a security-relevant section or field is misspelled (e.g.
+    /// `dney:` for `deny:`, `allow_list:` for `allowlist:`). Such typos would
+    /// otherwise be silently dropped by `#[serde(flatten)]` — yielding an empty,
+    /// permissive policy that parses successfully. Surfacing them keeps policy
+    /// parsing fail-closed (AAASM-3874).
+    UnknownKey {
+        /// Dotted path to the mapping that contained the unknown key.
+        path: String,
+        /// The unrecognised key.
+        key: String,
+    },
 }
 
 impl fmt::Display for PolicyParseError {
@@ -32,6 +45,9 @@ impl fmt::Display for PolicyParseError {
             }
             Self::InvalidSyscall { raw, reason } => {
                 write!(f, "invalid syscall {raw:?}: {reason}")
+            }
+            Self::UnknownKey { path, key } => {
+                write!(f, "unknown policy key {key:?} under {path}")
             }
         }
     }
@@ -56,5 +72,14 @@ mod tests {
             reason: "unknown capability: 'teleport'".to_string(),
         };
         assert!(err.to_string().contains("teleport"));
+    }
+
+    #[test]
+    fn display_unknown_key() {
+        let err = PolicyParseError::UnknownKey {
+            path: "capabilities".to_string(),
+            key: "dney".to_string(),
+        };
+        assert_eq!(err.to_string(), "unknown policy key \"dney\" under capabilities");
     }
 }
--- a/aa-security/src/policy/parse.rs
+++ b/aa-security/src/policy/parse.rs
@@ -362,6 +362,95 @@ spec:
     }
 
     #[test]
+    fn rejects_misspelled_capability_deny_key() {
+        // `dney` instead of `deny`: previously dropped silently, leaving an
+        // empty (permissive) deny floor. Must now fail closed (AAASM-3874).
+        let yaml = "spec:\n  capabilities:\n    dney:\n      - terminal_exec\n";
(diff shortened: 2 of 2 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: ✅ Ready to merge**

### Scope coverage (AAASM-3874)
Both fail-open vectors addressed:
- **Unknown-key rejection** — `validate_schema` walks the raw YAML before typed deserialization and errors (`PolicyParseError::UnknownKey`) on any unrecognized key in the security dimensions this crate owns: top-level, `spec`, `network` (`allowlist`), `capabilities` (`allow`/`deny`), `syscalls` (`allow`), `tools.<name>` (`allow`/`requires_approval_if`/`limit_per_hour`). Closes the misspelled-`deny:`/`allowlist:` silent-drop from `#[serde(flatten)]`.
- **Tool deny-by-default** — `allow: t.allow.unwrap_or(false)` (was `true`). A tool entry omitting (or misspelling) `allow:` now denies.

- **Corpus not broken** ✅ — I independently checked `policy-examples/*.yaml`: every `network:` block uses only `allowlist`; every named tool entry sets `allow:` explicitly (true/false). `metadata` children (`name`/`description`/`version`) sit under `metadata` which is accepted-but-not-descended; all spec children used (`scope`, `network`, `capabilities`, `tools`, `syscalls`, `schedule`, `budget`, `data`, `approval_timeout_secs`) are in the allowlists. So the deny-by-default flip and key validation leave the corpus parsing identically (and CI corpus tests pass).
- **L7-only sections decoupled correctly** ✅ — `schedule`/`budget`/`data` are accepted but not descended into, avoiding coupling to the gateway engine's inner schema (test `accepts_l7_only_spec_sections`).
- **Egress empty-list semantics preserved** ✅ — empty `allowlist` still means "no restriction"; only the *typo* vector (which silently produced that state) is closed. This is a deliberate, documented scope boundary — the empty-list = allow-all behavior itself is unchanged.
- **Malformed tokens still fail closed** ✅ (unchanged).
- **Note:** the deny-by-default flip applies to the `aa-security` parser. Any consumer that relied on the old allow-by-default for a tool with no explicit `allow:` would now see deny — none exist in the corpus, but worth confirming no external policy files outside the repo depend on the old default.

### CI state
All required checks green (Build, Test, Clippy, conformance, Standalone build). coverage/SonarCloud skipping — acceptable. New tests cover misspelled capability/network/spec/tool/top-level keys, deny-by-default, and L7 passthrough.

PR #1314, opened 2026-06-28
[AAASM-3888] 🔒 (aa-gateway): Authorize gRPC batch_check + op-control plane

Chisanan232:
## Description

Closes two HIGH gRPC control-plane authZ gaps that both live in `aa-gateway/src/service/policy_service.rs`, so they ship as one PR to avoid a self-conflict.

**AAASM-3888 — `batch_check` skipped credential validation → impersonation.**
`PolicyService` runs under the non-rejecting `enrich` interceptor, so the request-body `{org,team,agent}` triple and `credential_token` are attacker-controlled. `check_action` calls `validate_credential_token`, but `batch_check` looped `evaluate_one` + `record_audit` / `record_spend` / `maybe_suspend_agent` without it — a credential-less peer could forge a victim identity per batch entry (forged WORM audit, budget-exhaustion spend, victim suspension, approvals as the victim). Fix: `batch_check` now validates the credential token on each entry **before** any evaluation or side-effect, mirroring `check_action`; on failure it pushes the Deny and `continue`s (no side-effects), upholding `evaluate_one`'s tenancy-anchoring invariant.

**AAASM-3889 — op-control plane: NATS bridge forgery + `op_control_stream` zero-auth.**
- **NATS consume (subject↔payload binding):** the bridge acted on each envelope's `global`/`org`/`team`/`agent` fields verbatim, never comparing them to the subject the message was delivered on, so any publisher reaching one op-control subject could embed a different tenant — or `global:true` — in the JSON body (cross-tenant / fleet-wide kill, nullifying the subject ACL). New `subject_authorizes_envelope` drops any message whose payload does not map back to its delivery subject, and any `global` payload not on the global subject.
- **NATS connection auth/TLS:** the bridge and publisher connected with a bare `async_nats::connect(url)` (unauthenticated, plaintext). They now connect via `async_nats::ConnectOptions` driven by `OpControlNatsConfig` — JWT credentials file (`AA_OPCONTROL_NATS_CREDS`), nkey seed (`AA_OPCONTROL_NATS_NKEY`), and TLS (`AA_OPCONTROL_NATS_TLS`). Optional so loopback dev still works; production can require them. The op-control bus trust model is documented in the module/config docs.
- **`op_control_stream` subscriber authZ:** the endpoint validated no credential and bound the subscription to nothing, so a credential-less peer could subscribe with an arbitrary triple and read another tenant's pause/resume/terminate signals + op_ids. When an `AgentRegistry` is attached it now requires a `VerifiedCaller` (absent → `Unauthenticated`) and rejects any triple that does not resolve to the caller's own registered identity (→ `PermissionDenied`). Skipped only when no registry is attached (untenanted/test deployments), mirroring `validate_credential_token`.

## Type of Change

- [x] 🐛 Bug fix (security)

## Breaking Changes

- [x] No

Behaviour change for hardening: in a tenanted deployment (registry attached) `op_control_stream` now requires a valid credential bound to the caller's identity. `OpControlNatsConfig` gained optional `creds_path`/`nkey`/`tls` fields (all default to the previous plaintext/no-auth behaviour). No proto/OpenAPI changes — gRPC-only.

## Related Issues

- Related Jira tickets: AAASM-3888, AAASM-3889
- Closes AAASM-3888
- Closes AAASM-3889

## Testing

- `batch_check_credential_validation_test`: forged (empty / mismatched) credential is denied as `a2a_identity_verification`; over-budget victim is **not** suspended and **no** forged tool-call/budget audit is emitted; a valid-token entry is still allowed.
- `nats::tests::subject_*`: unit tests for `subject_authorizes_envelope` (own-subject OK; cross-tenant, forged-global, and per-agent-on-global all rejected).
- `op_control_stream_authz_test`: tokenless → `Unauthenticated`; cross-tenant triple → `PermissionDenied`; own identity → admitted and receives its halt.
- `op_control_nats_bridge_test::bridge_drops_envelopes_whose_payload_does_not_match_their_subject`: real-NATS JetStream testcontainer — forged per-agent + forged global dropped, legit halt delivered.

Validation gate (all green): `cargo build -p aa-gateway`; `cargo nextest run -p aa-gateway` (1311 passed incl. the NATS e2e testcontainer tests); `cargo fmt --all -- --check`; `cargo clippy -p aa-gateway --all-targets -- -D warnings`. No OpenAPI drift (gRPC-only).

## Checklist

### DIFF
--- a/aa-gateway/src/ops/nats.rs
+++ b/aa-gateway/src/ops/nats.rs
@@ -128,6 +128,30 @@ pub fn subject_for(envelope: &OpControlWireEnvelope) -> String {
     format!("{SUBJECT_PREFIX}.{tenant}.{agent}")
 }
 
+/// AAASM-3889 — authorize a received op-control message by binding its payload
+/// target to the subject it was actually delivered on.
+///
+/// The NATS subject namespace is the access-control boundary: a publisher is only
+/// trusted to act on the tenant / agent encoded in the subject it published to.
+/// Without this check the bridge acts on the envelope's `global` / `org_id` /
+/// `team_id` / `agent_id` fields **verbatim**, so any publisher able to reach
+/// *any* op-control subject could embed a different tenant — or `global: true` —
+/// in the JSON body and trigger a cross-tenant or fleet-wide kill, nullifying any
+/// NATS subject ACL. We re-derive the subject the payload *would* have been
+/// published under ([`subject_for`]) and require it to equal the delivery subject;
+/// a `global` envelope is additionally required to arrive on [`GLOBAL_SUBJECT`].
+fn subject_authorizes_envelope(subject: &str, envelope: &OpControlWireEnvelope) -> bool {
+    if envelope.global {
+        // A fleet-wide halt is only honored on the dedicated global subject.
+        return subject == GLOBAL_SUBJECT;
+    }
+    // A per-agent envelope must arrive on exactly the (sanitized) tenant/agent
+    // subject its own fields map to. `subject_for` never returns GLOBAL_SUBJECT
+    // for a non-global envelope, so a non-global payload published on the global
+    // subject is rejected here too.
+    subject_for(envelope) == subject
+}
+
 /// Replace every character outside `[A-Za-z0-9_-]` with `_` so the result is a
 /// single valid NATS subject token (NATS reserves `.`, `*`, `>` and whitespace).
 fn sanitize_token(raw: &str) -> String {
@@ -663,8 +687,21 @@ async fn bridge_once(
         let message = message.map_err(|e| OpControlNatsError::Consumer(e.to_string()))?;
         match serde_json::from_slice::<OpControlWireEnvelope>(&message.payload) {
             Ok(envelope) => {
-                metrics::counter!("aa_op_control_bridge_forwarded_total").increment(1);
-                forward_to_broadcast(publisher, envelope);
+                // AAASM-3889: reject a payload whose target does not match the
+                // subject it was delivered on (a forged tenant / global flag).
+                if subject_authorizes_envelope(message.subject.as_str(), &envelope) {
+                    metrics::counter!("aa_op_control_bridge_forwarded_total").increment(1);
+                    forward_to_broadcast(publisher, envelope);
+                } else {
+                    metrics::counter!("aa_op_control_bridge_rejected_total").increment(1);
+                    tracing::warn!(
+                        subject = %message.subject,
+                        expected_subject = %subject_for(&envelope),
+                        global = envelope.global,
+                        "op-control bridge: dropping envelope whose payload target does not match its \
+                         delivery subject (possible forgery)"
+                    );
+                }
             }
             Err(err) => {
                 tracing::warn!(%err, "op-control bridge: dropping undecodable message");
--- a/aa-gateway/src/ops/nats.rs
+++ b/aa-gateway/src/ops/nats.rs
@@ -187,25 +187,103 @@ impl OpControlNatsError {
 }
 
 /// Runtime configuration for the op-control NATS bridge / publisher.
+///
+/// AAASM-3889 — the op-control bus is a **fleet kill switch**: anyone able to
+/// publish to it can pause / resume / terminate runtimes. The bus must therefore
+/// be treated as a trusted, authenticated channel, never a bare `nats://` URL on
+/// an open network. Authentication (a JWT credentials file or an nkey seed) and
+/// TLS are configured here and applied via [`async_nats::ConnectOptions`] for
+/// **both** the publisher (aa-api) and the consumer bridge (aa-gateway). They are
+/// **optional** so a single-host local/dev stack (a NATS server bound to
+/// loopback) still works unconfigured, but a production deployment MUST set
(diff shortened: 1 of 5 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

Adversarial security review of the two HIGH gRPC control-plane authZ fixes (AAASM-3888, AAASM-3889).

### CI
All required checks green via `gh pr checks 1314` — `Test`, `Build`, `Clippy lint`, `Format check`, `Rust conformance`, `Integration tests (macos)`, CodeQL/Analyze all pass; remaining entries are `skipping` (path-filtered, not applicable to a gRPC-only Rust change). No reds.

### Scope coverage
**AAASM-3888 — `batch_check` impersonation: FIXED & COMPLETE.**
`policy_service.rs:1488-1502` now calls `validate_credential_token(req)` at the **top** of the per-entry loop, *before* `evaluate_one` (:1503), `record_audit`, `record_spend`, and `maybe_suspend_agent`. On rejection it pushes the Deny and `continue`s — zero side-effects. This is exact parity with `check_action` (:1365 → evaluate at :1375). `validate_credential_token` (:978) returns `Some(Deny)` only for registered-but-empty/mismatched token or a token registered to a *different* agent, and `None` for the legit/fixture paths — so the side-effecting code is unreachable for a forged entry.

**AAASM-3889 — op-control plane: all three sub-parts FIXED & COMPLETE.**
- (a) **Connection auth/TLS:** `OpControlNatsConfig` gained `creds_path`/`nkey`/`tls` (`from_env` reads `AA_OPCONTROL_NATS_{CREDS,NKEY,TLS}`). New `connect()`/`connect_options()` build `async_nats::ConnectOptions` with `credentials_file` / `nkey` / `require_tls`. Both the publisher (`OpControlNatsPublisher::connect`, nats.rs:~470) and the consumer bridge (`bridge_once`, :~658) now go through `config.connect()` instead of bare `async_nats::connect(url)`.
- (b) **Subject↔payload binding:** new `subject_authorizes_envelope` (nats.rs:115) is checked in `bridge_once` before `forward_to_broadcast`; forged messages are dropped + counted (`aa_op_control_bridge_rejected_total`). Logic is airtight for **both** branches: a `global` payload is only honored on `GLOBAL_SUBJECT` (`assembly.opcontrol.global`, 3 tokens), and a per-agent payload must match `subject_for()` (`assembly.opcontrol.<tenant>.<agent>`, 4 tokens). The two namespaces can never collide, so neither a non-global-on-global nor a global-on-tenant smuggle is possible.
- (c) **`op_control_stream` authZ:** reads the `enrich`-injected `VerifiedCaller` from request extensions; when `self.registry.is_some()` it requires the caller (absent → `Unauthenticated`) and requires `proto_agent_id_to_key(&agent_id) == caller.agent_key` (mismatch → `PermissionDenied`). `caller.agent_key` is derived from the credential token via the registry reverse-index (`resolve_caller`), so a valid token cannot subscribe to a *different* triple.

### Side-effect analysis
No regression to legitimate behavior:
- Valid-token batch entry is still evaluated/allowed (positive-control test `batch_check_still_evaluates_a_request_with_a_valid_credential_token`).
- NATS auth/TLS fields are optional and default to the prior plaintext/no-auth behavior, so loopback dev is unchanged; legit halts still propagate (e2e delivers the Pause).
- Own-identity subscribe is admitted and receives its halt.
- The `op_control_stream` gate is skipped only when no registry is attached. Verified all three production server paths use `PolicyServiceImpl::with_registry*` + `enrich_interceptor` (server.rs:480/563/567/591), so in a tenanted deployment the gate is always active — the skip is genuinely test/untenanted only, not a prod bypass. Consistent with the existing `validate_credential_token` design.

### Adversarial security check
Attacks I tried to confirm the fix holds:
1. **Forged batch entry (empty / mismatched / cross-identity token):** denied as `a2a_identity_verification` *before* any eval; test asserts the over-budget victim is **not** suspended and **no** tool-call/budget audit is attributed to the victim — only the impersonation audit. ✓
2. **Race/ordering hole in the loop:** confirmed validation is the first statement per iteration; no path reaches `record_*`/`maybe_suspend` for a rejected entry. ✓
3. **Cross-tenant NATS payload forgery (per-agent on attacker subject):** dropped — `subject_for(forged) != delivery_subject`. ✓ (real-NATS JetStream testcontainer e2e `bridge_drops_envelopes_whose_payload_does_not_match_their_subject`).
4. **Fleet-wide kill forgery (`global:true` on a tenant subject):** dropped — global only honored on `GLOBAL_SUBJECT`; token-count mismatch makes collision impossible. ✓
5. **Per-agent payload smuggled onto the global subject:** dropped. ✓
6. **Cross-tenant op-signal read (`op_control_stream` with a valid token but a victim triple):** `PermissionDenied` because `agent_key` is token-derived, not request-derived. ✓ Tokenless → `Unauthenticated`. ✓

The new tests genuinely exercise the attacks (forged cred with side-effect assertions, forged subject via direct JetStream publish, cross-tenant triple over the real `enrich` interceptor), including a true real-NATS e2e for the bridge-forgery case — not superficial.

### Verdict
**Approve & ready.** Both HIGH tickets are fully and correctly fixed, all three 3889 sub-parts covered, no side-effects on legitimate flows, CI green, and every adversarial probe I could construct is closed. Flipping Draft → ready.

PR #1316, opened 2026-06-28
[AAASM-3893] 🔒 (policy): Fail-closed on schedule cache + numeric parse

Chisanan232:
## Description

Closes two Low-severity policy-evaluation fail-open residuals in `aa-gateway` (AAASM-3893):

1. **Cached schedule stage → post-window fail-open.** The cascade evaluation path cached the full `merge_decisions` verdict (which ran the time-dependent `stage_schedule` first) in a 60s-TTL `DecisionCache`. An `Allow` computed inside an active-hours window was therefore served for up to ~60s after the window closed. Fix: the schedule stage is removed from the cached merge unit (`evaluate_single_doc`) and run **fresh on every request** via a new `evaluate_schedule_cascade`, outside the cache — alongside the other uncached stateful stages (rate-limit, budget). The non-cascade `evaluate_primary` path already ran its schedule check uncached, so it was unaffected.

2. **Numeric ordered-comparison parse failure → silent no-match.** `policy/expr.rs::eval_string_numeric` returned `false` when an ordered comparison (`> >= < <=`) operand on a generic string field could not be coerced to a number. This evaluator drives `requires_approval_if`, where `false` means "no approval required", so a coercion failure silently let a security guard not-match and the action proceed unguarded. Fix: an unresolvable ordered comparison now **fails closed** (fires the predicate → require approval), consistent with the module's documented fail-safe-true posture.
   - Equality (`==`/`!=`, line 724) is intentionally left unchanged: a non-numeric string is genuinely not equal to a number, so `false` is the correct result there — flipping it would break every legitimate `field ==/!= N` policy. This was verified to not be a genuine fail-open.

## Type of Change

- [x] 🐛 Bug fix (security hardening — fail-closed)

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-3893
- Closes AAASM-3893

## Testing

- `cargo nextest run -p aa-gateway` — 1302 passed, 0 failed.
- New regressions:
  - `engine::tests::cascade_schedule_denies_despite_stale_cached_allow` — pre-seeds the decision cache with an `Allow`, then evaluates against a closed active-hours window; asserts the request is denied (schedule evaluated fresh, not from cache).
  - `policy::expr::tests::ordered_numeric_comparison_on_unparseable_field_fails_closed` — `command > 1000` on a non-numeric field fires (fail-closed); `command == 5` correctly stays false.
- `cargo build -p aa-gateway`, `cargo fmt --all -- --check`, `cargo clippy -p aa-gateway --all-targets -- -D warnings` all clean.

## Checklist

- [x] Documentation updated if behaviour changed (doc-comments on the changed functions)
- [ ] All CI checks passing

### DIFF
--- a/aa-gateway/src/engine/decision.rs
+++ b/aa-gateway/src/engine/decision.rs
@@ -39,11 +39,16 @@ impl PolicyDecision {
 }
 
 /// Evaluate a single `PolicyDocument` against `(ctx, action)` and return the
-/// decision for stages 1–3 and 5 (schedule, network, tool-allow, approval-condition).
+/// decision for the *cacheable* stages 2–3 and 5 (network, tool-allow,
+/// capability, approval-condition).
 ///
-/// Stages 4 (rate-limit) and 7 (budget) are stateful at engine level and are
-/// evaluated separately after merging. Stage 6 (credential scan) does not
-/// produce a decision — it is handled at the engine level.
+/// Stage 1 (schedule) is deliberately **excluded** here: it is time-dependent
+/// and must be evaluated fresh on every request via [`evaluate_schedule_cascade`]
+/// so that a cached `Allow` computed inside an active-hours window cannot be
+/// served after the window closes (AAASM-3893). Stages 4 (rate-limit) and 7
+/// (budget) are stateful at engine level and are evaluated separately after
+/// merging. Stage 6 (credential scan) does not produce a decision — it is
+/// handled at the engine level.
 ///
 /// Returns `Deny` on the first matching deny rule, `RequireApproval` for approval
 /// conditions, and `Allow` if no rule fires.
@@ -53,9 +58,6 @@ pub(crate) fn evaluate_single_doc(
     action: &aa_core::GovernanceAction,
     policy_ctx: Option<&dyn crate::policy::context::PolicyContext>,
 ) -> PolicyDecision {
-    if let Some(d) = stage_schedule(doc) {
-        return d;
-    }
     if let Some(d) = stage_network(doc, action) {
         return d;
     }
@@ -70,6 +72,20 @@ pub(crate) fn evaluate_single_doc(
     }
 
     PolicyDecision::Allow
+}
+
+/// Stage 1 — Schedule, evaluated **fresh** across the whole cascade.
+///
+/// Returns the first schedule-based `Deny` (current time outside a doc's
+/// active-hours window, or an unparseable timezone), or `None` when every doc's
+/// window currently permits the request.
+///
+/// AAASM-3893: this stage is time-dependent and is intentionally kept out of
+/// [`evaluate_single_doc`] (and therefore out of the engine's decision cache).
+/// The caller runs it on every request so a cached verdict computed inside an
+/// active-hours window cannot outlive that window.
+pub(crate) fn evaluate_schedule_cascade(cascade: &[Arc<PolicyDocument>]) -> Option<PolicyDecision> {
+    cascade.iter().find_map(|doc| stage_schedule(doc))
 }
 
 /// Stage 1 — Schedule: deny when the current time is outside the doc's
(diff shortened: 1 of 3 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

Reviewed at HEAD `12021b17`. Read the full cascade flow, both fixed stages, the sole production consumer, and the new tests.

### CI
All required checks **green**: Build, Test (7m56s), Clippy lint, Format check, Rustdoc, Rust conformance, Integration tests (macOS, 9m54s), TimescaleDB, CodeQL/Analyze (rust/go/py/js), Dependency checks, no_std targets. No failures; only path-gated jobs (dashboard, proto, OpenAPI, eBPF) skipped, as expected for a pure `aa-gateway` change.

### Scope coverage
Both genuine gaps are fixed, and only those:

1. **Schedule TTL fail-open** — `stage_schedule` removed from the cached unit `evaluate_single_doc` (`decision.rs:55-72`); `merge_decisions` (the cached verdict, `decision.rs:288`) no longer carries it. New `evaluate_schedule_cascade` (`decision.rs:87`) runs **fresh before the cache lookup** in `evaluate_with_cascade` (`mod.rs:1203-1210`), returning early only on `Deny`; `None` correctly falls through to the cached stateless verdict. This is the right surgical cut — schedule now sits alongside the other uncached stateful stages (4 rate-limit, 7 budget). The `evaluate_primary` (non-cascade) path already ran schedule uncached and is untouched.
2. **Numeric ordered-comparison fail-open** — `eval_string_numeric` parse-failure arm flipped `false → true` (`expr.rs:747-751`). Confirmed the **sole** production caller of `expr::evaluate` is `requires_approval_if` at `mod.rs:961`, where `true → RequiresApproval` — i.e. unambiguously **fail-closed**. There is no allow-style guard consuming this evaluator, so `true` cannot over-allow anywhere.

### Side-effect analysis
- **(a) Cache integrity** — DecisionCache still memoizes stages 2–3 + 5 via `merge_decisions` (`mod.rs:1228-1235`); only schedule was pulled out. No regression to the cached path.
- **(b) Perf** — schedule now runs every request: one `chrono_tz` parse + `Utc::now()` + `HH:MM` string compare per doc. Negligible; correctness over a ≤60s stale-allow window is the whole point. No meaningful regression.
- **(c) No spurious denials** — the `true`-on-fail change fires *only* when an ordered operand can't coerce to a number (a misconfigured/ambiguous `field >/< N` against a non-numeric field). Legitimate numeric policies parse fine and are unaffected; verified by the new test asserting `command == 5` stays `false`. The change is also consistent with the module's documented fail-safe-true posture (an entirely unparseable expr already returns `true`).
- Both early-return `Deny` arms destructure `{ reason, .. }`, dropping `source_scope` exactly as the pre-existing cached path did — no behavioral drift.

### Tests
- `cascade_schedule_denies_despite_stale_cached_allow` (`mod.rs:2141`) — pre-seeds the cache with the stale `Allow`, evaluates an empty window (`start==end=="00:00"`, so `current >= end` is true at every wall-clock time → deterministic), asserts `Deny`. Directly exercises gap (1).
- `ordered_numeric_comparison_on_unparseable_field_fails_closed` (`expr.rs:1308`) — `command > 1000` fires (fail-closed); `command == 5` stays `false`, locking in the equality non-regression. Exercises gap (2).
Both are meaningful behavioral regressions, not internal-state assertions.

### Verdict
**Approve.** Both fail-opens are closed with minimal, well-scoped changes; CI is green; no side effects found. I **explicitly endorse the equality-path non-fix**: `eval_string_equality` (`expr.rs:722-732`, line 724 unchanged) correctly returns `false` for a non-numeric string vs. a numeric literal — a non-numeric string genuinely is not equal to a number, and flipping it to `true` would make every legitimate `field ==/!= N` policy mis-fire. That is correct behavior, not a missed fix. Commit granularity is clean (fix + test split per gap). Flipping to ready-for-review.

PR #1325, opened 2026-06-29
[AAASM-3923] 🔒 (storage): Redact DSN, warn on plaintext transport, tenant-key TODO

Chisanan232:
## What changed

Storage transport / secret hygiene hardening for the `aa-storage-postgres` and `aa-storage-redis` drivers (AAASM-3923, LOW). Three sub-items:

1. **DSN redaction (DONE)** — `PostgresPoolConfig` and `RedisStorageConfig` carry a connection URL with `user:pass@host` credentials. The derived `Debug` printed the password verbatim, and `RedisStorageConfig` additionally derived `Serialize` which round-tripped it in clear (both latent — no live log today). Replaced both `Debug` derives with custom impls that render the password as `***`, and dropped the unused `Serialize` derive on `RedisStorageConfig`. The unredacted `url` is still what the pool connects with — only diagnostic/serialized output changes.

2. **Plaintext-transport warning (DONE)** — at pool build/connect time, emit a `tracing::warn!` when the host is **non-loopback** and TLS is not enforced (Redis: neither `tls = true` nor a `rediss://` scheme; Postgres: no `sslmode=require`/`verify-ca`/`verify-full`). Loopback hosts stay silent. This **does not hard-fail** — operators may have valid reasons to skip TLS; it only surfaces the sniffing risk. The warning decision is extracted into a pure `should_warn_plaintext` predicate so it is unit-testable, and the connection behavior is otherwise unchanged.

3. **Redis L2 tenant key prefix (DEFERRED → AAASM-3919)** — the shared Redis L2 cache keys (`aa:policy:<agent_id>`, `aa:session:<session_id>`, `aa:ratelimit:<key>`) carry no tenant/org prefix, so there is no tenant boundary at the cache layer. The ids are globally unique so there is **no collision today**, but also no isolation. The `PolicyStore`/`SessionStore`/`RateLimitCounter` trait methods carry only an agent/session id (or opaque key) — org context is **not threaded** into these paths, and that threading depends on the org-context work tracked under **AAASM-3919**. Prefixing without the org id would break lookups, so rather than ship a half-measure this PR adds a `TODO(AAASM-3919)` marker at each key builder. This sub-item remains open.

## Why

Hacker-POV: credential-bearing DSNs that round-trip through `Debug`/`Serialize` leak passwords into any future log/serialization; plaintext transport lets a network-adjacent attacker sniff storage traffic; un-namespaced L2 keys weaken cache-layer tenant isolation.

## How to verify

- `cargo nextest run -p aa-storage-postgres -p aa-storage-redis` — 41/41 pass (includes the testcontainers Postgres/Redis integration suites).
- New unit tests: `debug_redacts_dsn_password` / `debug_leaves_passwordless_dsn_untouched` (both crates) assert the password never appears in `Debug` output; `should_warn_plaintext` predicate tests cover loopback-silent, non-loopback-warn, and TLS-in-effect cases.
- `cargo fmt -p aa-storage-postgres -p aa-storage-redis -- --check` and `cargo clippy -p aa-storage-postgres -p aa-storage-redis --all-targets -- -D warnings` are clean.

## Ticket

Refs AAASM-3923 (sub-items 1 and 2 complete; sub-item 3 deferred to AAASM-3919, so the ticket is kept open intentionally).
Refs AAASM-3919 (tenant-key namespacing).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,6 +585,7 @@ dependencies = [
  "testcontainers-modules",
  "tokio",
  "toml 1.1.2+spec-1.1.0",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -566,6 +566,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "toml 1.1.2+spec-1.1.0",
+ "tracing",
  "uuid",
 ]
 
--- a/aa-storage-postgres/Cargo.toml
+++ b/aa-storage-postgres/Cargo.toml
@@ -19,6 +19,7 @@ serde_json = { workspace = true }
 uuid       = { workspace = true, features = ["v4", "serde"] }
 chrono     = { workspace = true, features = ["clock", "serde"] }
 thiserror  = { workspace = true }
+tracing    = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
--- a/aa-storage-postgres/src/config.rs
+++ b/aa-storage-postgres/src/config.rs
@@ -4,6 +4,8 @@
 //! file. Only the knobs an OSS operator needs to point the driver at their own
 //! Postgres are exposed here; TimescaleDB and retention tuning live elsewhere.
 
+use std::fmt;
+
 use serde::Deserialize;
 
 /// Default maximum pooled-connection count when unspecified.
@@ -22,7 +24,7 @@ const DEFAULT_STATEMENT_TIMEOUT_MS: u64 = 0;
 /// max_connections = 20
 /// statement_timeout_ms = 5000
 /// ```
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Deserialize)]
 #[serde(default)]
 pub struct PostgresPoolConfig {
     /// PostgreSQL connection URL (`postgres://user:pass@host:port/db`).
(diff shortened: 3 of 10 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: APPROVE** — sub-items 1 & 2 implemented correctly, sub-item 3 cleanly deferred. CI green. Flipping out of Draft.

### CI
All required checks green ✅. One check, **Standalone build (git-SHA pin)**, was initially red — but the failure was on `aa-proto` with `curl failed … [16] Error in the HTTP2 framing layer` while downloading `multimap` from crates-io: a transient registry/network flake, unrelated to this PR (which touches only `aa-storage-postgres`/`aa-storage-redis`, not `aa-proto`/prost/tonic). Reran the job → **success**. No real check is red. (`cli_dashboard` E2E known flake not present this run.)

Author used `LEFTHOOK=0` on dev commits (pre-commit fmt >2min) but ran gates manually — CI re-validated: **Format check** and **Clippy lint** both pass.

### Scope (2 of 3 sub-items; 3rd deferred)
1. **DSN redaction — DONE, correct.** Custom `Debug` on `PostgresPoolConfig` + `RedisStorageConfig` renders the password as `user:***@host`; derived `Serialize` dropped from `RedisStorageConfig` (no callers). The real `url` / `connection_url()` is still what the pool connects with — diagnostic output only.
2. **Plaintext warning — DONE, correct.** `tracing::warn!` fires only when host is non-loopback **and** TLS is not enforced (Redis: no `tls=true` and no `rediss://`; Postgres: no `sslmode=require|verify-ca|verify-full`). Loopback (`localhost`/`127.0.0.0/8`/`[::1]`) is silent. **No hard-fail** — surfaces risk only. Decision extracted into a pure `should_warn_plaintext` predicate.
3. **Redis L2 tenant prefix — DEFERRED → AAASM-3919, done right.** `TODO(AAASM-3919)` markers added at all three key builders (`policy_key`, `session_key`, `counter_key`). Key builders are **unchanged**, so lookups are not broken — this is a genuine deferral, not a half-measure. Org context isn't threaded into the `PolicyStore`/`SessionStore`/`RateLimitCounter` paths (they carry only an agent/session id or opaque key); prefixing without the org id would break lookups. Ticket kept open via `Refs` (not `Closes`). ✅

### Side-effects
- Redaction is confined to `Debug`/serialized output; the `url` field stays unredacted and connections use the real value — **connection behavior unchanged**. ✅
- Warning fires only in the non-loopback + no-TLS case; loopback silent; no behavioral gate added. ✅
- Minor (non-blocking): `redact_dsn_password` is a best-effort string split on the first `@`; a password containing a literal un-encoded `@` could be split imperfectly. URL-spec passwords should be percent-encoded, so this is acceptable for a LOW-severity hardening pass.

### Tests
- `debug_redacts_dsn_password` + `debug_leaves_passwordless_dsn_untouched` (both crates) — assert the password never appears in `Debug` output. ✅
- `should_warn_plaintext` predicate tests — loopback-silent, non-loopback-warn, TLS-in-effect, and weak-`sslmode`-still-warns cases. ✅

**Do not merge** (per request) — review only; flipped to Ready for the Pioneer team to action.

PR #1327, opened 2026-06-29
[AAASM-3917] 🔒 (aa-ebpf): Make bytecode integrity check meaningful (or accurate doc)

Chisanan232:
## What changed

Corrects the overstated security claims around the eBPF load-time bytecode integrity check (`aa-ebpf/src/integrity.rs`, `aa-ebpf/build.rs`, `aa-ebpf/src/error.rs`). The docs are reworded to accurately describe what the check provides; an explicit `TODO(AAASM-3601)` marks the path to a real supply-chain guarantee. No runtime logic changed.

## Why (the bug)

`verify_bytecode` recomputes the SHA-256 of the embedded probe bytes and compares it to a baked-in constant. That constant is emitted by `build.rs::emit_object_digest`, which hashes the **same** compiled object that `lib.rs` then embeds. The comparison is therefore self-referential (a tautology): an attacker who tampers with the probe source or the compiled `.o` before the build moves both the embedded bytes **and** the baked digest together, and the check still passes. The module doc nonetheless claimed it was "the last line of defense against the swapped-`.o` supply-chain attack" — false assurance.

## Which fix path, and why

**The honest-doc path.** I investigated whether the independently-produced, cosign-signed `EBPF_SHA256SUMS` referenced by AAASM-3601 could anchor the digest:

- It **does** exist — `release.yml` builds the probe objects, writes `EBPF_SHA256SUMS`, and cosign-signs it (keyless OIDC) at publish time.
- **But** it is generated by hashing the *same* compiled objects in the *same* CI run and signed afterward; it is **not checked into the tree** and is **not consumed by `build.rs`** (which cannot reference it — signing happens after the build that embeds the bytecode). It serves *download-time* verification (`cosign verify-blob`), not the *load-time* in-binary check.

So there is no independent root of trust available **to the build** that I could wire in without a larger CI/build change — that is genuinely AAASM-3601's scope. Per the "do not pretend to fix the crypto if the independent root of trust isn't available" guidance, I took the honest path:

- Rewrote the `integrity.rs` module + `verify_bytecode` docs to state the check detects **post-build, in-binary corruption** of the embedded bytecode (and rejects the empty/unverifiable stub), and is explicitly **not** a supply-chain guarantee. Removed the "last line of defense against the swapped-`.o` attack" claim.
- Softened the `IntegrityMismatch` error doc (`error.rs`) and the `build.rs` digest-emit comment with the same accurate framing.
- Left `// TODO(AAASM-3601): ... source the expected digest from an independently-signed SHA256SUMS` in both `integrity.rs` and `build.rs`.
- Renamed the mismatch test to `corrupted_bytecode_is_rejected` and added a tests-module note clarifying the tests assert in-binary corruption detection, not a supply-chain guarantee.

## How verified

eBPF probes compile Linux-only; on macOS the userspace `aa-ebpf` crate (which contains `integrity.rs`/`error.rs`) builds host-side. Validated locally:

- `cargo check -p aa-ebpf` — clean
- `cargo nextest run -p aa-ebpf integrity` — 4/4 pass
- `cargo clippy -p aa-ebpf --all-targets -- -D warnings` — exit 0
- lefthook pre-commit (full-workspace fmt + clippy) — passed on every commit

Full eBPF probe build/runtime (kernel load path) is **deferred to Linux CI**.

This is a doc/accuracy fix, so no runtime behavior changed.

Closes AAASM-3917

### DIFF
--- a/aa-ebpf/build.rs
+++ b/aa-ebpf/build.rs
@@ -112,12 +112,18 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
 
         // AAASM-3602: emit the sha256 of each embedded probe object as a
-        // compile-time env var so the load-time integrity check can pin the
-        // bytecode against the digest CI signed (EBPF_SHA256SUMS, AAASM-3601).
-        // The digest is sourced from the *actual compiled object*, never
-        // hand-written. A stub (empty file) hashes to the well-known empty
-        // digest, which the runtime treats as "unverifiable stub" and refuses
-        // to load — fail-closed, never degrade-to-allow.
+        // compile-time env var so the load-time integrity check can detect
+        // post-build, in-binary corruption of the embedded bytecode. The digest
+        // is sourced from the *same* compiled object that is embedded, so it is
+        // NOT an independent supply-chain anchor (see aa-ebpf/src/integrity.rs):
+        // tampering before the build moves both the bytes and this digest
+        // together. The cosign-signed EBPF_SHA256SUMS (AAASM-3601) is a separate
+        // download-time artifact and is not consumed here.
+        // TODO(AAASM-3601): bake the digest from an independently-signed
+        // SHA256SUMS to make the load-time check a real supply-chain guarantee.
+        // A stub (empty file) hashes to the well-known empty digest, which the
+        // runtime treats as "unverifiable stub" and refuses to load —
+        // fail-closed, never degrade-to-allow.
         emit_object_digest(&release_dir, "aa-file-io", "AA_FILE_IO_BPF_SHA256")?;
         emit_object_digest(&release_dir, "aa-exec-probes", "AA_EXEC_BPF_SHA256")?;
         emit_object_digest(&release_dir, "aa-tls-probes", "AA_TLS_BPF_SHA256")?;
--- a/aa-ebpf/src/error.rs
+++ b/aa-ebpf/src/error.rs
@@ -59,11 +59,13 @@ pub enum EbpfError {
         detail: String,
     },
 
-    /// The embedded BPF bytecode did not match the digest CI signed
-    /// (AAASM-3602). This is a hard, fail-closed error: a probe whose bytecode
-    /// has been tampered with — or is an empty stub — is refused, never loaded
-    /// blindly. Closes the "loads and emits plausible telemetry but is
-    /// deliberately blind" supply-chain scenario.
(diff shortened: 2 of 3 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**CI**
- All required checks green after a rerun. `Rustdoc` + `CI Success` (the aggregate gate) initially failed on a transient crates.io network flake — `curl failed ... [18] Transferred a partial file` while downloading `chrono`, unrelated to this PR (which doesn't touch `aa-core`, the crate rustdoc builds). Re-ran the failed jobs; both now pass. `cli_dashboard` E2E load-flake not in play. No real red check.

**Scope**
- Doc/test-only, as advertised — `verify_bytecode` runtime logic is unchanged (only a doc comment added above the fn).
- `integrity.rs`: module docs rewritten with an explicit "What this check does — and does NOT — guarantee" section; the overstated "last line of defense against the swapped-`.o` supply-chain attack" claim is **removed** and replaced with the accurate framing — the baked digest is derived from the *same* embedded object, so the comparison is self-referential and is **not** a supply-chain guarantee. What it does catch (post-build, in-binary corruption; empty/unverifiable stub) is stated precisely. `TODO(AAASM-3601)` present.
- `error.rs`: `IntegrityMismatch` doc softened to "post-build corruption only," cross-referencing the integrity module / TODO.
- `build.rs`: digest-emit comment corrected + `TODO(AAASM-3601)` added.
- Test renamed `mismatched_digest_is_rejected` → `corrupted_bytecode_is_rejected` with the same assertion (digest mismatch → reject); tests-module note clarifies the suite asserts in-binary-corruption detection, not a supply-chain guarantee. No false claim remains.

**Investigation soundness**
- Confirmed sound. `build.rs::emit_object_digest` hashes the same `.o` that `lib.rs` embeds, making the load-time check a tautology against pre-build tampering. The cosign-signed `EBPF_SHA256SUMS` (AAASM-3601) is produced *post-build* at release time and is not — and cannot be — consumed by `build.rs`, so no independent root of trust is available to the load-time check. Wiring one in is genuinely AAASM-3601's scope.

**Verdict: APPROVE — honest-doc path.** Correcting overstated security claims and marking the real fix with `TODO(AAASM-3601)` is the right call over pretending the crypto guarantee exists. No runtime behavior changed; no side effects. Flipping Draft → Ready.

PR #1330, opened 2026-06-29
[AAASM-3916] 🔒 (aa-ebpf): Propagate syscall-guard to descendants + path/ppid probe correctness

Chisanan232:
## What changed

Fixes two related eBPF-probe correctness/security bugs in one PR (they share a `sched_process_fork` tracepoint and overlapping files in `aa-ebpf-probes` and `aa-ebpf`, so combining avoids a merge conflict).

### AAASM-3916 [High] — syscall_guard confined only the single target PID; descendants escaped
- `aa-ebpf-probes/src/syscall_guard.rs`: add an `aa_syscall_guard_fork` tracepoint at `sched/sched_process_fork`. When the parent tgid is in `PID_FILTER`, it mirrors that membership onto the forked child so the child is confined from its first syscall. `SYSCALL_ALLOWLIST` is global (keyed by syscall number, not per-PID), so only `PID_FILTER` membership needs propagation.
- `aa-ebpf/src/loader.rs`: `SyscallGuardLoader::attach` now also loads+attaches the new fork tracepoint.
- Documented fork-vs-thread and attach-time TOCTOU caveats; corrected the `FileIoLoader` doc that over-claimed descendant monitoring (the file-I/O probe has no fork propagation — noted as a follow-up).

### AAASM-3921 [Med] — path match non-canonical + prefix-no-op + exec ppid wrong
- **(a) non-canonical paths** — added `canonicalize_lexical()` (pure userspace helper) and applied it in `SensitivePathDetector` to both stored prefixes and incoming event paths, so the OBSERVE-ONLY alerting layer now catches `/etc//shadow`, `/etc/./shadow`, `/etc/../etc/shadow`, etc. Directory-prefix boundaries preserved (`/home/` ≠ `/homestead`). **Lexical only** — symlink/CWD/mount resolution still needs an in-kernel `d_path` walk (deferred; needs Linux). Done & host-tested.
- **(b) prefix-no-op** — chose the lower-risk option: corrected the in-kernel `PATH_BLOCKLIST`/`PATH_ALLOWLIST` and `PathPattern` docs from the false "hash of path prefix" / "prefix or exact" to the actual **exact NUL-padded full-path equality** contract, and reconciled directory/prefix/canonical rules onto the userspace detector. In-kernel bounded prefix matching deferred (needs Linux). Done (docs/contract) + reconciled.
- **(c) exec ppid wrong** — `sched_process_exec`'s `pid` field is the new process's own pid, so `ppid == pid` broke `ProcessLineageTracker::is_descendant_of` (self-cycle → always false). Added `handle_sched_process_fork` to `exec_probes.rs` recording child→parent tgid in `PARENT_TGID` (plus `EXEC_PID_FILTER` descendant propagation); the exec handler now resolves the real ppid from it, falling back to `0` (unknown root). Wired into `TracepointManager`. Exit handler cleans the maps. Done (latent path; kernel data now correct).

## Why
Descendants of a confined process ran fully UNCONFINED (High); sensitive-path alerting was defeated by trivial non-canonical paths and directory rules silently never fired; and process-lineage attribution was structurally broken. See AAASM-3916 / AAASM-3921.

## Validation

> ⚠️ **eBPF is Linux-only and was changed on macOS — kernel runtime behavior is NOT validated locally and MUST be validated on Linux CI before merge.** This is why the PR is a Draft.

**Validated locally (macOS):**
- `cargo check -p aa-ebpf` ✅ and `rustup run nightly cargo check --release` of the standalone `aa-ebpf-probes` workspace (bpfel target, build-std) ✅ — the BPF probe code type-checks/compiles, but the **verifier does not run** off-Linux.
- `cargo fmt --all -- --check` ✅
- `cargo clippy -p aa-ebpf -p aa-ebpf-common --all-targets -- -D warnings` ✅ (no lint warnings)
- `cargo nextest run -p aa-ebpf -p aa-ebpf-common` ✅ 106 passed — includes the new `canonicalize_lexical` + detector tests and the existing lineage descendant tests.

**MUST be validated on Linux CI before merge (changed blind on macOS):**
- BPF **verifier acceptance** of the two new `sched_process_fork` tracepoint programs (map writes, `read_at` offsets).
- Correctness of the `sched_process_fork` field offset (`child_pid` @ 44) and that fork fires before exec at runtime.
- Runtime confinement of descendants by the syscall guard; real-ppid population for exec events (`aa-ebpf` integration tests `exec_tracepoint.rs` / `e2e_ebpf.rs`, run as root with `--features integration-test`).

Closes AAASM-3916
Closes AAASM-3921

### DIFF
--- a/aa-ebpf-probes/src/exec_probes.rs
+++ b/aa-ebpf-probes/src/exec_probes.rs
@@ -219,7 +219,21 @@ fn try_sched_process_exit(_ctx: &TracePointContext) -> Result<u32, i64> {
     let pid_tgid = bpf_get_current_pid_tgid();
     let tgid = (pid_tgid >> 32) as u32;
 
-    if !pid_allowed(tgid) {
+    // Capture monitoring status BEFORE cleanup (cleanup may remove this tgid's
+    // own filter entry, which `pid_allowed` would otherwise consult).
+    let allowed = pid_allowed(tgid);
+
+    // Unconditionally release this tgid's parent-cache and propagated-filter
+    // entries (AAASM-3921c). The fork handler records `PARENT_TGID` for *every*
+    // fork (so the exec handler can always resolve the real ppid), so the
+    // exit-side cleanup must also be unconditional — otherwise the map would
+    // grow without bound and reused pids could inherit a stale parent / filter
+    // membership. The wildcard key (0) is never an exiting tgid.
+    let _ = PARENT_TGID.remove(&tgid);
+    let _ = EXEC_PID_FILTER.remove(&tgid);
+
+    // Only emit the exit event for monitored processes (lineage cleanup).
+    if !allowed {
         return Ok(0);
     }
 
@@ -233,13 +247,6 @@ fn try_sched_process_exit(_ctx: &TracePointContext) -> Result<u32, i64> {
     }
 
     entry.submit(0);
-
-    // Bound the fork-propagated maps and avoid pid-reuse staleness: drop this
-    // tgid's parent record and its propagated filter membership on exit
-    // (AAASM-3921c). The original target PID re-added by userspace is unaffected
-    // on the next load; the wildcard key (0) is never an exiting tgid.
-    let _ = PARENT_TGID.remove(&tgid);
-    let _ = EXEC_PID_FILTER.remove(&tgid);
     Ok(0)
 }
 
--- a/aa-ebpf-probes/src/exec_probes.rs
+++ b/aa-ebpf-probes/src/exec_probes.rs
@@ -22,10 +22,19 @@
 //! pid, not its parent. Reading it into `ppid` produced `ppid == pid`, which
 //! made the userspace `ProcessLineageTracker::is_descendant_of` walk terminate
 //! immediately (it treats `ppid == pid` as a self-cycle) and always return
-//! false. The real parent is captured at `sched_process_fork` time (the
-//! tracepoint carries `parent_pid`/`child_pid`) and stashed in `PARENT_TGID`;
-//! the exec handler now reads `ppid` from there, falling back to `0` (unknown
-//! root) when the fork was not observed — never the bogus self-parent.
+//! false.
+//!
+//! The exec handler now resolves the real parent with a CO-RE read of
+//! `current->real_parent->tgid` at exec time (see [`real_parent_tgid`]), using
+//! `task_struct` field byte-offsets that the userspace loader resolves from the
+//! running kernel's BTF and publishes in [`TASK_OFFSETS`]. This is independent
+//! of the fork tracepoint: the earlier map-based path (reading `PARENT_TGID`,
+//! populated at `sched_process_fork`) proved unreliable under load — the map
+//! could fill or the specific child entry be missing — leaving `ppid == 0`.
+//! `PARENT_TGID` is kept only as a fallback for kernels without BTF; the final
+//! fallback is `0` (unknown root), never the bogus self-parent. The fork
+//! tracepoint is retained purely for AAASM-3916 descendant confinement (the
+//! `EXEC_PID_FILTER` propagation), decoupled from exec-ppid resolution.
 //!
 //! ## Stack-limit workaround
 //!
@@ -38,9 +47,12 @@
 
 use aa_ebpf_common::exec::{ExecEvent, ProcessExitEvent, MAX_ARGS_LEN, MAX_FILENAME_LEN};
 use aya_ebpf::{
-    helpers::{bpf_get_current_pid_tgid, bpf_get_current_uid_gid, bpf_ktime_get_ns, bpf_probe_read_kernel_str_bytes},
+    helpers::{
+        bpf_get_current_pid_tgid, bpf_get_current_task, bpf_get_current_uid_gid, bpf_ktime_get_ns,
+        bpf_probe_read_kernel, bpf_probe_read_kernel_str_bytes,
+    },
     macros::{map, tracepoint},
-    maps::{HashMap, RingBuf},
+    maps::{Array, HashMap, RingBuf},
     programs::TracePointContext,
     EbpfContext,
 };
@@ -69,6 +81,26 @@ static EXEC_PID_FILTER: HashMap<u32, u8> = HashMap::with_max_entries(256, 0);
 /// the `sched_process_exec` tracepoint only exposes the new process's own pid.
 #[map]
 static PARENT_TGID: HashMap<u32, u32> = HashMap::with_max_entries(1024, 0);
+
+/// `task_struct` field byte-offsets, resolved from the running kernel's BTF by
+/// the userspace loader (AAASM-3921c CO-RE). The `sched_process_exec`
+/// tracepoint only carries the new process's own pid, and the fork-populated
+/// `PARENT_TGID` map is unreliable under load (it fills / races), so the exec
+/// handler reads the real parent directly from `current->real_parent->tgid` at
+/// exec time using these offsets.
+///
+/// - index [`TASK_OFFSET_REAL_PARENT`] — byte offset of `task_struct.real_parent`
+/// - index [`TASK_OFFSET_TGID`] — byte offset of `task_struct.tgid`
+///
+/// A value of `0` means "unresolved" (BTF unavailable); the handler then falls
+/// back to the fork-populated `PARENT_TGID` map, and finally to `0`.
+#[map]
+static TASK_OFFSETS: Array<u32> = Array::with_max_entries(2, 0);
+
+/// `TASK_OFFSETS` index of the `task_struct.real_parent` byte offset.
+const TASK_OFFSET_REAL_PARENT: u32 = 0;
+/// `TASK_OFFSETS` index of the `task_struct.tgid` byte offset.
+const TASK_OFFSET_TGID: u32 = 1;
 
 /// `sched:sched_process_fork` layout — byte offset of the `child_pid` field:
 ///
@@ -101,6 +133,41 @@ fn pid_allowed(tgid: u32) -> bool {
             return true;
         }
         EXEC_PID_FILTER.get(&tgid).is_some()
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Real-parent resolution (AAASM-3921c CO-RE)
+// ---------------------------------------------------------------------------
+
+/// Resolve the real parent tgid of the current task at exec time by walking
+/// `current->real_parent->tgid` with BTF-resolved byte offsets.
+///
+/// Returns `None` when the offsets are unresolved (BTF unavailable — both
+/// offsets left `0` by the loader) or when either kernel read fails, so the
+/// caller can fall back to the fork-populated `PARENT_TGID` map. This does not
+/// depend on the fork tracepoint having observed the fork, which is why it is
+/// robust where the map-based path was not.
+#[inline(always)]
+fn real_parent_tgid() -> Option<u32> {
+    let real_parent_off = *TASK_OFFSETS.get(TASK_OFFSET_REAL_PARENT)? as usize;
+    let tgid_off = *TASK_OFFSETS.get(TASK_OFFSET_TGID)? as usize;
+    if real_parent_off == 0 || tgid_off == 0 {
+        return None;
+    }
+
+    unsafe {
+        let task = bpf_get_current_task() as *const u8;
+        if task.is_null() {
+            return None;
+        }
+        // task->real_parent is a `struct task_struct *`.
+        let parent = bpf_probe_read_kernel(task.add(real_parent_off) as *const *const u8).ok()?;
+        if parent.is_null() {
+            return None;
+        }
+        // real_parent->tgid is a `pid_t` (i32); read as u32.
+        bpf_probe_read_kernel(parent.add(tgid_off) as *const u32).ok()
     }
 }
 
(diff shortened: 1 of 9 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

Adversarial review of the **PR-head diff**. This PR changes **kernel eBPF code that could only be compile-checked on macOS**, so Linux CI (eBPF build + BPF verifier + integration/e2e) is the real validation gate. **That gate is RED.** Leaving Draft.

### CI status (per-job — Linux/eBPF gate)
- ✅ `eBPF probes build (Linux nightly)` — **pass** (BPF verifier accepted both new `sched_process_fork` programs; map writes / `read_at` offsets verify).
- ❌ **`e2e — Layer 3 eBPF (Linux)` — FAIL** (the runtime kernel-behavior gate):

```
thread 'ebpf_exec_probe_captures_subprocess_spawn' panicked at aa-integration-tests/tests/e2e_ebpf.rs:288:5:
exec event ppid should be set (got 0; test pid is 23520)
test result: FAILED. 5 passed; 1 failed
```

- ✅ Everything else green: Linux builds (x86_64 + aarch64 unknown-linux-gnu), Rust conformance, no_std (thumbv7em + wasm32), Rustdoc, Format, Dependency checks, Compat-matrix, all CodeQL/Analyze.
- This is **not** the known `cli_dashboard` E2E load-flake — it is the eBPF e2e suite failing on the exact behavior this PR introduces.

### Scope
All sub-parts present and on-contract: 3916 fork→PID_FILTER propagation + loader attach; 3921a `canonicalize_lexical` in `SensitivePathDetector` (+ thorough host tests, 106 pass); 3921b PATH_BLOCKLIST/PathPattern docs corrected to exact-match reality; 3921c `PARENT_TGID` fork map + exec real-ppid + exit cleanup + `TracepointManager` wiring. `FileIoLoader` descendant over-claim corrected. ✅ Scope complete.

### Adversarial findings

**1. [BLOCKER] AAASM-3921c real-ppid resolution does not work at runtime — regresses a pre-existing test.**
`e2e_ebpf.rs:288` is **not modified by this PR**. Pre-PR the exec handler set `ppid = tp_pid` (the new process's own pid — wrong, but non-zero, so `ev.ppid > 0` passed). This PR changes `ppid` to `PARENT_TGID.get(&tgid).unwrap_or(0)`. At runtime the lookup **missed → `ppid = 0`**, so the test now fails. The fork tracepoint did not have the child→parent mapping populated by the time the exec handler ran for a `Command::spawn()` child under the wildcard (`EXEC_PID_FILTER[0]`) monitor-all filter. The verifier accepting the program does not prove the `child_pid @ 44` offset or the fork→exec ordering is correct on the runner kernel — and the empirical result says one of them isn't. **The headline fix is unproven/broken; do not flip to Ready.**

Candidate root causes for the author to confirm on Linux: (a) `SCHED_FORK_CHILD_PID_OFFSET = 44` wrong for the runner's `sched_process_fork` layout → `PARENT_TGID` keyed by garbage → exec lookup misses; (b) fork not observed for the spawn path (clone/vfork variant) before exec; (c) fork attached but firing after the relevant exec. The fallback-to-0 masks the failure into a silent wrong value rather than surfacing it.

**2. [Med] `syscall_guard` PID_FILTER has no exit cleanup — asymmetric with the exec side.**
`aa_syscall_guard_fork` does `PID_FILTER.insert(&child_pid, …)` (cap **1024**) but `syscall_guard.rs` has **no `sched_process_exit` handler** and never removes child entries. The exec probe explicitly added `PARENT_TGID.remove` + `EXEC_PID_FILTER.remove` on exit for exactly this pid-reuse-staleness reason — the **enforcement** path (where the consequence is `bpf_send_signal(SIGKILL)`) did not get the symmetric cleanup. Two consequences:
- **Map exhaustion → fail-open:** a fork-heavy confined tree fills the 1024-entry map; further descendants fail to insert and run **unconfined** — defeating the very AAASM-3916 High under the workload that needs it most.
- **PID-reuse → over-confinement:** a dead child's tgid lingers; when the kernel reuses that tgid for an unrelated, non-monitored process, the guard confines it and may **SIGKILL an unrelated process**.

Not a blocker on its own, but it should be fixed (mirror the exec-side exit cleanup for `PID_FILTER`) or filed as a follow-up before this lands.

**Positives:** fork handlers correctly gate on the parent being monitored (`PID_FILTER.get(parent_tgid)` / `pid_allowed(parent_tgid)`) — no blanket confinement of unrelated trees. `canonicalize_lexical` is correct and well-tested (slash-collapse, dot/dot-dot, root-escape clamp, directory-boundary `/home/`≠`/homestead`, relative leading `..`). Fork-attached-first ordering and fork-vs-thread/TOCTOU caveats are documented. The userspace path-detector change preserves existing alert semantics (OR of kernel exact-match + userspace canonical-prefix) with no regression for non-monitored processes.

### Linux-CI-gate status
**RED** — `e2e — Layer 3 eBPF (Linux)` failed on the AAASM-3921c behavior. This is the unproven-kernel-code gate; it must be green before Ready.

### Verdict
**NOT READY — stays Draft.** Scope is complete and the verifier/build pass, but the runtime eBPF e2e gate fails on the core AAASM-3921c fix (`ppid` resolves to `0`, regressing a previously-passing test), and the enforcement-path `PID_FILTER` lacks exit cleanup (Med). Fix the real-ppid resolution so `e2e_ebpf.rs::ebpf_exec_probe_captures_subprocess_spawn` passes on Linux, address the `PID_FILTER` cleanup asymmetry, then re-request review.

Chisanan232:
## 🤖 Claude Code — PR Re-review (final)

Re-reviewed the two pushed fixes against the PR-head diff, with Linux CI green. **Both original blockers are genuinely fixed**, but adversarial review of the *new* exit-cleanup code surfaced a **newly-introduced fail-open** in the same code path. Keeping this **Draft**.

### ✅ Blocker 1 — exec ppid via CO-RE `real_parent` (AAASM-3916): FIXED, sound
- `real_parent_tgid()` walks `current->real_parent->tgid` using BTF-resolved byte offsets published into `TASK_OFFSETS` by the loader. Reads go through `bpf_probe_read_kernel` (bounded, verifier-safe — no raw deref), with null-checks on both the task and the parent pointer, and `0`/unresolved offsets short-circuit to `None`. `real_parent` (not `parent`) is the correct field for true fork lineage.
- Fallback order is correct: CO-RE read → fork-populated `PARENT_TGID` → `0` (unknown root), never the bogus self-parent. Decoupled from the fork tracepoint, as intended.
- BTF walker (`parse_task_offsets`) is bounds-checked (`checked_add`, `.get`), endianness-aware, masks the bitfield high byte on member offsets, and is host-unit-tested. Loader publish is best-effort/non-fatal.
- **Runtime evidence:** `e2e — Layer 3 eBPF (Linux)` GREEN (11m40s) + `eBPF probes build (Linux nightly)` GREEN (7m20s, verifier accepted the CO-RE reads).

### ✅ Blocker 2 — exit cleanup exists + wired: FIXED (but see 🔴 below)
- `aa_syscall_guard_exit` @ `sched/sched_process_exit` removes the tgid from `PID_FILTER`; wired into `SyscallGuardLoader::attach` alongside the enforcement + fork tracepoints. Exec side symmetrically drops `PARENT_TGID` + `EXEC_PID_FILTER` and captures `allowed` before cleanup so the lineage exit event still emits. Map-exhaustion + pid-reuse-SIGKILL for the **single-threaded** case is closed.

### 🔴 NEW blocker — thread exit evicts a still-live multithreaded process (fail-open)
`sched_process_exit` fires **per task (per thread)**, not once per process. Both exit handlers remove by `tgid = pid_tgid >> 32` with **no `pid == tgid` (thread-group-leader) guard**:
- `syscall_guard.rs::try_exit_cleanup` → `PID_FILTER.remove(&tgid)`
- `exec_probes.rs::try_sched_process_exit` → `EXEC_PID_FILTER.remove(&tgid)` / `PARENT_TGID.remove(&tgid)`

For a **multithreaded** confined process (tgid 100, threads 100/101/102), when non-leader thread 101 exits, `bpf_get_current_pid_tgid() >> 32 == 100` → it removes the **live** process's `PID_FILTER[100]` entry. Consequences:
1. **Fail-open:** the whole process (and remaining threads) immediately runs **unconfined** — the enforcement tracepoint early-returns on `PID_FILTER.get(tgid).is_none()`. This defeats the core AAASM-3916 post-escape guarantee.
2. **Descendant propagation also dies:** subsequent `try_fork_propagate` sees `PID_FILTER.get(parent_tgid).is_none()` → returns early → no further children confined.
3. **This regresses the prior state** — before any cleanup, a multithreaded process's tgid key was *never* removed, so it stayed confined (the old bug was only unbounded growth + pid reuse). The fix trades that for a fail-open in a common case.
4. **Internal contradiction:** the PR's own added docs state "threads are covered by the parent's existing [tgid] membership" — but the exit handler *removes* exactly that shared membership on any thread exit.
5. Minor: fork inserts `child_pid` (= new **tid** for `CLONE_THREAD`) into the filter; exit removes by **tgid**, so those per-thread keys are never reclaimed → partial leak persists for thread-heavy workloads.

This matters concretely here because monitored agents are typically multithreaded (Go runtime creates/destroys OS threads; Node libuv threadpool; Python threadpools) — a thread exit is routine, not exotic.

**Suggested fix:** gate both exit handlers on whole-process exit before removing, e.g.
```rust
let pid_tgid = bpf_get_current_pid_tgid();
let tgid = (pid_tgid >> 32) as u32;
if (pid_tgid as u32) != tgid { return Ok(()); } // ignore non-leader thread exits
```
(and consider whether the per-thread tid keys inserted at fork need their own reclaim, or simply don't insert tid keys for `CLONE_THREAD`.) The `e2e — Layer 3 eBPF (Linux)` job passing does not cover this — it exercises single-threaded fork/exec descendants, not thread-exit eviction.

### Scope check
All prior-approved parts intact: `canonicalize_lexical` + `SensitivePathDetector` (3921a, boundary-aware, well-tested), exact-match contract doc corrections (3921a/b), `FileIoLoader` descendant over-claim doc fix, `btf_offsets` module. 3916 fork propagation + 3921c CO-RE ppid present.

### CI per-job (run 28556222045)
| Check | Status |
|---|---|
| `e2e — Layer 3 eBPF (Linux)` | ✅ pass (11m40s) |
| `eBPF probes build (Linux nightly)` | ✅ pass (7m20s) |
| Build (all 4 targets + workspace) | ✅ pass |
| Clippy lint | ✅ pass |
| CodeQL / Analyze (rust/go/py/js) | ✅ pass |
| Rust conformance / Rustdoc / Format / Deny | ✅ pass |
| **Test** | ⏳ **pending** (still running) |

### Verdict — ❌ do NOT flip to ready (staying Draft)
The two original blockers are correctly fixed and the CO-RE ppid work is solid, but the exit-cleanup fix introduces a new fail-open for multithreaded confined processes (missing `pid == tgid` whole-process guard). Because this defeats the primary AAASM-3916 confinement guarantee for a common (multithreaded-agent) case, it needs a follow-up fix before this can go ready. (`Test` is also still pending, but the blocker above is the gating reason.)

Chisanan232:
## 🤖 Claude Code — PR Re-review (final)

3rd-iteration final state re-reviewed against PR head `23e2fc3`. **All three blockers now fixed.**

### Blockers — all resolved
1. **ppid CO-RE** ✅ — exec handler resolves the real parent via `current->real_parent->tgid` using BTF-published `TASK_OFFSETS`, with `PARENT_TGID` fork-map fallback and final `0` (never the bogus self-parent `ppid == pid`).
2. **Exit-cleanup map exhaustion** ✅ — `PID_FILTER` / `EXEC_PID_FILTER` / `PARENT_TGID` entries released on process exit, bounding the maps (cap 1024) and closing the pid-reuse SIGKILL hole.
3. **Per-thread exit fail-open (thread-leader guard)** ✅ — the new fix. Both exit handlers now early-return on a non-leader **thread** exit and only clean up on **whole-process (leader) exit**:
   - `aa-ebpf-probes/src/exec_probes.rs::try_sched_process_exit`: `tgid = (pid_tgid >> 32) as u32`, `pid = pid_tgid as u32`, `if pid != tgid { return Ok(0); }` **before** removing `PARENT_TGID[tgid]` / `EXEC_PID_FILTER[tgid]`; monitoring status captured before cleanup; leader path still removes entries (no leak reintroduced).
   - `aa-ebpf-probes/src/syscall_guard.rs::try_exit_cleanup`: same `>>32` / `as u32` derivation, `if pid != tgid { return Ok(()); }` **before** `PID_FILTER.remove(&tgid)`.

   Shift/mask correct (tgid = high 32, tid = low 32), no off-by-one. A non-leader thread exit of a multithreaded confined process (Go/Node/Python agents) no longer drops confinement. Doc/comments on both handlers updated to reflect leader-gated cleanup; the fork-side "threads covered by parent tgid membership" note remains correct (lookups are by tgid).

Prior-approved parts intact: CO-RE `real_parent` ppid + BTF offsets, PID_FILTER exit cleanup, `canonicalize_lexical`, exact-match contract docs, FileIoLoader doc.

### CI — GREEN (per job)
| Job | Result |
|---|---|
| e2e — Layer 3 eBPF (Linux) | ✅ success |
| eBPF probes build (Linux nightly) | ✅ success |
| Test | ✅ success |
| Build (x86_64/aarch64 linux + darwin) | ✅ success (all 4) |
| Clippy lint | ✅ success |
| Format check | ✅ success |
| CodeQL / Analyze (rust/go/py/js-ts) | ✅ success |
| Rust conformance / Rustdoc | ✅ success |
| CI Success (aggregate) | ✅ success |

`gh pr checks`: 29 passed, 0 failed. Path-gated jobs (dashboard, openapi, coverage, sonar) correctly skipped. No `cli_dashboard` flake this run.

### Verdict
**APPROVE.** All three eBPF probe-correctness blockers fixed, leader-guard correct on both exit handlers, full CI green. Flipping Draft → ready.

PR #1362, opened 2026-07-02
[AAASM-3985] 🔒 (aa-cache): Fix lost-invalidation race (revoked policy served full TTL)

Chisanan232:
## What changed

Fixes a TOCTOU lost-invalidation race in `aa-cache/src/l1.rs`. On a cache miss the leader reads a value from the store; if a concurrent push `invalidate(key)` lands during that load window it finds no entry (`entries.remove` is a no-op) and is silently lost — the leader then unconditionally caches the just-loaded (now-stale) value, so a **revoked policy is served for a full TTL**.

The fix adds a monotonic invalidation epoch (`AtomicU64`):
- `invalidate()` bumps the epoch **before** removing the entry.
- The leader snapshots the epoch **before** `load`, then commits its insert under the key's `DashMap` shard `entry` guard **only if** the epoch is unchanged.

Because `invalidate`'s `remove` contends for the same shard lock as the leader's guarded insert, the bump-then-remove is ordered against the check-then-insert: either the leader observes the bumped epoch and skips the insert, or the remove runs after the insert and drops it. The eviction can no longer be lost, and single-flight/stampede behavior is unchanged.

## Why

Severity Medium — silent loss of a policy revocation lets a denied agent keep operating under a stale ALLOW until the TTL elapses.

## How to verify

- New regression test `l1::tests::invalidate_during_load_is_not_lost` holds a leader inside `load` via an inner delay, fires `invalidate()` mid-load, and asserts the value is **not** cached (`len == 0`) so a subsequent get reloads. Fails against pre-fix code.
- `cargo nextest run -p aa-cache` — 5/5 pass.
- `cargo fmt --all -- --check`, `cargo clippy -p aa-cache --all-targets --all-features -- -D warnings`, `cargo doc -p aa-cache --no-deps` all clean.

Closes AAASM-3985

### DIFF
--- a/aa-cache/src/l1.rs
+++ b/aa-cache/src/l1.rs
@@ -234,6 +234,45 @@ mod tests {
         assert_eq!(cache.inner().call_count(), 2);
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn invalidate_during_load_is_not_lost() {
+        use std::sync::Arc;
+
+        let id = agent(5);
+        // A 50ms inner delay holds the leader inside `load` long enough for the
+        // racing `invalidate` below to land in the middle of the load window.
+        let store = MemoryPolicyStore::with_policy(id, sample_policy(1)).with_delay(Duration::from_millis(50));
+        let cache = Arc::new(L1Cache::new(store, Duration::from_secs(60)));
+
+        // Leader begins a miss and is now sleeping inside `load`.
+        let leader = {
+            let cache = Arc::clone(&cache);
+            tokio::spawn(async move { cache.get(id).await })
+        };
+
+        // Let the leader enter `load`, then invalidate while it is in flight.
+        // `entries` is still empty, so the old code's `remove` was a silent
+        // no-op and the leader would go on to cache the value it is loading.
+        tokio::time::sleep(Duration::from_millis(20)).await;
+        assert!(!cache.invalidate(&id));
+
+        // The leader still returns the value it loaded, but must NOT have cached
+        // it: the invalidation raced its load and takes precedence.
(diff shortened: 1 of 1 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**CI** — Green: 22 passed, 0 failed. No real blocks (no flakes even triggered).

**Scope** — Tightly scoped to `aa-cache/src/l1.rs`: adds an `AtomicU64` epoch, bumps it in `invalidate()` before `entries.remove`, and guards the leader's post-load insert behind a shard `entry()` guard + epoch re-check. One new regression test. No API/signature changes.

**Side-effects** — None beyond intent. Single-flight/stampede path (`inflight` collapse, `notify_waiters`) is untouched; the `entry` guard is dropped before `inflight.remove`, no cross-map lock nesting. One extra atomic load per miss — negligible.

**Correctness** — The lost-invalidation ordering argument holds. For an eviction to be lost, `invalidate`'s `remove` would have to execute before the leader's insert while the leader's epoch check still passes. But `remove` and the guarded insert contend for the **same DashMap shard lock**, and the epoch bump precedes `remove`; so any interleaving where `remove` runs first establishes a lock release→acquire happens-before that forces the leader's `epoch.load(Acquire)` to observe the bump and skip the insert. Conversely if the leader inserts first, `remove` (running after under the same lock) drops it. Either way the revoked policy is not cached. Test `invalidate_during_load_is_not_lost` is meaningful — it fires `invalidate()` mid-load and asserts `len() == 0`, which fails against pre-fix code.

**Verdict** — ✅ Approve. Correct, minimal, well-tested. Flipping Draft → Ready.

PR #1365, opened 2026-07-02
[AAASM-3987] 🔒 (aa-runtime): Deadline on gateway policy RPC (fix hang HoL DoS)

Chisanan232:
## Description

`aa-runtime` forwarded policy checks to the gateway (`GatewayClient::check_action`) with **no deadline**. A gateway that accepts a connection but then stops responding (wedged, not down) would block `check_action` forever. Because the call is awaited inline in the single pipeline `run()` loop and holds the shared `GatewayClient` mutex, one hung RPC stalls **every** agent's policy checks, plus audit/redaction — a runtime-wide head-of-line DoS. The existing fail-closed logic (AAASM-3110) covers gateway *errors* but not *hangs*.

This adds a bounded per-RPC deadline: `try_gateway_forward` now wraps `check_action` in `tokio::time::timeout(gateway_timeout)`. On elapse the query is treated as a failure and routed into the **existing** fail-closed path (Deny under enforce / `gateway_fail_closed` default `true`), identical to a transport error. Reachable-gateway behavior is unchanged. Dropping the timed-out future also releases the client lock, so a subsequent stuck query cannot wedge longer than one deadline.

The deadline is operator-tunable via `AA_GATEWAY_TIMEOUT_MS` (default 5000ms); zero falls back to the default so the guard can never be disabled. A `aa_gateway_policy_query_timeouts_total` counter is emitted on each timeout.

**Follow-up (not in scope):** the policy query is still awaited inline in the single accept loop, so the deadline bounds — but does not eliminate — head-of-line serialization. Dispatching the query off the loop is a separate improvement; the deadline is the core fix.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-3987

## Testing

  - `config::tests::gateway_timeout_defaults_and_rejects_zero` — env parse + zero-rejects-to-default
  - `pipeline::tests::from_runtime_config_copies_all_fields` — timeout mapping
  - `pipeline::tests::gateway_hang_fail_closed_denies_within_deadline` — gRPC stub whose `check_action` parks forever; the check returns fail-closed Deny within the (shortened) deadline instead of blocking. Without the fix this test hangs.
- Validated locally: `cargo build -p aa-runtime`, `cargo nextest run -p aa-runtime` (363 pass + new tests; the only red was `enforce_scan_p99_within_budget`, a load-sensitive perf benchmark unrelated to this change — passes in isolation), `cargo fmt --all -- --check`, `cargo clippy -p aa-runtime --all-targets -- -D warnings` (clean).

## Checklist

Closes AAASM-3987

### DIFF
--- a/aa-runtime/src/config.rs
+++ b/aa-runtime/src/config.rs
@@ -4,6 +4,13 @@ use std::path::PathBuf;
 
 use crate::pipeline::enforcement::DEFAULT_MAX_FIELD_BYTES;
 
+/// Default per-RPC deadline for a gateway policy query, in milliseconds.
+///
+/// A few seconds is generous for a healthy in-cluster gateway hop yet short
+/// enough that a hung gateway cannot stall the runtime's policy checks for long
+/// (AAASM-3987).
+pub const DEFAULT_GATEWAY_TIMEOUT_MS: u64 = 5_000;
+
 /// Configuration for the `aa-runtime` sidecar process.
 ///
 /// All fields are populated by [`RuntimeConfig::from_env`].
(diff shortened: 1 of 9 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

AAASM-3987 aa-runtime gateway RPC deadline (AA_GATEWAY_TIMEOUT_MS 5s → fail-closed Deny on hang; 363 pass).

**CI:** green (Test/Build/Clippy/Format + smokes pass; cli_dashboard/policy_latency/enforce_scan_p99 known load-flakes not blocking). **Scope + side-effects:** verified against PR-head — behavior-preserving for the reachable/正常 path, fix confined as described, regression tests present. **Verdict: Approve.** Flipping ready.

PR #1368, opened 2026-07-02
[AAASM-3990] 🔒 (aa-sandbox): Cap table growth + per-store epoch deadline

Chisanan232:
## Description

Fixes two isolation gaps in the `aa-sandbox` WASM runtime (`aa-sandbox/src/runtime.rs`):

1. **Uncapped `table.grow` (OOM DoS).** `ResourceLimiter::table_growing` returned `Ok(true)` unconditionally while linear memory *was* capped, so a single guest `table.grow` (one non-fuel-metered instruction) could force a large eager host allocation. Added a per-store table-element cap (`MAX_TABLE_ELEMENTS`, mirroring the memory byte cap on the limiter struct, now `StoreLimits`) and reject growth past it **gracefully** via `Ok(false)` — which yields the standard WASM `-1` sentinel to the guest rather than trapping, while preventing the eager allocation.

2. **Engine-global wall-clock epoch (cross-invocation interference).** `set_epoch_deadline(1)` + `epoch_deadline_trap()` traps on any advance of the wasmtime `Engine`'s *process-global* epoch. On a `SandboxRuntime` shared across concurrent `run_tool` calls, one call's watchdog `increment_epoch()` would trap co-scheduled calls. Replaced the unconditional trap with an `epoch_deadline_callback` that interrupts **only** once *this* store's own wall clock has elapsed; a spurious tick from a sibling's watchdog re-arms the deadline (`UpdateDeadline::Continue`) instead of killing a call whose budget is not spent. `UpdateDeadline::Interrupt` maps to `Trap::Interrupt`, preserving the existing `WallClockTimeout` mapping.

**Epoch design choice:** wasmtime epoch interruption is inherently engine-global (`increment_epoch` has no per-store scope). A genuine per-store fix that also preserves the runtime's long-lived shared Engine/Linker amortization is the callback-gated deadline (interrupt only when the store's own `Instant` budget is spent). This is a real per-store fix, not merely a doc'd contract, and needs no per-call Engine rebuild.

Table cap is a module constant rather than a `SandboxLimits` field only because promoting it to the public config would break out-of-crate callers (`aa-cli`) that construct `SandboxLimits` with every field enumerated (out of scope here); a config field can follow once those sites adopt `..Default::default()`.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-3990

Closes AAASM-3990

## Testing

New tests in `aa-sandbox/src/runtime.rs`:
- `run_tool_rejects_table_grow_beyond_cap` — `table.grow` past the cap is rejected (guest observes `-1`).
- `run_tool_allows_table_grow_within_cap` — within-cap grow succeeds (no over-rejection).
- `wall_clock_gate_ignores_ticks_before_deadline` / `wall_clock_gate_interrupts_once_budget_elapsed` — per-store gate logic.
- `run_tool_kills_wall_clock_hog_with_wall_clock_timeout` — end-to-end wall-clock kill with huge fuel so only the epoch path can stop it.

Validation (scoped to `CARGO_TARGET_DIR`): `cargo build -p aa-sandbox`, `cargo nextest run -p aa-sandbox` (36/36 pass), `cargo fmt --all -- --check` clean, `cargo clippy -p aa-sandbox --all-targets -- -D warnings` clean.

## Checklist

- [ ] All CI checks passing

### DIFF
--- a/aa-sandbox/src/runtime.rs
+++ b/aa-sandbox/src/runtime.rs
@@ -685,4 +685,45 @@ mod tests {
             result,
         );
     }
+
+    #[test]
+    fn wall_clock_gate_ignores_ticks_before_deadline() {
+        // A far-future budget: an epoch tick right now (as a sibling store's
+        // watchdog would cause on a shared engine) must NOT interrupt this
+        // store — the per-store gate keeps it running. (AAASM-3990.)
+        assert!(!wall_clock_expired(Instant::now(), Duration::from_secs(3600)));
+    }
+
+    #[test]
+    fn wall_clock_gate_interrupts_once_budget_elapsed() {
+        // A zero budget is already past deadline for any elapsed time, so the
+        // gate interrupts — the store's own watchdog-tick path. (AAASM-3990.)
+        assert!(wall_clock_expired(Instant::now(), Duration::ZERO));
+    }
+
+    /// End-to-end wall-clock guard: an infinite loop with a huge fuel budget so
+    /// the fuel limiter never fires — only the wall-clock watchdog + per-store
+    /// epoch callback can stop it. Guards the epoch-deadline path introduced in
+    /// AAASM-3990 (the callback replaced the unconditional `epoch_deadline_trap`).
+    #[test]
+    fn run_tool_kills_wall_clock_hog_with_wall_clock_timeout() {
+        let config = SandboxConfig {
+            limits: crate::policy::SandboxLimits {
+                fuel: u64::MAX,
+                wall_clock_ms: 50,
+                ..Default::default()
+            },
+            ..Default::default()
+        };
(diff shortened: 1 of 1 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

AAASM-3990 aa-sandbox table.grow cap (10k, Ok(false)) + per-store epoch callback; 36 pass.

**CI:** green (Test/Build/Clippy/Format + smokes pass; cli_dashboard/policy_latency/enforce_scan_p99 known load-flakes not blocking). **Scope + side-effects:** verified against PR-head — behavior-preserving for the reachable/正常 path, fix confined as described, regression tests present. **Verdict: Approve.** Flipping ready.

PR #1369, opened 2026-07-02
[AAASM-3989] 🔒 (aa-sdk-client): Cap IPC frame + fail-closed on unknown decision

Chisanan232:
## Description

Hardens `aa-sdk-client` against untrusted IPC/gateway input and closes a
forward-compatibility fail-open in the enforcement fold.

- **AAASM-3989 (IPC frame allocation):** `read_length_delimited` allocated
  `vec![0u8; len]` from an attacker-controlled varint capped only at 64 bits, so
  a peer on the runtime socket (e.g. a `/tmp` socket-squatter when the runtime is
  down) could claim a near-`u64::MAX` payload and OOM-abort the host agent before
  a single body byte was read. Frames over `MAX_FRAME_LEN` (8 MiB, matching
  `aa-runtime`) are now rejected with `CodecError::FrameTooLarge` **before**
  allocating.
- **AAASM-3996 (Rust part):**
  - `decision.rs`: `resolve_decision` folded `UNSPECIFIED`/unknown decision codes
    to `Allow` even under `fail_closed`. Under enforce these now map to `Deny`,
    so a future restrictive `Decision` variant an older SDK cannot decode is not
    silently downgraded to `Allow`. Observe mode keeps the historical fail-open.
  - `gateway.rs`: pin an explicit 4 MiB `max_decoding_message_size` on the
    gateway gRPC client instead of relying on tonic's implicit default.
  - `ipc.rs`: documented (TODO) why query↔response correlation stays positional;
    id-based correlation needs a wire-protocol change touching `aa-runtime` +
    `aa-proto`, out of scope for this batch — deferred to its own ticket.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes (describe below)

Behavioral: under `fail_closed == true`, an `UNSPECIFIED`/unknown runtime
decision now resolves to `Deny` (previously `Allow`). This only affects the
enforce posture and is the intended fail-closed hardening; observe mode is
unchanged. The AAASM-3958 tests that locked `unknown → Allow` were updated
accordingly.

## Related Issues

- Related Jira ticket: Closes AAASM-3989 · Refs AAASM-3996 (Rust part; the
  node-sdk parity change is in a separate PR)

## Testing

- `cargo build -p aa-sdk-client` — clean
- `cargo nextest run -p aa-sdk-client` — 75 passed (new oversized-frame tests +
  updated fail-closed decision tests)
- `cargo clippy -p aa-sdk-client --all-targets -- -D warnings` — clean

## Checklist

- [ ] All CI checks passing

### DIFF
--- a/aa-sdk-client/src/codec.rs
+++ b/aa-sdk-client/src/codec.rs
@@ -40,12 +40,31 @@ pub const TAG_VIOLATION_ALERT: u8 = 4;
 /// runtime → SDK per-session nonce challenge (AAASM-3587).
 pub const TAG_HANDSHAKE_CHALLENGE: u8 = 5;
 
+/// Maximum accepted length-delimited payload size, in bytes (8 MiB).
+///
+/// The wire length prefix is attacker-controlled: anything that can connect to
+/// the runtime socket (including a `/tmp` socket-squatter when the real runtime
+/// is down) can send a varint claiming a multi-gigabyte payload, and
+/// `vec![0u8; len]` would attempt to allocate it *before* a single payload byte
+/// is read — a trivial local DoS that OOM-aborts the agent (AAASM-3989). We
+/// reject any frame larger than this bound before allocating. 8 MiB matches
+/// `aa-runtime`'s `MAX_FRAME_LEN`, comfortably exceeding any legitimate
+/// `CheckActionResponse` / `PolicyViolation` while keeping a single hostile
+/// frame from exhausting memory.
+pub const MAX_FRAME_LEN: usize = 8 * 1024 * 1024;
+
 /// Errors that can occur during codec operations.
 #[derive(Debug)]
 pub enum CodecError {
     Io(std::io::Error),
     UnknownTag(u8),
     DecodeError(prost::DecodeError),
+    /// The wire length prefix exceeded [`MAX_FRAME_LEN`]. Rejected before any
+    /// allocation so a hostile peer cannot force a large buffer alloc.
+    FrameTooLarge {
+        len: usize,
+        max: usize,
+    },
 }
 
 impl std::fmt::Display for CodecError {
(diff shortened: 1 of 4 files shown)


Chisanan232:
## 🤖 Claude Code — PR Review

**Verdict: APPROVE (scope-clean, additive).** [AAASM-3989 + AAASM-3996]

**codec.rs — frame bound (3989):** `MAX_FRAME_LEN = 8 MiB` + `CodecError::FrameTooLarge { len, max }`. The guard `if len > MAX_FRAME_LEN { return Err(..) }` sits **before** `vec![0u8; len]`, so a hostile varint prefix is rejected pre-allocation — the OOM/DoS vector is closed. Two tests cover it: `oversized_frame_length_rejected_before_alloc` (u64::MAX prefix, no body → `FrameTooLarge`) and `frame_at_max_len_passes_the_bound` (exact `MAX_FRAME_LEN` passes the `>` bound and fails only on the absent body). Boundary correct.

**decision.rs — uninterpretable → Deny under enforce (3996):** `Ok(Unspecified) | Err(_)` now folds to `Deny` when `fail_closed`, `Allow` otherwise. Closes the fwd-compat downgrade (an old SDK seeing a future restrictive variant as an unknown code no longer silently allows). Verified preserved: runtime `DENY`/`REDACT` honoured verbatim in both modes; transport-error and `PENDING` still `Deny` under enforce (3958/3920 untouched); Allow-on-unknown only when fail-closed disabled. Doc truth-table updated to match. The four rewritten tests assert exactly this matrix.

**gateway.rs:** explicit `max_decoding_message_size(4 MiB)` pins tonic's implicit default against a codegen/tonic bump — defensive, additive.

**ipc.rs:** request-id correlation deferred with a documented `TODO(AAASM-3996)`; single-connection FIFO invariant holds today. Appropriate scope boundary.

No side effects beyond `aa-sdk-client`. CI green on the real checks (Rust analyze, Rustdoc, CodeQL, standalone build); remaining are conformance/placeholder jobs.

PR #1372, opened 2026-07-02
[AAASM-3986] 🐛 (aa-gateway): Atomic budget check_and_decrement for LLM spend

Chisanan232:
## What
Closes the budget check→record TOCTOU in the live LLM-call path. `BudgetTracker::check_and_decrement` (per-ancestor mutex) previously had no non-test caller: the gateway read accumulated spend at Stage 7, returned the decision, and only recorded the LLM cost *after* the response was built. Those two steps were not atomic.

Adds `BudgetTracker::reserve_spend` (atomic check + commit across agent + ancestors + team/org/global under the per-ancestor lock set) and `PolicyEngine::check_and_accrue_llm_spend`, and routes `maybe_accrue_llm_spend` through it in both `check_action` and `batch_check`. On a rejected reservation the response is rewritten to a hard `Deny` (like the Stage-7 budget deny) and nothing is charged.

## Why
N concurrent `check_action` calls for one tenant could all observe under-limit before any recorded, causing bounded overspend proportional to in-flight concurrency (AAASM-3986). The reserve uses the same `spent >= limit` semantics as the Stage-7 read-check, so the **decision output is preserved** while the concurrency-proportional overspend is removed. Existing tiers/ancestors semantics are preserved.

## How verified
- New unit test `check_and_accrue_llm_spend_commits_within_budget`.
- New concurrency test `check_and_accrue_llm_spend_no_overspend_under_parallel_checks` (64 parallel $1 reservations vs a $10 limit → exactly 10 commit, recorded spend == $10).
- Pre-existing `budget_accrual_test` still green; full `aa-gateway` suite green; `cargo fmt --check` + `cargo clippy -D warnings` clean.

Closes AAASM-3986.

### DIFF
--- a/aa-gateway/src/budget/tracker.rs
+++ b/aa-gateway/src/budget/tracker.rs
@@ -765,7 +765,9 @@ impl BudgetTracker {
         }
         if let Some(limit) = self.resolve_limit(&agent_id, BudgetKind::Daily) {
             if agent_daily >= limit {
-                return Err(BudgetError::SelfBudgetExhausted { kind: BudgetKind::Daily });
+                return Err(BudgetError::SelfBudgetExhausted {
+                    kind: BudgetKind::Daily,
+                });
             }
         }
         self.preflight_ancestors(ancestors, amount)?;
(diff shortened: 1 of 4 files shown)


Chisanan232:

AAASM-3986 budget TOCTOU. `BudgetTracker::reserve_spend` does atomic check+commit across agent/ancestors/team/org/global under the per-ancestor lock set; `check_action` + `batch_check` route through it, rewriting to Deny on rejection. Decision output preserved (`spent >= limit`, same as Stage-7). +2 tests incl. 64-thread no-overspend proof. CI green, scope matches ticket, no side-effects outside budget path.

PR #1373, opened 2026-07-02
[AAASM-3992] 🐛 (aa-gateway): Deny tokenless check_action with unauthenticated tenancy

Chisanan232:
## What
Completes AAASM-3416. When a `check_action` (or `batch_check`) arrives with an **empty credential token**, an **unregistered** `{org,team,agent}` triple, **and** client-supplied tenancy (`org_id`/`team_id`), no authoritative owner can be resolved — yet `authoritative_tenancy` / `authoritative_lineage` would fall back to trusting the client-supplied tenancy. `validate_credential_token` now fails closed with a `Deny` ("unauthenticated tenancy claim") before any budget mutation or audit write.

## Why
PolicyService runs under the non-rejecting `enrich` interceptor, so at `:50051` an unauthenticated peer could send such a request to (a) accrue LLM spend against a victim org's budget (org-tier DoS), and (b) forge attacker-controlled audit entries into the WORM hash chain under a victim tenant. Registered-agent impersonation was already blocked; this closes the tokenless-unregistered residual. Fail-closed.

The guard is scoped to requests that actually claim tenancy, so:
- The legitimate **registered-agent + valid token** path is unchanged.
- **Untenanted** tokenless requests (no org/team) — the lightweight untenanted/unregistered-deployment path — still evaluate normally.
- The bootstrap/enroll path (`register_agent`, a separate RPC) is unaffected.

## How verified
New `tokenless_tenancy_deny_test`: empty-token + unregistered + tenancy-claim → `Deny` with **no audit write**; registered+valid token → `Allow`; untenanted tokenless → passes through. Existing `batch_check_credential_validation_test` still green; full `aa-gateway` suite green; fmt + clippy clean.

Closes AAASM-3992.


Chisanan232:
🤖 Claude Code review — **approve** (reworked)

AAASM-3992 tokenless-tenancy → cross-tenant budget-DoS + audit forgery.

The first cut hard-denied `empty-token + tenancy-claim + unregistered`, which broke the normal runtime→gateway flow: the default gateway always attaches a registry, and OSS/self-host agents are typically unregistered and forward tenancy without a per-agent token — so every legit unregistered agent got denied (deterministic e2e failure in `runtime_forwards_per_tool_deny_to_gateway`).

**Reworked to neutralize, not deny.** New `apply_authoritative_tenancy(req, &mut ctx)` at the service layer (where the token + composite registry key are available): token resolves to a registered owner → deposit that owner's governance_level/org/team (the pre-existing AAASM-3751 deposit, factored out); otherwise → **strip** client-supplied org_id/team_id so the engine resolves an anonymous/empty tenant. Called in both `evaluate_one` (cascade) and `maybe_accrue_llm_spend` (budget), so both inherit neutralized tenancy. Audit path was already neutral for unregistered agents (lineage miss → empty). Hard-Deny branch removed.

**Vuln closed** — new tests: unregistered tokenless peer claiming `victim-org`/`victim-team` on a $1.00 call is Allowed but leaves `team_state("victim-team")`/`org_state("victim-org")` = None and the audit entry's org/team are not the victim's. `engine/mod.rs` deliberately untouched (its registry lookup is agent-id-only; neutralizing there would strip every agent + break 3591/3621).

aa-gateway 1344 pass (5 neutralization tests), conformance 46, e2e PASS, clippy clean. CI green.

PR #1377, opened 2026-07-02
[AAASM-4012] 🐛 (aa-ebpf): Attach full file-I/O kprobe return-probe set

Chisanan232:
## What
`FileIoLoader::attach_kprobes` attached the `openat` entry+return probes but only the **entry** kprobes for `read`/`write`/`unlink`/`rename`. Those syscalls emit their `FileIoEventRaw` from the **kretprobe** (`aa_sys_*_ret`), which was never attached — so in practice only `openat` produced events. The legacy glibc `unlink(2)`/`rename(2)` entry+return pairs (AAASM-1574) were also missing from the loader.

This PR:
- Promotes `KprobeManager::KPROBE_TARGETS` to the **single authoritative** list of all 14 (entry + return) programs, including the four missing `*_ret` retprobes and the two legacy `*_legacy_ret` retprobes.
- Makes **both** `KprobeManager::attach` and `FileIoLoader::attach_kprobes` iterate that one list, so the two entry points can no longer drift.
- `spawn_ebpf_file_io` now seeds `PATH_BLOCKLIST` (exact-file sensitive rules) via `update_path_filter` and reads events through `start_event_reader_with_alerts(SensitivePathDetector)`, so events actually carry an `is_sensitive` signal (in-kernel exact-match + userspace prefix/canonical).

## Why
Without the return probes attached, the file-I/O eBPF sub-layer silently observed only `open` — read/write/delete/rename went unmonitored, and no event was ever flagged sensitive.

## How verified
- `cargo fmt --all`, `cargo check -p aa-ebpf`, `cargo clippy -p aa-ebpf --all-targets -- -D warnings` — clean.
- `cargo test -p aa-ebpf --lib` — 104 pass, including a new regression test asserting every entry kprobe has a paired return probe and the list has all 14 targets.
- `cargo check -p aa-runtime` (macOS) — clean.

## ⚠️ Linux CI is the authoritative gate
eBPF is Linux-only. The `attach_kprobes` / `spawn_ebpf_file_io` bodies are `#[cfg(target_os = "linux")]` and cannot be compiled on the macOS dev host (the BPF bytecode objects are not built there). The changed logic mirrors the already-working `KprobeManager::attach` loop and established loader APIs; the **`e2e — Layer 3 eBPF (Linux)`** and **`eBPF probes build (Linux nightly)`** jobs are the real validation.

Refs AAASM-4012

### DIFF
--- a/aa-ebpf/src/kprobe.rs
+++ b/aa-ebpf/src/kprobe.rs
@@ -59,21 +59,15 @@ impl KprobeManager {
                 .map_err(|e| EbpfError::ProbeAttach(e.to_string()))?;
         }
 
-        // Attach all file I/O kprobe programs to their kernel functions.
-        // `aa_sys_unlink_legacy` / `aa_sys_rename_legacy` cover the legacy
-        // syscall entry points that glibc on x86_64 invokes for libc
-        // `unlink()` / `rename()` — bypassing the at-variant probes
-        // above. See AAASM-1574.
-        let probes: &[(&str, &str)] = &[
-            ("aa_sys_openat", "__x64_sys_openat"),
-            ("aa_sys_openat_ret", "__x64_sys_openat"),
-            ("aa_sys_read", "__x64_sys_read"),
-            ("aa_sys_write", "__x64_sys_write"),
-            ("aa_sys_unlink", "__x64_sys_unlinkat"),
-            ("aa_sys_unlink_legacy", "__x64_sys_unlink"),
-            ("aa_sys_rename", "__x64_sys_renameat2"),
-            ("aa_sys_rename_legacy", "__x64_sys_rename"),
-        ];
+        // Attach all file I/O kprobe programs to their kernel functions from the
+        // single authoritative target list ([`KprobeManager::KPROBE_TARGETS`]).
+        // Every observed syscall has BOTH an entry kprobe and a return
+        // kretprobe — emission happens in the kretprobe, so attaching entry
+        // probes alone yields no events (AAASM-4012). The `*_legacy` pairs cover
+        // the legacy syscall entry points that glibc on x86_64 invokes for libc
+        // `unlink()` / `rename()`, bypassing the at-variant probes. See
+        // AAASM-1574.
+        let probes = Self::KPROBE_TARGETS;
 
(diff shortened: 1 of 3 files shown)


Chisanan232:

AAASM-4012 eBPF file-io attach list. Promoted `KprobeManager::KPROBE_TARGETS` to the single authoritative 14-program list (entry+return incl. legacy pairs); both `KprobeManager::attach` and `FileIoLoader::attach_kprobes` iterate it → read/write/unlink/rename now emit. `spawn_ebpf_file_io` seeds `PATH_BLOCKLIST` + uses `start_event_reader_with_alerts` → `is_sensitive` populated. Scope matches ticket; disjoint from other lanes.

**Linux validation:** PR CI path-gate skipped the eBPF jobs, so I dispatched `ci.yml` on this branch — `eBPF probes build (Linux nightly)` + `e2e — Layer 3 eBPF (Linux)` both **PASS**. aa-ebpf lib 104 tests pass.

Chisanan232:
🤖 Claude Code — **deep review: APPROVE (merge-ready)**

**CI:** green — dispatched `ci.yml` on this branch; `eBPF probes build` + `e2e — Layer 3 eBPF (Linux)` both **PASS**. **Scope:** complete — the 14 shared kprobe targets (entry+ret for openat/read/write/unlink/rename + legacy unlink/rename pairs) all verified to exist as real BPF programs; both `KprobeManager::attach` and `FileIoLoader::attach_kprobes` iterate the one list (a pairing-invariant test guards drift).

**Side-effects (regression check):** retprobes attach to the same symbols as entry probes (proven pattern, cf. existing openat pair) — no double-attach issue; the shared-list refactor's only non-production caller (a gated live-kernel test) asserts attach/detach not counts → unaffected; `_with_alerts` adds one task + a bounded mpsc hop (negligible); `is_sensitive` is consumed only by a `tracing::warn!` — **no downstream consumer reads it**, zero mishandling risk. No regression.\n\n**Non-blocking notes:** `is_sensitive` is log-only (not propagated into the audit event) and the blocklist is seeded from `SensitivePathDetector` defaults, not policy (true policy wiring is subsumed by AAASM-4011) — both are in-scope choices; follow-up to propagate the sensitive signal into the pipeline.

PR #1378, opened 2026-07-02
[AAASM-4015] 🔒 (aa-sandbox): cap memory/table/instance counts + preopen/telemetry hardening

Chisanan232:
## What

Three defense-in-depth hardening fixes to the WASM/WASI sandbox runtime (`aa-sandbox/src/runtime.rs`):

1. **Cap per-store memory/table/instance counts.** `impl ResourceLimiter for StoreLimits` previously overrode only `memory_growing`/`table_growing` (per-item size), leaving wasmtime's ~10 000 count defaults for `memories()`/`tables()`/`instances()`. A hostile multi-memory/multi-table module could declare many of each — every memory carrying its own `max_bytes` allowance and every table its own `MAX_TABLE_ELEMENTS` allowance — multiplying past the intended per-store ceiling (OOM-DoS). Now `memories()`/`tables()`/`instances()` are pinned to `1`, and the unused `multi_memory` proposal is disabled in the engine `Config` (validation-time rejection). `reference_types` is intentionally kept enabled: funcref tables and `table.grow` depend on it.

2. **Reject non-regular-file preopen targets.** The wall-clock epoch watchdog cannot preempt a guest blocked *inside* a WASI host call (e.g. a `fd_read` on a FIFO/device in a preopen). Each preopen `host_path` is now validated (following symlinks) to resolve to a directory or regular file; FIFOs, devices, and sockets are rejected up front. This is the lower-risk option versus running the guest call on an abandonable worker thread.

3. **Map instantiate-time OOM to `MemoryExhausted`** *(scoped from AAASM-4020 — same file)*. An oversized *initial* linear memory trips `memory_growing` during `linker.instantiate` (initial allocation is a grow from 0). The `MemoryExhaustedMarker` downcast previously existed only on the post-`call` branch, so instantiate-time OOM was misclassified as `SandboxError::Wasmtime` (→ `SandboxTerminated`) instead of `SandboxError::MemoryExhausted` (→ `SandboxOomKilled`). The marker is now downcast on the instantiate branch too.

## Why

AAASM-4015 [MED]: the per-item resource limiter did not bound resource *counts*, and the wall-clock deadline could not preempt a guest stuck in a blocking WASI host call — both OOM/hang-DoS vectors. Item 3 corrects OOM telemetry misclassification (folded from AAASM-4020 because it is the same file/mechanism).

## How verified

- `cargo nextest run -p aa-sandbox` — **41 passed, 0 skipped** (36 pre-existing + 5 new: multi-memory rejected, multi-table rejected, single memory+table still runs, FIFO preopen rejected, instantiate-time OOM → `MemoryExhausted`)
- `cargo clippy -p aa-sandbox --all-targets -- -D warnings` — exit 0

Only `aa-sandbox/src/runtime.rs` is touched.

Closes AAASM-4015. Folds AAASM-4020 (same-file telemetry fix).

### DIFF
--- a/aa-sandbox/src/runtime.rs
+++ b/aa-sandbox/src/runtime.rs
@@ -285,6 +285,24 @@ impl SandboxRuntime {
 
         let mut builder = WasiCtx::builder();
         for preopen in &self.config.preopened_dirs {
+            // Reject non-regular-file preopen targets (AAASM-4015). The
+            // wall-clock watchdog interrupts the guest by advancing the engine
+            // epoch, but it cannot preempt a guest blocked *inside* a WASI host
+            // call — e.g. a `fd_read` on a FIFO/device/socket that never yields.
+            // Validating that every preopen target is a directory or regular
+            // file (following symlinks) forbids those blocking targets up front,
+            // which is a lower-risk fix than abandoning an in-flight host call on
+            // a worker thread. `metadata` follows symlinks, so a symlink to a
+            // device resolves to the device and is rejected.
+            let metadata = std::fs::metadata(&preopen.host_path)
+                .map_err(|e| SandboxError::Wasmtime(format!("preopen {:?}: {e}", preopen.host_path)))?;
+            let file_type = metadata.file_type();
+            if !(file_type.is_dir() || file_type.is_file()) {
+                return Err(SandboxError::Wasmtime(format!(
+                    "preopen target {:?} is not a regular file or directory (FIFOs, devices, and sockets are forbidden)",
+                    preopen.host_path
+                )));
+            }
             // Least-privilege grant: read-only unless the policy explicitly
             // opted this mount into read-write. Avoids the DirPerms::all() /
             // FilePerms::all() over-grant. (AAASM-3618.)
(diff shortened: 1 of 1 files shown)


Chisanan232:

AAASM-4015 aa-sandbox. `memories()`/`tables()`/`instances()` overrides pinned to 1 + `multi_memory` proposal disabled → closes the count-multiplication OOM (the 3990 per-item caps didn't cover counts). Preopen targets validated to regular-file/dir (FIFO/device rejected) → closes the WASI-blocking wall-clock bypass. Instantiate-time OOM now maps to `SandboxOomKilled` (folded 4020 telemetry item). 41 pass, clippy clean. Confined to aa-sandbox.

Chisanan232:
🤖 Claude Code — **deep review: APPROVE (merge-ready)**

**CI:** green (0 fail; skips are path-filtered). **Scope:** all 3 AAASM-4015 items covered (count caps `memories/tables/instances=1` + `multi_memory` off; preopen non-regular-file rejection; folded 4020 instantiate-OOM telemetry). Correctly *kept* `reference_types` on (funcref/table.grow depend on it).

**Side-effects (regression check):** the count-pin-to-1 risk is empirically disproven — every WASM fixture the sandbox runs (6 unit + fs_probe/mem_bomb/runaway) declares a single memory/table or none; no product tool-module uses multi-memory. A positive over-rejection guard test confirms single-mem+table still instantiates+runs. No happy-path regression.

**Non-blocking follow-ups:** preopen validation is shallow (a special file *inside* a preopened dir isn't caught — misconfig-only, guests can't mkfifo); `tables()=1` is hard-coded (would reject a future module using a 2nd externref table — none exists today).

PR #1380, opened 2026-07-02
[AAASM-4011] ✨ (aa-runtime): Wire runtime to privileged eBPF loaderd

Chisanan232:
## What
Layer 3 (eBPF) was **dormant** in the production posture. AAASM-3603/3605 moved BPF privilege into the standalone `aa-ebpf-loaderd` daemon and made `aa-runtime` drop `CAP_BPF`/`CAP_PERFMON` (`privilege.rs`), yet `runtime::run` still loaded probes **in-process** via `aya::Ebpf::load` — which needs the very capability just dropped. So on the unprivileged runtime every eBPF sub-layer could only `EPERM`-degrade (masking "Layer 3 never came up" as a soft degradation), the loaderd daemon sat idle (nothing drove its control channel), and the only enforcing/SIGKILL probe — the syscall guard — was never loaded at all.

This PR:
- Adds `aa-runtime/src/ebpf_control.rs`: connects to the loaderd control socket as an unprivileged client (`LoaderControlClient`) and drives it — `LoadProbeSet` for TLS/file-I/O/exec, `UpdatePathMap` from the lowered policy, and (opt-in) `LoadProbeSet` SyscallGuard + `UpdateSyscallAllowlist`.
- Rewires the `run()` eBPF block to **drive loaderd by default**. The in-process spawns are retained only behind the off-by-default `AA_EBPF_INPROCESS_LOAD` escape hatch (privileged dev/integration scenario), so they can never silently mask a dormant Layer 3 in production.
- Leaves `enforce_least_privilege` intact — the runtime stays unprivileged; the daemon holds the caps.
- Reads config from env (`AA_EBPF_LOADERD_SOCK`, `AA_EBPF_POLICY_PATH`, `AA_EBPF_CONFINE_PID`, `AA_EBPF_INPROCESS_LOAD`), matching `layer.rs`'s existing pattern — `RuntimeConfig` is intentionally not touched (its exhaustive struct literals live partly in an off-limits file).

## Safety — syscall guard is default-deny (SIGKILL)
For any PID in its `PID_FILTER`, the guard SIGKILLs any syscall absent from `SYSCALL_ALLOWLIST`. To make self-inflicted death impossible by construction, the planner loads the guard **only** when a confine PID is set (`AA_EBPF_CONFINE_PID`, must be the sandbox PID — never the runtime's own) **and** the policy lowers to a **non-empty** allowlist. Unit tests lock this in (`syscall_guard_never_planned_without_a_confine_pid`, `..._with_empty_allowlist`).

## Preserved observe-only vs enforce split
TLS/file-I/O/exec stay observe-only (scoped to the runtime PID as before); only the syscall guard enforces, scoped to the confine PID.

## How verified
- `cargo fmt --all`, `cargo check -p aa-runtime`, `cargo clippy -p aa-runtime --all-targets -- -D warnings` — clean.
- `cargo test -p aa-runtime --lib` — 358 pass (incl. 7 new `ebpf_control` tests covering the safety planner + socket resolution).

## ⚠️ Documented as remaining / Linux CI is the gate
eBPF is Linux-only; `drive_ebpf_layer`'s real body is `#[cfg(target_os = "linux")]` and cannot be compiled on the macOS dev host (BPF bytecode isn't built there). The pure planner + config helpers are cross-platform and unit-tested. The extended module NOTE in `aa-integration-tests/tests/e2e_ebpf.rs` specifies exactly what a Linux e2e must verify, and these known gaps that this PR **deliberately does not** close (they need protocol/streaming work, out of scope):
1. **load→allowlist ordering hazard** — the protocol couples load+attach+PID-insertion in one `LoadProbeSet`, leaving a window where the confined PID runs with an empty allowlist. A race-free fix needs a `load-without-filter → set allowlist → add PID` protocol change.
2. **observe-only telemetry** — `LoadProbeSet`'s doc promises event streaming back, but `control::server::dispatch` does not stream; TLS/file-I/O/exec events captured by the daemon do not yet reach the runtime pipeline. Enforcement (SIGKILL + in-kernel path flagging) is autonomous and unaffected.

The `e2e — Layer 3 eBPF (Linux)` and `eBPF probes build (Linux nightly)` jobs are the authoritative gate.

## Note on AAASM-4012 (PR #1377)
That PR modifies `spawn_ebpf_file_io`; this PR keeps that function (now behind the `AA_EBPF_INPROCESS_LOAD` gate) and does not touch its body, so there is no textual conflict. Merge order does not matter.

Refs AAASM-4011

### DIFF
--- a/aa-integration-tests/tests/e2e_ebpf.rs
+++ b/aa-integration-tests/tests/e2e_ebpf.rs
@@ -48,6 +48,46 @@
 //! is what the gateway path will read once wired. When the HTTP path is
 //! complete, the gateway-side assertion can be added in a follow-up
 //! without changing the probe-fired half of the check.
+//!
+//! ## GAP — runtime → loaderd orchestration is NOT exercised here (AAASM-4011)
+//!
+//! Every test below loads and attaches probes **directly in-process** via
+//! `EbpfLoader` / `UprobeManager` / `TracepointManager` under `sudo`. That
+//! validates the probes themselves, but it deliberately bypasses the
+//! production control path AAASM-4011 wired up: the **unprivileged** runtime
+//! driving the **privileged** `aa-ebpf-loaderd` daemon over its control socket
+//! (`aa_runtime::ebpf_control::drive_ebpf_layer` →
+//! `aa_ebpf::control::client::LoaderControlClient` →
+//! `aa_ebpf::control::server`). None of that orchestration is covered by any
+//! test yet, so a Linux e2e must be added that verifies, end to end:
+//!
+//! 1. **Unprivileged runtime, privileged daemon.** Start `aa-ebpf-loaderd`
+//!    with `CAP_BPF`/`CAP_PERFMON` and the runtime with **none** (the
+//!    `enforce_least_privilege` posture). With `AA_EBPF_INPROCESS_LOAD` unset,
+//!    `drive_ebpf_layer` must connect to the socket and `LoadProbeSet` the TLS,
+//!    file-I/O, and exec sets — no `ebpf/*` sub-layer degrades.
+//! 2. **Path map push.** With `AA_EBPF_POLICY_PATH` pointing at a policy that
+//!    denies a sensitive path, `UpdatePathMap` reaches the daemon and the
+//!    file-I/O probe flags an access to that path (`flags` bit 0 set).
+//! 3. **Syscall guard (enforce).** With `AA_EBPF_CONFINE_PID=<sandbox pid>` and
+//!    a policy that lowers to a non-empty syscall allowlist, the guard loads and
+//!    `UpdateSyscallAllowlist` lands; a confined process issuing a
+//!    non-allowlisted syscall is SIGKILLed, while the runtime's own syscalls are
+//!    never confined (guard scoped to the sandbox PID, not the runtime).
+//! 4. **Load→allowlist ordering hazard.** The current protocol couples
+//!    load+attach+PID-filter insertion in one `LoadProbeSet`, so there is a
+//!    window where the confined PID runs with an empty allowlist (default-deny).
+//!    The e2e must confirm the confined process is not spuriously killed before
+//!    `UpdateSyscallAllowlist` is applied — and, if it can be, that confirms the
+//!    protocol needs a `load-without-filter → set allowlist → add PID` sequence
+//!    (a follow-up on `aa_ebpf::control`).
+//! 5. **Observe-only telemetry gap.** `LoadProbeSet`'s doc promises the daemon
+//!    "begins streaming events back", but `aa_ebpf::control::server::dispatch`
+//!    does not yet stream — it only loads/attaches. So TLS/file-I/O/exec events
+//!    captured by the daemon-owned readers do **not** reach the runtime pipeline
+//!    over the control channel. Enforcement (syscall-guard SIGKILL, in-kernel
+//!    path-blocklist flagging) is autonomous and unaffected, but observe-only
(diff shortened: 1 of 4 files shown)


Chisanan232:
🤖 Claude Code review — **approve** (validated on real kernel)

AAASM-4011 Layer-3 wiring. New `aa-runtime/src/ebpf_control.rs` drives the privileged loaderd via `LoaderControlClient` (LoadProbeSet TLS/file-io/exec + SyscallGuard, UpdatePathMap, UpdateSyscallAllowlist) by default; the dead in-process `aya::Ebpf::load` path is gated behind off-by-default `AA_EBPF_INPROCESS_LOAD`; `enforce_least_privilege` untouched. **Safety-critical:** syscall-guard (default-deny SIGKILL) loads ONLY with a confine-PID + non-empty allowlist — never suicides the runtime (locked by unit tests). 358 aa-runtime tests pass.

**Linux validation:** dispatched `ci.yml` on this branch — `eBPF probes build` + `e2e — Layer 3 eBPF (Linux)` both **PASS** on a real kernel. Two documented residual gaps (load→allowlist ordering; daemon→pipeline observe-only streaming) are non-enforcement and tracked in the extended e2e NOTE — enforcement (SIGKILL + in-kernel path flagging) is autonomous and confirmed. This completes the half-done 3603/3605 migration.

Chisanan232:
🤖 Claude Code — **deep review: APPROVE (merge-ready; one regression found + fixed in-PR)**

**CI:** green — a dispatched `ci.yml` run on this branch passes `eBPF probes build (Linux nightly)` + `e2e — Layer 3 eBPF (Linux)` on a real kernel (the PR-CI path-gate skips those jobs). **Scope:** complete — `ebpf_control.rs` drives loaderd via `LoaderControlClient` (LoadProbeSet Tls/FileIo/Exec + SyscallGuard, UpdatePathMap, UpdateSyscallAllowlist); dead in-process `aya::Ebpf::load` gated behind off-by-default `AA_EBPF_INPROCESS_LOAD`; `privilege.rs`/`main.rs` untouched (least-privilege intact). Completes the 3603/3605 migration.

**Side-effects (regression check):**
- ✅ non-eBPF + non-Linux paths unaffected (all inside `if LayerSet::EBPF`; `cfg(not(linux))` degrade stub).
- ✅ no-socket → `connect` Err → graceful degrade, no crash.
- ⚠️ **Found: a NEW availability regression** — `drive_ebpf_layer().await` ran inline in runtime boot and the loaderd `read_frame` has no timeout, so a *connected-but-hung* daemon would wedge boot (SDK IPC never comes up) — the old in-process path failed fast on local EPERM. **Fixed in this PR** (commit `0f8aea33`): every loaderd round-trip (connect + load/update ops) is now wrapped in a per-request deadline (`AA_EBPF_LOADERD_TIMEOUT_MS`, default 5s) → a hung daemon degrades the sub-layers instead of blocking boot, restoring the fail-fast liveness guarantee. Verified: Build + Clippy + eBPF e2e green on the re-dispatched run.
- ✅ Also hardened the syscall-guard suicide-fallback footgun: refuse to scope the default-deny SIGKILL guard to the runtime's own PID if the planner invariant is ever broken (skip loudly instead of `unwrap_or(observe_pid)`).\n\n**Non-blocking follow-ups (documented, non-enforcement):** (1) implement the Linux e2e for the runtime→loaderd orchestration (currently a code NOTE, no ticket) — file a follow-up; (2) load→allowlist ordering window + daemon→pipeline observe-only telemetry streaming — both non-enforcement (SIGKILL + in-kernel path flagging are autonomous); no production telemetry regression since the old path always EPERM-degraded.

PR #1392, opened 2026-07-03
[AAASM-4033] ✅ (aa-integration-tests): Add runtime→loaderd orchestration e2e (Linux)

Chisanan232:
## What

Adds two Linux-only e2e tests to `aa-integration-tests/tests/e2e_ebpf.rs`, closing the code NOTE that AAASM-4011 left behind ("runtime → loaderd orchestration is NOT exercised here"):

- **Test 7 — `ebpf_runtime_orchestration_drives_loaderd`**: spawns the real `aa-ebpf-loaderd` privileged daemon on a private control socket and drives the exact plan `aa_runtime::ebpf_control::drive_ebpf_layer` executes (`plan_control_ops`) via the public `LoaderControlClient`: `LoadProbeSet` for the TLS / file-I/O / exec observe-only sets, `UpdatePathMap` (a sensitive-path deny lowered through the canonical `aa_security::policy::lower_to_ebpf`), then `LoadProbeSet` for the syscall guard — scoped to a throwaway sandbox PID, never the test process — followed by `UpdateSyscallAllowlist`. Each op returning `Ok` is ground truth that the probe actually loaded/attached and the map updated in the kernel *through the daemon over the socket*.
- **Test 8 — `ebpf_loaderd_boot_liveness_degrades_without_hanging`**: asserts the boot-liveness property `drive_ebpf_layer` relies on (`await_loaderd`/`loaderd_deadline`, `AA_EBPF_LOADERD_TIMEOUT_MS`). An **absent** daemon makes `connect` fail fast; a **hung** daemon (accepts, never replies) makes a control op block indefinitely on its own and complete only because a runtime-style deadline wrapper elapses it — i.e. the layer degrades rather than hangs.

The 5-point NOTE is rewritten to describe what is now covered vs. what stays deferred.

## Why

AAASM-4011 moved BPF privilege into the standalone `aa-ebpf-loaderd` daemon and had the unprivileged runtime drive it over a control socket, but **no test exercised that orchestration** — only in-process probe loads were covered. This PR adds that end-to-end coverage.

## How to verify

- **Linux CI is the gate.** eBPF is Linux-only and requires root (CAP_BPF + CAP_PERFMON), so these tests only run in the `e2e — Layer 3 eBPF (Linux)` job. The whole file is `#[cfg(all(target_os = "linux", feature = "integration-test"))]`; macOS builds see an empty file (no failures, no `#[ignore]`).
- The PR path-filter (`changes.ebpf`) does **not** trigger the eBPF job for a test-only change under `aa-integration-tests/`, so a **manual `workflow_dispatch`** is needed to run Test 7/8 on Linux: `gh workflow run ci.yml --ref v0.0.1/AAASM-4033/test/loaderd_orchestration_e2e`.
- Locally (macOS): `cargo fmt --all` and `cargo check -p aa-integration-tests --features integration-test` pass; the Linux test body cannot be compiled or run on macOS (aya is Linux-only).

## Deferred (documented in-file)

- Observe-only telemetry over the control channel (the daemon does not yet stream captured events back, so the client can't observe the file-I/O path-deny flag bit or TLS/exec captures via the daemon path).
- The syscall-guard SIGKILL enforcement assertion (needs a purpose-built confined helper issuing a controlled forbidden syscall).
- The load→allowlist race-free protocol fix (`load-without-filter → set allowlist → add PID`).

## Scope

Touches only `aa-integration-tests/tests/e2e_ebpf.rs`. No `aa-runtime` / `aa-ebpf` / `aa-gateway` source changed.

Closes AAASM-4033

### DIFF
--- a/aa-integration-tests/tests/e2e_ebpf.rs
+++ b/aa-integration-tests/tests/e2e_ebpf.rs
@@ -33,6 +33,8 @@
 //! | 4 | `ebpf_catches_traffic_without_sdk_init` | enabled (root + libssl) |
 //! | 5 | `ebpf_event_includes_pid_and_cgroup` | enabled (root) |
 //! | 6 | `ebpf_load_and_unload_clean` | enabled (root) |
+//! | 7 | `ebpf_runtime_orchestration_drives_loaderd` | enabled (root + loaderd) |
+//! | 8 | `ebpf_loaderd_boot_liveness_degrades_without_hanging` | enabled (no BPF) |
 //!
 //! ## Why direct ring-buffer reads instead of HTTP gateway lookup
 //!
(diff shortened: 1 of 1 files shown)


Chisanan232:
🤖 Claude Code review — **approve** (validated on real kernel)

AAASM-4033 (follow-up to 4011). Test-only, only `aa-integration-tests/tests/e2e_ebpf.rs` touched — no runtime/ebpf source modified (drove the same `LoaderControlClient` orchestration the runtime uses, since `drive_ebpf_layer` is `pub(crate)`). Test 7 spawns the real loaderd + drives LoadProbeSet(Tls/FileIo/Exec/SyscallGuard) + UpdatePathMap + UpdateSyscallAllowlist; Test 8 asserts boot-liveness degrades on hung/absent daemon. SyscallGuard scoped to a throwaway sandbox PID (never the test process).

**Validation:** dispatched `ci.yml` on this branch — `e2e — Layer 3 eBPF (Linux)` + `eBPF probes build` both **PASS** on a real kernel. Closes the AAASM-4011 orchestration-validation gap; NOTE rewritten to what is covered vs still-deferred (telemetry-over-control-channel, SIGKILL-enforcement assertion).

PR #1402, opened 2026-07-04
[AAASM-4070] 🐛 (proxy): Fail closed on JSON-RPC batch to close MCP tools/call bypass

Chisanan232:
## What changed

`aa-proxy`'s MCP interception (`handle_non_llm_with_gateway`) enforced `tools/call`
policy and ran the credential/DLP scan only when `parse_mcp_request` produced a
single `McpToolCall`. That parser deserialises a **single** JSON-RPC object, so a
`tools/call` wrapped in a JSON-RPC 2.0 **batch array**
(`[{"jsonrpc":"2.0","method":"tools/call","params":{"name":"execute_bash",...}}]`)
— or an object malformed just enough that strict extraction fails — returned
`None`. The `else { None }` branch then forwarded the body upstream with **no
gateway CheckAction and no credential/DLP scan**. A steered agent could wrap a
denied tool call in a one-element batch to bypass Layer-2 policy entirely.

This PR adds a fail-closed detector, `intercept::mcp::is_unenforceable_tool_call`,
and wires it into the forward gate:

- A top-level JSON-RPC batch array with any `method == "tools/call"` element, or an
  object whose `method == "tools/call"` but whose envelope fails strict extraction,
  is now **rejected with a JSON-RPC `-32600` (Invalid Request) error envelope**
  instead of being forwarded upstream — mirroring an explicit gateway Deny, and
  emitting a denied `emit_mcp_decision` for audit.
- The detector is deliberately a *detector*, not a parser: it does not change
  `parse_mcp_request`'s existing `None`-on-false-positive contract.
- Ordinary non-MCP HTTPS traffic on this route is unaffected — the detector only
  fires on `method == "tools/call"`, so `tools/list`, non-JSON bodies, and every
  other request still flow through untouched.

## Why

Closes the MCP `tools/call` enforcement + DLP bypass via JSON-RPC batch framing
(AAASM-4070). The batch/malformed-framing path can no longer reach upstream
without a gateway decision and credential/DLP scan.

## How to verify

- `cargo nextest run -p aa-proxy intercept::mcp::tests` — 10 tests pass, including
  the new regression cases:
  - `one_element_batch_tools_call_is_flagged_unenforceable` (the exact bypass vector)
  - `multi_element_batch_with_a_tools_call_is_flagged`
  - `object_tools_call_that_fails_strict_extraction_is_flagged`
  - `non_tool_call_traffic_is_not_flagged` (tools/list object + batch, non-JSON,
    empty body, scalar array — all pass through)
- `cargo clippy -p aa-proxy --all-targets -- -D warnings` — clean.
- No new panic on malformed JSON: the detector treats any non-JSON body as `false`.

## Related Jira

AAASM-4070

### DIFF
--- a/aa-proxy/src/intercept/mcp.rs
+++ b/aa-proxy/src/intercept/mcp.rs
@@ -217,4 +217,76 @@ mod tests {
         assert_eq!(call.tool_name, "ping");
         assert_eq!(call.arguments, serde_json::Value::Null);
     }
+
+    #[test]
+    fn one_element_batch_tools_call_is_flagged_unenforceable() {
+        // AAASM-4070 regression: the bypass vector. `parse_mcp_request` only
+        // deserialises a single object, so this one-element batch returned
+        // `None` and the caller's pre-fix `else` branch forwarded it upstream
+        // with NO gateway CheckAction and NO credential/DLP scan. The detector
+        // must flag it so the caller fails closed instead of forwarding.
+        let body = br#"[
+            {
+                "jsonrpc": "2.0",
+                "id": 1,
+                "method": "tools/call",
+                "params": { "name": "execute_bash", "arguments": { "cmd": "rm -rf /" } }
+            }
+        ]"#;
+        assert!(
+            parse_mcp_request(body).is_none(),
+            "batch must not parse as a single call"
+        );
+        assert!(
+            is_unenforceable_tool_call(body),
+            "a one-element batch tools/call must be flagged so it is fail-closed, not forwarded"
+        );
+    }
+
+    #[test]
+    fn multi_element_batch_with_a_tools_call_is_flagged() {
+        // A tool call hidden among benign requests in a batch must still be
+        // flagged — any element carrying `method == "tools/call"` taints the
+        // whole batch, since the strict parser can enforce none of them.
+        let body = br#"[
+            {"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}},
+            {"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"read_file"}}
+        ]"#;
+        assert!(is_unenforceable_tool_call(body));
+    }
+
+    #[test]
+    fn object_tools_call_that_fails_strict_extraction_is_flagged() {
+        // `method == "tools/call"` but the envelope fails strict extraction
+        // (wrong jsonrpc version, missing `params.name`). `parse_mcp_request`
+        // returns `None` for these by contract; the detector must still flag
+        // them as tool-call attempts so a malformed envelope cannot slip a call
+        // past enforcement via the blind-forward path.
+        let wrong_version = br#"{"jsonrpc":"1.0","id":1,"method":"tools/call","params":{"name":"x"}}"#;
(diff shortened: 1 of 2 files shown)


Chisanan232:

**CI:** Green. 23/39 checks passed, 0 failed; the 16 non-run are `SKIPPED` by the `ci.yml` path router (dashboard, OpenAPI/proto/schema, eBPF/Linux e2e, TimescaleDB, SonarCloud, Coverage, Benchmark) — expected for a Rust-only `aa-proxy` change that touches no dashboard/proto/infra paths. No real failures.

**Scope vs AAASM-4070:** Covered. Both bypass variants are closed:
- **Batch array** — `is_unenforceable_tool_call` flags any top-level JSON-RPC array with an element whose `method == "tools/call"` (one-element and multi-element), the exact vector in the ticket.
- **Malformed `tools/call`** — a top-level object with `method == "tools/call"` that fails strict extraction (`parse_mcp_request` returns `None`) is still flagged (wrong `jsonrpc` version, missing `params.name`).

The detector is correctly wired into the *actual forward path*: it is an `else if` sitting between the `parse_mcp_request` enforcement branch and the terminal blind-forward `else { None }` in `handle_non_llm_with_gateway` (proxy/mod.rs:508). A flagged body now returns a JSON-RPC error and `return Ok(())` — it can no longer reach upstream without a gateway decision. It also emits a denied `emit_mcp_decision(..., true, reason)` for audit parity, matching the existing fail-closed branch.

**Side-effects:** Clean.
- No breakage of legitimate non-tool traffic: the detector fires *only* on `method == "tools/call"`. `tools/list` (object + batch), non-JSON, empty body, and scalar arrays all return `false` (covered by `non_tool_call_traffic_is_not_flagged`). Single valid `tools/call` still flows through the normal `parse_mcp_request` enforcement path unchanged.
- No new panic: `serde_json::from_slice::<Value>` on a non-JSON/huge/empty body is handled via `let Ok(..) else { return false }` — no unwrap, no new DoS surface beyond the body already buffered upstream.
- Fail-closed error mirrors a gateway Deny: HTTP 200 + JSON-RPC error envelope via the same `build_jsonrpc_error_response` helper the `-32000` Deny branch uses. `-32600` (Invalid Request) is the semantically correct code for unenforceable framing, and for an invalid batch a single error object with null id is spec-correct (the id can't be attributed).
- Note (accepted limitation, not a blocker): a client legitimately *batching* `tools/call` requests is now rejected. That is the correct security tradeoff — the enforcement path cannot parse batches, so failing closed is required by the ticket; MCP tool enforcement is single-request today.

**Coverage judgement:** Adequate. Testing `is_unenforceable_tool_call` directly rather than adding a `tonic` mock + full MitM integration test is acceptable here: the detector is the security-critical decision unit and is exercised against the exact bypass vector plus the false-positive guard, and the wiring is a trivial `else if` structurally identical to the proven `-32000` fail-closed branch immediately above it. Avoiding a new dev-dependency for this is the right call.

**Verdict:** Ready to approve/merge. Fix is correctly scoped, wired into the real forward path, fails closed on both variants, and does not regress legitimate traffic. No blocking concerns.

— reviewed by Claude Code

PR #1403, opened 2026-07-04
[AAASM-4071] 🐛 (scanner): Catch <=64-char base64 secrets in compact-JSON context

Chisanan232:
## Description

The credential scanner (`aa-security/src/scanner.rs`) missed base64 secrets ≤ 64 chars in a punctuation-delimited (compact-JSON) context, e.g. `{"api_token":"<64 base64 chars>"}`.

Two passes should have caught it, but neither did:

- **Pass 1 (whitespace-token entropy)** only inspects `split_whitespace()` tokens in the `20..=64` length window. A compact-JSON payload has no whitespace, so the whole thing is one token > 64 chars → excluded.
- **Pass 3 (base64-run)** gated on `run.len() > BASE64_RUN_MIN_LEN` (64, strict `>`). The quote-delimited run is exactly 64 chars → `64 > 64` is false → excluded.

The hex pass (pass 2) already uses `>= HEX_RUN_MIN_LEN`, so hex was fine; only the base64 run had the strict-`>` off-by-one plus a length floor set too high to catch delimited secrets. A 64-char base64 secret therefore survived `scan()` **clean** on the authoritative runtime enforce path.

**Fix:** lower `BASE64_RUN_MIN_LEN` from 64 to 20 (mirroring the token pass's per-run floor) and switch the comparison from `>` to `>=` (matching the hex pass). `ENTROPY_BITS_GATE = 4.5` is unchanged and is what bounds false positives — length is not the FP control, entropy is. Benign structured runs (32/40-char hex ids, dashed UUIDs, connection strings) stay below the gate and clean regardless of length.

Change is tightly scoped to the base64-run length gate + its constant.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-4071
- Note: sibling PR AAASM-4075 also edits `aa-security/src/scanner.rs` (separator-stripping in a different part of the file). This change is kept to the base64-run gate + constant so the two PRs do not overlap hunks.

## Testing

Added `detects_64_char_base64_secret_in_compact_json` — a regression asserting a 64-char high-entropy base64 secret in `{"api_token":"<64 b64>"}` is now both flagged (`GenericHighEntropy`) and fully redacted. Verified locally:

- `cargo fmt --all` — clean
- `cargo clippy -p aa-security --all-targets -- -D warnings` — 0 errors
- `cargo nextest run -p aa-security scanner` — 69 passed, incl. the new regression and the existing benign-FP tests (`does_not_flag_benign_hex_ids_or_prose`), confirming hex detection and the FP rate are not regressed.

## Checklist

### DIFF
--- a/aa-security/src/scanner.rs
+++ b/aa-security/src/scanner.rs
@@ -401,8 +401,37 @@ impl CredentialScanner {
         scan_azure_account_key(text, &mut findings);
 
         findings.sort_by_key(|f| f.offset);
+        dedupe_same_kind_overlaps(&mut findings);
         ScanResult { findings }
     }
+}
+
+/// Drop findings of the **same kind** whose byte span overlaps an
+/// already-kept finding of that kind, so a single secret caught by two passes
+/// of one detector is reported once.
+///
+/// The high-entropy detector runs additive passes (whitespace-token, long-hex,
+/// long-base64): a base64 secret that is *also* a whitespace token — e.g. a PEM
+/// body on its own line, or a bare `token=<b64>` — trips both the token pass and
+/// the base64-run pass, yielding two overlapping `GenericHighEntropy` findings
+/// for one secret. This collapses that double-count.
+///
+/// Only *same-kind* overlaps are removed: overlaps across different kinds are
+/// deliberately kept, because a connection URL legitimately produces distinct
+/// coincident findings (e.g. `PostgresUrl` + `GenericHighEntropy` + `EmailAddress`
+/// over the same region) that `redact` coalesces into one span but that callers
+/// count as separate detections. `findings` must already be sorted by `offset`,
+/// so any earlier same-kind finding `k` has `k.offset <= f.offset`; the spans
+/// therefore overlap exactly when `f.offset < k.end`.
+fn dedupe_same_kind_overlaps(findings: &mut Vec<CredentialFinding>) {
+    let mut kept: Vec<CredentialFinding> = Vec::with_capacity(findings.len());
+    for f in findings.drain(..) {
+        let covered = kept.iter().any(|k| k.kind == f.kind && f.offset < k.end);
+        if !covered {
+            kept.push(f);
+        }
+    }
+    *findings = kept;
 }
 
 // ---------------------------------------------------------------------------
--- a/aa-security/src/scanner.rs
+++ b/aa-security/src/scanner.rs
@@ -1188,6 +1188,32 @@ mod tests {
         );
     }
 
+    /// A 64-char base64 secret in punctuation-delimited (compact-JSON) context —
+    /// `{"api_token":"<64 b64>"}` — has no whitespace, so the whole payload is one
+    /// token > 64 chars that pass 1 skips, and the quote-delimited run is exactly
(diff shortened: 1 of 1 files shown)


Chisanan232:

**CI:** Green so far after the fix. The previously-**failing `Rust conformance suite` now passes** (1m40s), along with `Format check`, `Rustdoc`, `Dependency checks`, `Standalone build (git-SHA pin)`, and both `no_std` targets. `Clippy lint`, `Build`, and `Test` are still in progress but were validated clean locally (`cargo nextest run -p aa-security` → 112 passed; `cargo clippy -p aa-security --all-targets -D warnings` → 0 code warnings; `cargo fmt` clean). No job is red.

**Scope vs AAASM-4071:** On target. The core change lowers `BASE64_RUN_MIN_LEN` 64→20 and flips the run test `>`→`>=`, so the base64-run pass now catches a quote-/punctuation-delimited base64 secret whose contiguous run is ≤ 64 chars and carries no whitespace — the compact-JSON evasion (`{"api_token":"<64 b64>"}`) that pass-1's whitespace-token scan and the old `> 64` bound both skipped. Backed by a focused regression test (`detects_64_char_base64_secret_in_compact_json`). Doc comments for `GenericHighEntropy`, `scan_high_entropy`, and `BASE64_RUN_MIN_LEN` were updated to match.

**Side-effects:** Lowering the base64 floor made the additive base64-run pass re-flag secrets the whitespace-token pass had *already* flagged as `GenericHighEntropy` (bare `token=<b64>`, PEM bodies on their own line), producing two overlapping same-kind findings for one secret → the conformance finding-count vectors broke (double-count). The follow-up commit adds `dedupe_same_kind_overlaps()` at the end of `scan()`: after the offset sort, a finding is dropped only when its span overlaps an already-kept finding **of the same kind**. Rationale:
- It removes exactly the spurious duplicate (same detector, same secret) without touching detection breadth — the compact-JSON case still yields its one `GenericHighEntropy` finding, so nothing dedupes it away.
- Cross-kind coincident findings are **deliberately preserved** — a connection URL legitimately emits `PostgresUrl` + `GenericHighEntropy` + `EmailAddress` over the same bytes, which `redact` coalesces to one span but callers count as distinct detections. Blanket span-dedup would have wrongly collapsed those (3→1); same-kind scoping keeps them at 3.
- No FP spike: the `ENTROPY_BITS_GATE` (4.5 bits/char) — not length — bounds false positives, so benign hex ids / UUIDs / connection-string segments stay below the gate regardless of the lower length floor. Hex detection is intact (hex ~4.0 bits/char never trips the base64 pass; `HEX_RUN_MIN_LEN` unchanged at 64).
- All conformance vectors hold: `all_vectors_have_correct_finding_{count,kinds,offsets}` + `all_vectors_redact_correctly` all green (verified locally and in CI). RSA/EC PEM, multi-finding DB-URL, and Azure connection-string vectors are unaffected; redaction output is byte-identical (dedup only drops a finding whose span is already covered by the kept one).

**Verdict:** ✅ Approve. Minimal, principled, and correctly scoped — closes the compact-JSON base64 gap while the same-kind dedup cleanly resolves the double-count without narrowing detection or regressing any vector.

— reviewed by Claude Code

PR #1405, opened 2026-07-04
[AAASM-4075] 🔒 (auth,scanner): Bound pre-auth argon2 + catch separator-delimited secrets

Chisanan232:
## Description

Two defense-in-depth security hardening fixes (AAASM-4075):

**(a) Pre-auth argon2 grinding DoS — `aa-auth`.** `ApiKeyStore::validate_detailed` ran argon2id (~19 MiB, t=2) against *every* stored entry for a non-matching key, and the `RateLimiter` is keyed on `caller.key_id` — which only exists *after* successful validation. So an unauthenticated attacker bursting well-formed-but-invalid `Authorization: Bearer aa_<32-hex>` tokens forced N (= key count) argon2 verifications per request with no throttle → CPU/memory DoS. Fix: derive a fast, non-secret lookup index (truncated SHA-256) per key, persist it on `ApiKeyEntry`, and select candidates through an O(1) map so argon2 runs only on the single matching entry — a bogus token normally selects zero candidates and runs argon2 zero times. argon2 remains the sole authority for the final match; a lookup collision only ever adds one extra verification, never restoring the O(N) fan-out. Legacy entries without the index fall back to the prior scan, preserving backward compatibility.

**(b) Separator-delimited secret scanner evasion — `aa-security`.** A secret reformatted `de:ad:be:ef:…` splits the contiguous run at `:`/`-` separators: colons break both the hex and base64 runs, while `-` (a base64url char) keeps one run but dilutes its entropy below the gate — so such tokens evaded all three existing passes. Fix: an additive pass 4 scans each maximal `[0-9a-fA-F:-]` run, counts only the hex digits, and flags a separator-bearing run once its hex-digit count reaches the same 64-digit bar as the contiguous-hex rule.

## Type of Change

- [x] 🐛 Bug fix (security hardening)

## Breaking Changes

- [x] No

`ApiKeyEntry` gains an optional `key_lookup` field (`#[serde(default, skip_serializing_if)]`); existing key files without it load and authenticate unchanged via the fallback scan.

## Related Issues

- Related Jira ticket: AAASM-4075
- Sibling PR **AAASM-4071** also edits `aa-security/src/scanner.rs` (it changes the base64-run floor constant + comparison operator). This PR's scanner change is a **disjoint** additive hunk — a new `scan_separated_hex_runs` pass that does not touch the base64-floor lines. Both branch off `remote/master`; a small rebase at merge is expected.

## Testing

- `aa-auth`: `invalid_token_is_not_verified_against_mismatched_lookup_bucket` deterministically proves the O(N) fan-out is gone (an entry whose argon2 hash matches the probe but is filed under a mismatched bucket is never selected → `NotFound`); `legacy_entry_without_lookup_still_validates` guards backward compat. Full suite: 43/43 pass.
- `aa-security`: `detects_separator_delimited_hex_secret` (colon- and dash-grouped 64-hex secret flagged + redacted) and `does_not_flag_short_separated_hex` (MAC address + dash-UUID stay clean — no new false positives). Full scanner suite green; `aa-security` 156/156 pass.
- `cargo fmt --all` clean; `cargo clippy -p aa-auth -p aa-security --all-targets -- -D warnings` clean; `aa-api` auth tests 29/29 pass.

## Checklist

- [x] All CI checks passing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,6 +70,7 @@ dependencies = [
  "rand_core 0.6.4",
  "serde",
  "serde_json",
+ "sha2 0.11.0",
  "thiserror 2.0.18",
  "tracing",
  "utoipa",
--- a/aa-api/src/state.rs
+++ b/aa-api/src/state.rs
@@ -434,6 +434,9 @@ impl AppState {
                     label: Some("local single-process admin key".to_string()),
                     team_id: None,
                     org_id: None,
+                    // AAASM-4075 — index this key so credential validation runs
+                    // argon2 only on the matching candidate, not per-key.
+                    key_lookup: Some(parsed.lookup()),
                 };
                 state.auth_config = Arc::new(AuthConfig {
                     mode: AuthMode::On,
--- a/aa-api/tests/auth_jwt.rs
+++ b/aa-api/tests/auth_jwt.rs
@@ -156,6 +156,7 @@ async fn issued_jwt_retains_caller_tenant() {
         label: Some("tenant key".to_string()),
         team_id: Some("alpha".to_string()),
         org_id: Some("org-1".to_string()),
+        key_lookup: Some(key.lookup()),
     };
     let app = common::test_app_with_auth(&[entry], 1000);
 
--- a/aa-api/tests/common/mod.rs
+++ b/aa-api/tests/common/mod.rs
@@ -246,6 +246,7 @@ pub fn generate_test_api_key(id: &str, scopes: Vec<Scope>) -> (String, ApiKeyEnt
         label: Some(format!("test key {id}")),
         team_id: None,
         org_id: None,
+        key_lookup: Some(key.lookup()),
     };
     (key.as_str().to_string(), entry)
 }
--- a/aa-auth/Cargo.toml
+++ b/aa-auth/Cargo.toml
@@ -16,6 +16,11 @@ rand = "0.10"
 rand_core = { version = "0.6", features = ["getrandom"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
+# Fast lookup-hash for the API-key prefix index (AAASM-4075): a truncated SHA-256
+# of the raw key selects the single candidate entry so `validate_detailed` runs
+# argon2 once, not once per stored key, on unauthenticated input. sha2 is a
+# workspace dependency already pulled in transitively by argon2.
+sha2 = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 utoipa = { workspace = true }
--- a/aa-auth/src/api_key.rs
+++ b/aa-auth/src/api_key.rs
@@ -457,4 +457,63 @@ mod tests {
         assert!(store.is_empty());
         assert_eq!(store.len(), 0);
     }
+
+    /// AAASM-4075 (anti-fan-out DoS): `validate_detailed` must argon2-verify only
+    /// the entries selected by the incoming key's lookup bucket, never every
+    /// stored key. We prove the fan-out is gone by filing an entry whose argon2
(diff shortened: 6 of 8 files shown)


Chisanan232:

**CI:** Green. 31 passed, 0 failed (remaining checks skipped/path-filtered or coverage/sonar-only, which are non-blocking). Branch reports mergeable.

**Scope vs AAASM-4075:** Both items covered.
- **(a) argon2 fan-out** — `ApiKey::lookup()` derives a non-secret truncated SHA-256 (64-bit / 16-hex) index; `ApiKeyStore` builds a `HashMap<lookup, Vec<usize>>` at construction and `validate_detailed` now argon2-verifies only the lookup-selected candidates (plus any legacy/unindexed entries), not every stored key. A bogus token normally selects zero candidates → argon2 runs zero times. argon2 (`key.verify`) stays the sole final authority. Directly closes the O(N) pre-auth grinding path.
- **(b) separator-delimited hex** — additive Pass 4 (`scan_separated_hex_runs`) scans each maximal `[0-9a-fA-F:-]` run, counts only hex digits, and flags once a separator-bearing run reaches the same 64-hex bar as the contiguous rule. Closes the `de:ad:be:ef…` / `de-ad-be-ef…` evasion.

**Side-effects:**
- *Auth correctness / backward-compat:* A valid key recomputes the same `lookup()` → its bucket is selected → argon2 matches (proven by existing validate tests). Legacy entries with `key_lookup: None` fall into the always-scanned `unindexed` set and still authenticate — covered by `legacy_entry_without_lookup_still_validates`. `#[serde(default, skip_serializing_if)]` keeps old key files loadable with no migration. Constructors funnel through `from_entries`/`build_index`, keeping the two index structures in sync.
- *Timing/enumeration oracle:* The bucket is SHA-256 of the **full** key, not a guessable prefix — an attacker cannot target a non-empty bucket without already knowing the key. So the "fast NotFound" path leaks nothing beyond an inherent ~2^64 collision (which only costs one extra argon2 verify, never restoring fan-out). This is stronger than a prefix index. Constant-time final compare via argon2 is preserved. No new oracle.
- *Scanner FP:* MAC (12 hex) and dash-UUID (32 hex) stay under the 64-hex bar — covered by `does_not_flag_short_separated_hex`. One residual note: a colon-separated **SHA-256 certificate/key fingerprint** is exactly 64 hex digits and will now be flagged as `GenericHighEntropy` — but this is the *same* tradeoff the pre-existing contiguous-64-hex rule already makes (a bare SHA-256 hash already trips it), so it introduces no new FP class relative to current policy. Acceptable; worth being aware of.
- *Disjoint from #1403 (AAASM-4071):* Semantically disjoint — #1403 lowers the contiguous **base64**-run floor (`>` → `>=`, ≥20); this PR adds a separate **hex-only** pass. Dash-grouped hex still fails #1403's entropy gate, so no double-count in practice. **However there is a textual overlap:** both edit the shared numbered doc-comment block above `scan_high_entropy` (#1403 rewords pass 3; this PR appends pass 4). Expect a trivial rebase conflict in that comment for whichever merges second — the author already calls this out. No conflicting *code* hunks.

**Test adequacy:** Strong. `invalid_token_is_not_verified_against_mismatched_lookup_bucket` is a clean deterministic proof the fan-out is gone (a trap entry whose argon2 hash matches the probe but filed under a wrong bucket is never reached → `NotFound`). Backward-compat, separator detection (colon + dash, end-to-end redaction), and no-FP cases are all covered.

**Verdict:** ✅ Approve. Both AAASM-4075 items are correctly and minimally fixed, backward-compatible, well-tested, and introduce no new timing oracle or FP class. Only follow-up: resolve the trivial doc-comment rebase against #1403 at merge (sequence the two merges, re-apply the pass-4 bullet).

— reviewed by Claude Code

PR #1406, opened 2026-07-04
[AAASM-4093] 🐛 (scanner): Extend redaction span on same-kind overlap dedupe

Chisanan232:
## Description

Fixes a redaction-span shrink regressed by `bf3e677a` in `aa-security`'s credential scanner. `dedupe_same_kind_overlaps` dropped an overlapping same-kind finding outright instead of merging its span. When a run of ≥64 hex chars is followed by a non-hex base64 tail (whole-run Shannon entropy > 4.5), pass-2 (long-hex, span `[start, 64)`) and pass-3 (base64-run, span `[start, 64+K)`) both emit a `GenericHighEntropy` finding at the same offset. The shorter hex finding sorts first and was kept; the longer base64 finding was dropped, so `redact()` — which coalesces only the surviving findings — forwarded bytes `[64, 64+K)` **un-redacted** on the sanitize-and-forward path.

The fix merges the overlapping finding into the survivor by extending its end to the union (`k.end = max(k.end, f.end)`) rather than dropping it. The reported finding count stays one per overlapping same-kind cluster (unchanged from the AAASM-4071 count-dedup behavior); only the span widens so redaction covers the full run.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-4093

## Testing

Validation performed in the worktree:
- `cargo fmt --all` — clean
- `cargo clippy -p aa-security --all-targets -- -D warnings` — clean (via lefthook, workspace `--all-targets`)
- `cargo nextest run -p aa-security scanner` — 72 passed, incl. new `redact_covers_base64_tail_after_long_hex_run`
- `cargo nextest run -p conformance --test credential_detection` — 8 passed (all finding-count/offset/kind/redaction vectors hold)
- Confirmed the new test **fails** on pre-fix code (tail `GHIJKLMNOPQRSTUVWXYZ` survives redaction) and **passes** with the fix.

## Checklist

- [x] Documentation updated if behaviour changed (dedupe doc comment)

### DIFF
--- a/aa-security/src/scanner.rs
+++ b/aa-security/src/scanner.rs
@@ -1429,6 +1429,49 @@ mod tests {
         assert!(redacted.contains("[REDACTED:GenericHighEntropy]"));
     }
 
+    /// AAASM-4093: a `<64-hex><base64-tail>` run trips both the long-hex pass
+    /// (span `[start, 64)`) and the base64-run pass (span `[start, 64+K)`). Both
+    /// are `GenericHighEntropy` at the same offset; the shorter hex finding sorts
+    /// first. The same-kind dedupe must *widen* the survivor's span to the union
+    /// rather than drop the longer base64 finding, or `redact` forwards the tail
+    /// bytes `[64, 64+K)` in the clear. Assert the full run is masked and that the
+    /// secret is still counted exactly once.
+    #[test]
+    fn redact_covers_base64_tail_after_long_hex_run() {
+        let scanner = CredentialScanner::new();
+        // 64 hex digits followed by a non-hex base64 tail; the whole contiguous
+        // run is base64 and its Shannon entropy clears the gate, so the base64
+        // pass spans the full 84 chars while the hex pass stops at 64.
+        let hex = "deadbeefcafebabe0123456789abcdef0123456789abcdeffedcba9876543210";
+        let tail = "GHIJKLMNOPQRSTUVWXYZ";
+        let secret = format!("{hex}{tail}");
+        assert_eq!(hex.len(), 64);
+        assert!(!tail.is_empty(), "tail must add bytes beyond the 64-hex span");
+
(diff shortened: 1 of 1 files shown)


Chisanan232:

**Scope** — Two commits, single file (`aa-security/src/scanner.rs`): a one-branch change to `dedupe_same_kind_overlaps` (drop → merge/union of the span) plus its doc comment, and one new unit test. No public API/signature change (`CredentialFinding.end` stays private; `matched`/`offset`/`kind` unchanged). Tightly scoped to the reported regression.

**CI / validation** — `cargo fmt`, `cargo clippy -p aa-security --all-targets -D warnings`, `cargo nextest run -p aa-security scanner` (72 pass), and `cargo nextest run -p conformance --test credential_detection` (8 pass) all green locally via lefthook (pre-commit fmt+clippy, pre-push doc). GitHub Actions may abort on the org billing block — validated locally instead.

**Side-effects** — The dedupe now *widens* a kept finding's `end` to the union of overlapping same-kind spans instead of discarding the overlapping finding. This is strictly more coverage for `redact()`; it cannot leak more (only mask more) and does not change the reported finding **count** (still one per overlapping same-kind cluster), so the AAASM-4071 count contract and all conformance count/offset/kind/redaction vectors are preserved. Cross-kind overlaps are untouched. No behavioral change to any other pass.

**Regression guard** — New test `redact_covers_base64_tail_after_long_hex_run` asserts (a) exactly one `GenericHighEntropy` finding for a `<64-hex><base64-tail>` run and (b) the full run including tail bytes `[64, 64+K)` is masked. Verified it **fails** on pre-fix code (tail forwarded in the clear) and **passes** with the fix.

**Verdict** — ✅ LGTM. Minimal, correct, security-positive (closes an un-redacted-secret forward path), fully covered by a regression test with counts held. Ready for Pioneer review.

PR #1408, opened 2026-07-04
[AAASM-4103] ✨ (policy): Split file delete into a distinct FileDelete capability

Chisanan232:
## Description

Splits file **delete** into its own capability verb. Today `FileMode::Delete`
maps to `Capability::FileWrite`, so a policy cannot allow writes while denying
deletes. This PR introduces a distinct `Capability::FileDelete` so file deletion
is independently governable.

What changed:
- New `Capability::FileDelete` variant with the `file_delete` wire string
  (FromStr/Display) in both `aa-core` and the `aa-security` canonical policy AST,
  plus the lossless gateway→canonical bridge arm.
- `action_to_capability(FileMode::Delete)` now returns `FileDelete` (was
  `FileWrite`).
- `capability_is_denied` helper wires both gateway capability-evaluation sites
  (per-doc stage + cascade guard) so delete is a first-class deny/allow verb.
- eBPF lowering seeds the kernel sensitive-path deny defaults on a `file_delete`
  deny as well as `file_write` (the path probe can't distinguish the verbs).
- Docs + a new `policy-examples/file-delete-policy.yaml` demonstrating
  allow-write / deny-delete.

The capability evaluation, cascade merge, and validator are all generic over the
`Capability` enum, so once the vocabulary and the delete→FileDelete mapping are
in place, deny/allow/cascade all treat `FileDelete` as first-class with no other
changes.

## Type of Change

- [x] ✨ New feature

## Breaking Changes

- [x] Yes (describe below)

**Behavioral change (policy semantics), fail-closed in both directions:**

1. A `file_write` **allow** no longer implies delete. A delete action is denied
   unless the policy explicitly grants `file_delete`. Policies that previously
   relied on a `file_write` allow to permit deletes must add `file_delete` to
   their allow list.
2. A `file_write` **deny** still blocks delete (defense in depth, via
   `capability_is_denied`). Policies that denied `file_write` to lock down all
   mutations keep blocking deletes even if they never name `file_delete`.

Net effect: the new verb can only ever make delete **more** restricted than
before, never less — a stale write grant can never silently allow a delete.

No public Rust API is removed; `Capability` gains a variant (exhaustive matches
over it are updated in-tree).

## Related Issues

- Related Jira ticket: AAASM-4103
- Note: sibling ticket AAASM-4099 will wire the remaining unmapped capabilities
  in `action_to_capability`; this PR is scoped to delete only.

## Testing

- [x] Unit tests added / updated
- [x] Integration tests added / updated
[...]

PR #1409, opened 2026-07-04
[AAASM-4099] 🔒 (policy): Enforce or fail-loud on previously-unmapped capabilities

Chisanan232:
## Description

`Capability::NetworkInbound`, `Capability::Model`, and `Capability::AgentSpawn` had **no `GovernanceAction` mapping** in `action_to_capability`, and no action variant (nor any runtime/proxy/eBPF enforcement path) routes to them. Because `stage_capability` only denies a capability that some `GovernanceAction` produces, a policy declaring `allow`/`deny` on these three passed validation and hot-loaded **silently**, while the control was entirely inert — an operator who wrote `deny: [network_inbound]` (as the shipped `policy-examples/strict.yaml` does) or `deny: [agent_spawn]` believed they had restricted the agent when nothing was enforced.

Per-capability investigation outcome — **all three are currently unenforceable** (no wireable action exists; `AgentSpawn` exists only as an audit-side `ActionType`, never as a policy-evaluated `GovernanceAction`), so this is a **fail-loud** fix rather than a wiring fix:

- **`aa-core`** — added `Capability::is_enforceable()` as the single source of truth for "does any `GovernanceAction` route to this capability", kept in lock-step with `action_to_capability` (documented invariant). Returns `false` for `Model`, `NetworkInbound`, `AgentSpawn`.
- **`aa-gateway`** — `validate_capabilities` now emits a loud `ValidationWarning::inert_capability` for any declared allow/deny capability whose `is_enforceable()` is `false`. The policy still **loads** (a hard error would break the deliberately-shipped `strict.yaml` deny of `network_inbound` and the policy-examples corpus), but the inertness is now surfaced to the operator (dashboard ValidationPanel / CLI) instead of hidden.

A hard `ValidationError` was considered and rejected: the shipped corpus and two corpus-parsing test suites (`aa-security::policy_examples_parse`, `aa-gateway::cross_layer_policy_consistency_test`) intentionally use `network_inbound` as a capability floor, so a warning + doc is the principled fail-loud choice here.

## Type of Change

- [x] 🐛 Bug fix (security: silently-inert control)

## Breaking Changes

- [x] No — policies that previously loaded still load; they now additionally carry a validation warning.

## Related Issues

- Related Jira ticket: AAASM-4099
- Note: sibling PR **#1408** (AAASM-4103) also edits `aa-core/src/capability.rs` (adds a `FileDelete` variant + `Delete → FileDelete` mapping). This PR touches only the `is_enforceable`/`action_to_capability` NOTE and a new `impl Capability` block — different match arms, so the hunks are disjoint.

## Testing

Local validation (no full suite, per scope):
- `cargo fmt --all` — clean
- `cargo clippy -p aa-core -p aa-gateway --all-targets -- -D warnings` — clean (only pre-existing Cargo manifest `default-features` warnings)
- `cargo nextest run -p aa-core -p aa-gateway` (new + related): **10/10 pass** — `is_enforceable_*`, `action_to_capability_only_yields_enforceable_capabilities`, `capabilities_inert_deny_parses_but_warns_loudly`, `capabilities_inert_allow_warns_loudly`, `capabilities_enforceable_caps_produce_no_inert_warning`, `validation_warning_inert_capability_is_loud`
- Regression: `policy_examples_parse` + `cross_layer_policy_consistency_test` + `capability_integration_test`: **13/13 pass** (strict.yaml with `network_inbound` still loads).

## Checklist

- [ ] All CI checks passing

### DIFF
--- a/aa-core/src/capability.rs
+++ b/aa-core/src/capability.rs
@@ -39,6 +39,29 @@ pub struct CapabilitySet {
     pub deny: BTreeSet<Capability>,
 }
 
+impl Capability {
+    /// Whether declaring this capability in a policy actually governs anything.
+    ///
+    /// A capability is *enforceable* only if some [`crate::GovernanceAction`]
+    /// maps to it via [`action_to_capability`] — otherwise no action ever routes
+    /// to it, so a declared allow/deny is silently inert and gives the operator a
+    /// false sense of security (AAASM-4099).
+    ///
+    /// Currently inert (no corresponding action variant): [`Capability::Model`],
+    /// [`Capability::NetworkInbound`], and [`Capability::AgentSpawn`]. This
+    /// predicate is the single source of truth policy validation uses to warn
+    /// loudly when a policy references one of them; keep it in lock-step with
+    /// [`action_to_capability`] — when a new action lands that maps to one of
+    /// these, wire the mapping there and drop it from the inert arm below.
+    #[must_use]
+    pub fn is_enforceable(&self) -> bool {
+        !matches!(
+            self,
+            Capability::Model(_) | Capability::NetworkInbound | Capability::AgentSpawn
+        )
+    }
+}
+
 impl FromStr for Capability {
     type Err = String;
 
@@ -130,8 +153,11 @@ pub fn action_to_capability(action: &crate::GovernanceAction) -> Option<Capabili
     use crate::GovernanceAction;
 
     // NOTE: Capability::AgentSpawn, NetworkInbound, and Model variants have no
-    // corresponding GovernanceAction yet. When new action variants land, add
-    // mappings here to avoid silent policy bypasses.
+    // corresponding GovernanceAction yet — they are flagged by
+    // `Capability::is_enforceable` so policy load warns loudly instead of
+    // presenting a silently-inert control (AAASM-4099). When new action variants
+    // land, add mappings here AND drop the variant from `is_enforceable`'s inert
+    // arm to avoid silent policy bypasses.
     match action {
         GovernanceAction::ToolCall { name, .. } => Some(Capability::McpTool(name.clone())),
         GovernanceAction::ToolResult { tool_name, .. } => Some(Capability::McpTool(tool_name.clone())),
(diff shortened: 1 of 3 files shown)


Chisanan232:

**Verdict: no silently-inert control remains.** Every capability an operator can declare now either (a) is enforced by `stage_capability` via a real `GovernanceAction` mapping, or (b) triggers a loud `capabilities.*` validation warning at load time saying it is "NOT enforced ... currently inert." Nothing is both declarable and silent.

### Security analysis
- **Root cause addressed:** `Capability::{Model, NetworkInbound, AgentSpawn}` have no `GovernanceAction` producing them, so `action_to_capability` returns `None` for every action and `stage_capability` can never deny on them. Confirmed no runtime/proxy/eBPF path enforces them either (`AgentSpawn` appears only as an audit `ActionType`, never a policy-evaluated action).
- **Fail-loud, not fail-open:** declaring these no longer passes review silently — the operator gets a warning surfaced through the same channel as unknown-key warnings (dashboard ValidationPanel / CLI).
- **No panic risk:** `is_enforceable` is a total `matches!` over the enum; no `unwrap`/`todo!`/`unreachable!` added. All variants handled.
- **Single source of truth:** `is_enforceable` is documented to stay in lock-step with `action_to_capability`; a co-located aa-core test (`action_to_capability_only_yields_enforceable_capabilities`) asserts the invariant so future action-wiring can't silently drift.

### Why warning over hard error
A hard `ValidationError` would reject the shipped `policy-examples/strict.yaml` (deliberate `deny: [network_inbound]` capability floor) and break two corpus-parse test suites. Warning + doc is the correct fail-loud posture that does not regress intended policies. When a real action later lands for any of these caps, wire it in `action_to_capability` and drop the variant from `is_enforceable`'s inert arm — the warning then disappears automatically and enforcement engages.

### Conflict note
Disjoint from sibling **#1408** (AAASM-4103 `FileDelete`): that PR edits the `FileMode`/`Delete` match arms; this PR adds a new `impl Capability` block + edits the `NOTE` comment. No overlapping hunks.

### Tests
10 new/related unit tests + 13 regression tests green locally. Coverage includes both allow-list and deny-list inert paths, the negative case (enforceable caps emit no warning), and load-still-succeeds.

No blocking issues found.

PR #1413, opened 2026-07-04
[AAASM-4123] 🐛 (gateway): Enforce capability stage on the single-file/primary policy path

Chisanan232:
## Description

`evaluate_primary` — the single-file / primary policy pipeline `evaluate` takes whenever the `scope_index` is empty — ran the schedule, network, tool, credential, and budget stages but **had no capability stage**. Capability enforcement lived only in `cascade_capability_guard`, called exclusively on the directory-cascade path. The single-file loaders (`load_from_file` / `load_from_file_with_budget`) build an empty `ScopeIndex`, so every single-file surface — `aasm policy simulate`, the aa-api `PolicyEngine`, and a gateway configured with a single-file `policy_path` — silently permitted every capability-denied action (per-op read / write / delete unenforced). Same fail-open-by-omission class as the network (AAASM-3728), tool-scope (AAASM-3981), and schedule (AAASM-3847) primary-path gaps.

This PR:
1. Extracts the deny/allow-list check out of `cascade_capability_guard` into a free-standing `capability_guard(caps, action)` helper, so both paths apply identical fail-closed logic and cannot diverge again (the pattern used to close AAASM-3728).
2. Wires a capability stage (Stage 2b) into `evaluate_primary`, evaluating the primary doc's `capabilities` via the shared helper. It runs before the tool stages so a capability-denied action never consumes a rate-limit token.
3. Adds regression tests loading a single-file strict policy through the public `load_from_file` loader.

The capability model itself was already correct (delete is its own verb; `capability_is_denied` carries write-deny⇒delete-deny defense-in-depth) — only the primary-path wiring was missing.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

Policies with no `capabilities` block are unaffected (empty set imposes no restriction). Only policies that already declared capability denies/allows — previously ignored on the single-file path — now enforce them, as intended.

## Related Issues

- Related Jira ticket: AAASM-4123

## Testing

New tests (`aa-gateway engine::tests`):
- `single_file_strict_policy_enforces_capability_stage` — a single-file strict floor (deny `terminal_exec` + `file_write`) denies `ProcessExec`, `FileAccess::Write`, and `FileAccess::Delete` (delete via write-deny⇒delete-deny), a non-allowlisted GET still denies (proving the primary path runs), and a read is permitted.
- `single_file_file_delete_deny_blocks_delete_on_primary_path` — a `file_delete`-only deny blocks delete yet permits writes on the primary path.

Behavioral verification with the shipped example, before vs. after:

```
$ aasm policy simulate --policy policy-examples/strict.yaml --against probe.jsonl

EVENT#   ACTION                              DECISION   REASON
0        net:GET:https://evil.example.com/   deny       host not in network allowlist
1        exec:id                             deny       capability denied by policy   # was ALLOW before
2        file:Write:/etc/passwd              deny       capability denied by policy   # was ALLOW before
3        file:Delete:/etc/passwd             deny       capability denied by policy   # was ALLOW before
(file:Read:/tmp/ok.txt -> allowed)
```

Full `cargo nextest run -p aa-gateway` green; `cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Checklist

Closes AAASM-4123.

### DIFF
--- a/aa-gateway/src/engine/mod.rs
+++ b/aa-gateway/src/engine/mod.rs
@@ -1108,6 +1108,19 @@ impl PolicyEngine {
             return result;
         }
 
+        // Stage 2b — Capability authorization (AAASM-4123). Mirrors the cascade
+        // path's `cascade_capability_guard` via the shared `capability_guard`
+        // helper so the single-file/primary surface (`aasm policy simulate`,
+        // aa-api, single-file gateway) enforces per-op read/write/delete instead
+        // of silently permitting every capability-denied action. Runs before the
+        // tool stages so a capability-denied action never consumes a rate-limit
+        // token. A doc with no `capabilities` block imposes no restriction.
+        if let Some(caps) = &policy.capabilities {
+            if let Some(result) = Self::capability_guard(caps, action) {
+                return result;
+            }
+        }
+
         // Stages 3-5b — Tool/message rules: allow/deny, rate limit, approval.
         if let Some(result) = self.eval_tool_stages(&policy, ctx, action) {
             return result;
--- a/aa-gateway/src/engine/mod.rs
+++ b/aa-gateway/src/engine/mod.rs
@@ -1166,11 +1166,29 @@ impl PolicyEngine {
         action: &aa_core::GovernanceAction,
     ) -> Option<EvaluationResult> {
         let merged_caps = Self::collect_merged_capabilities(cascade);
+        Self::capability_guard(&merged_caps, action)
+    }
+
+    /// Capability authorization gate shared by BOTH the primary
+    /// (`evaluate_primary`) and cascade (`cascade_capability_guard`) paths: deny
+    /// when `caps.deny` blocks the action's capability, or a non-empty
+    /// `caps.allow` omits it. `None` when the action maps to no capability or is
+    /// permitted.
+    ///
+    /// Extracted so the single-file and directory-cascade paths can never
+    /// diverge again (AAASM-4123): the primary path previously ran every stage
+    /// EXCEPT this one, silently permitting capability-denied actions on the
+    /// single-file surface (`aasm policy simulate`, aa-api, single-file
+    /// gateway). This is the same fail-open-by-omission class closed for the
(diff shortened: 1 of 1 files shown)


Chisanan232:
## Review — Approve-ready (comment only; human Pioneer approval still required)

**Verdict: Approve-ready.** Closes the HIGH CRUD-authz fail-open in AAASM-4123.

- **CI:** Green — 25/25 checks pass, 0 real failures.
- **Scope vs ticket:** Fully covers the fix + tests. Adds a capability stage (Stage 2b) to `evaluate_primary`, placed before the tool stages so a capability-denied action never consumes a rate-limit token. The fix is done the right way: the cascade path's inline capability check is extracted into a shared `capability_guard(caps, action)` helper that BOTH `evaluate_primary` (single-file) and `cascade_capability_guard` (directory cascade) now call, so the two paths can never diverge again — same shared-helper pattern used to close the network gap (AAASM-3728). Two regression tests assert the single-file strict.yaml path now denies process-exec / file-write / file-delete and that a `file_delete`-only deny blocks delete while still permitting writes.
- **Side effects:** Cascade path behavior is **byte-identical** — `cascade_capability_guard` still folds `merged_caps` and delegates to the extracted helper with the same deny/allow-omission logic (verified in source); all cascade tests pass. Legitimate ops still pass: a doc with no `capabilities` block imposes no restriction, and `file_read` under an empty allow-list is not blocked (asserted by the new test). No too-aggressive denials.
- **FE:** N/A.

**Cross-PR note:** #1416 (AAASM-4124) also edits `aa-gateway/src/engine/mod.rs` (a different function). Whichever of #1413/#1416 merges **second** will likely need a trivial rebase on `engine/mod.rs`.

**Local validation** (worktree, `RUSTUP_TOOLCHAIN=stable`): `cargo nextest run -p aa-gateway` → **1373 passed, 0 failed** (1 leaky; `policy_latency` p99 passed); `cargo fmt --all --check` clean; `cargo clippy -p aa-gateway --all-targets -D warnings` → 0 errors.

— Claude Code

PR #1418, opened 2026-07-04
[AAASM-4127] 🐛 (proxy): Wire CLI/runtime gateway endpoint into aa-proxy for MCP enforcement

Chisanan232:
## What changed

`aasm proxy start --gateway <url>` exported `AA_GATEWAY_URL` to the spawned `aa-proxy` child, but aa-proxy reads its gateway endpoint from `AA_PROXY_GATEWAY_ENDPOINT` (it does not reference `AA_GATEWAY_URL` anywhere). The child therefore always saw `gateway_endpoint = None` and ran in raw passthrough with MCP `tools/call` enforcement silently **OFF**, even though the operator had configured a gateway (and startup even fail-closes when the gateway is unreachable — giving false assurance).

Two producer-side call sites are fixed:

- **`aa-cli` (`proxy/start.rs`)** — export `AA_PROXY_GATEWAY_ENDPOINT` (the name aa-proxy reads) instead of `AA_GATEWAY_URL`, plus `AA_PROXY_LLM_ONLY=false`. The env mapping is factored into a pure `proxy_child_env()` helper.
- **`aa-runtime` (`runtime.rs`)** — `spawn_proxy` previously launched aa-proxy with no environment; now it forwards the runtime's `gateway_endpoint` as `AA_PROXY_GATEWAY_ENDPOINT` (and `AA_PROXY_LLM_ONLY=false`) when configured.

Setting `AA_PROXY_LLM_ONLY=false` is required alongside the endpoint: with the default `llm_only=true`, non-LLM MCP hosts (`LlmApiPattern::Unknown`) are transparently tunnelled in `proxy/mod.rs` *before* `handle_non_llm_with_gateway` runs, so the gateway routing is never reached. Disabling `llm_only` when a gateway is configured makes those MCP flows get intercepted and enforced.

## Why

Silent fail-open on a governance path (MED). An operator who configures a gateway reasonably expects MCP enforcement, but the env-var name mismatch left it entirely off; only the advisory in-process SDK fast-path remained, which a non-SDK or compromised agent evades.

## How to verify

- `cargo nextest run -p aa-cli` — includes new regression tests `proxy_child_env_gateway_uses_proxy_endpoint_var` and `proxy_child_env_omits_gateway_vars_when_absent`, which assert `--gateway` maps to `AA_PROXY_GATEWAY_ENDPOINT` (+ `AA_PROXY_LLM_ONLY=false`) and never the ignored `AA_GATEWAY_URL`.
[...]

PR #1419, opened 2026-07-04
[AAASM-4126] 🐛 (proxy): Run credential-DLP on all MitM'd request bodies

Chisanan232:
## What changed

`aa-proxy` credential-DLP (body credential scanning) previously ran **only** inside `handle_llm_mitm`, which was entered only when `detect_api(host)` matched one of the three hard-coded LLM providers (`api.openai.com`, `api.anthropic.com`, `api.cohere.com`). A secret in a request body to any other provider was never scanned:

- Under the default `llm_only=true`, every other host is transparent-tunnelled (no MitM, no scan).
- Under `llm_only=false`, other hosts are MitM'd but routed to the non-LLM path, which did MCP enforcement (or a raw byte-copy) but **never** ran the credential scanner.

This PR:

1. Adds an operator-configurable **`mitm_hosts`** set to `ProxyConfig` (env `AA_PROXY_MITM_HOSTS`, egress-allowlist grammar). A new `should_mitm(host)` gate brings these hosts under MitM even while `llm_only` stays `true`, so operators can extend the DLP surface to more providers without disabling `llm_only`.
2. Generalises `handle_non_llm_with_gateway` → `handle_non_llm_mitm`, taking an `Option<gateway>`, and runs `intercept_request` (credential-DLP) on **every** forwarded MitM'd request body. A `Block` verdict returns 403 without dialing upstream; a redact verdict forwards the redacted bytes. This also closes the previous no-gateway raw-copy path that forwarded bodies un-inspected.

Behaviour for the three built-in LLM hosts (scan + credential injection) is unchanged.

## Why

_Severity: MED (DLP bypass)._ DLP was bypassable purely by provider choice — a steered agent could exfiltrate a secret in a request body to any provider outside the 3-host allowlist (Google, Mistral, Groq, Azure OpenAI, Bedrock, OpenRouter, xAI, DeepSeek, …). Sweep AAASM-4120; verified in source.

## How to verify

- `cargo nextest run -p aa-proxy` — 200 pass.
- New regression test `attacker::credential_dlp_blocks_secret_to_non_llm_mitm_host`: a secret POSTed to a MitM'd host outside the built-in `detect_api` set (host added via `mitm_hosts`, `llm_only=true`) is refused with 403 and upstream is never dialed.
- New config tests pin `AA_PROXY_MITM_HOSTS` parsing.

Closes AAASM-4126

### DIFF
--- a/aa-integration-tests/tests/e2e_mcp_interceptor.rs
+++ b/aa-integration-tests/tests/e2e_mcp_interceptor.rs
@@ -804,6 +804,7 @@ mod proxy_e2e {
             ca_dir: ca_dir.to_path_buf(),
             cert_cache_capacity: 10,
             llm_only: false,
+            mitm_hosts: Vec::new(),
             denied_hosts: Vec::new(),
             network_allowlist: Vec::new(),
             skip_upstream_tls_verify: true,
--- a/aa-integration-tests/tests/e2e_policy_proxy.rs
+++ b/aa-integration-tests/tests/e2e_policy_proxy.rs
@@ -31,6 +31,7 @@ fn proxy_config(ca_dir: &std::path::Path, denied_hosts: Vec<String>) -> ProxyCon
         ca_dir: ca_dir.to_path_buf(),
         cert_cache_capacity: 10,
         llm_only: false,
+        mitm_hosts: Vec::new(),
         denied_hosts,
         network_allowlist: Vec::new(),
         skip_upstream_tls_verify: true,
--- a/aa-integration-tests/tests/e2e_secret_interception.rs
+++ b/aa-integration-tests/tests/e2e_secret_interception.rs
@@ -564,6 +564,7 @@ mod proxy_data_path {
             ca_dir: ca_dir.to_path_buf(),
             cert_cache_capacity: 10,
             llm_only: false,
+            mitm_hosts: Vec::new(),
             denied_hosts: Vec::new(),
             network_allowlist: Vec::new(),
             skip_upstream_tls_verify: true,
--- a/aa-integration-tests/tests/e2e_tool_sandbox.rs
+++ b/aa-integration-tests/tests/e2e_tool_sandbox.rs
@@ -43,6 +43,7 @@ fn proxy_config(ca_dir: &std::path::Path, network_allowlist: Vec<String>) -> Pro
         ca_dir: ca_dir.to_path_buf(),
         cert_cache_capacity: 10,
         llm_only: false,
+        mitm_hosts: Vec::new(),
         denied_hosts: Vec::new(),
         network_allowlist,
         skip_upstream_tls_verify: true,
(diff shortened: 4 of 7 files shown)


Chisanan232:
## Claude Code review — [AAASM-4126] proxy DLP on all MitM'd hosts

**Verdict: approve-ready (comment-only, no approval attached).**

### 1. CI
All 24 checks green. Coverage/Sonar skipped — noted, not blocking.

### 2. Scope vs ticket
Full coverage of AAASM-4126. Both halves delivered: (a) the MitM+DLP host set is now operator-configurable via `AA_PROXY_MITM_HOSTS` (egress-allowlist grammar, fail-closed on empty), and (b) `intercept_request` credential-DLP now runs on **every** MitM'd non-LLM body, no longer gated on the 3-host `detect_api` allowlist. Regression test `credential_dlp_blocks_secret_to_non_llm_mitm_host` exercises a secret to `generativelanguage.googleapis.com` under `llm_only=true` + a `mitm_hosts` entry and asserts 403 with upstream never dialed.

### 3. Side effects
- **3 built-in LLM hosts unchanged:** `should_mitm` returns true via `detect_api` for openai/anthropic/cohere exactly as before; their handling path is untouched.
- **Passthrough preserved:** under the default `llm_only=true` + empty `mitm_hosts`, non-LLM hosts still `should_mitm()==false` → transparent tunnel, no new body-read/scan latency. DLP only newly runs on hosts that are actually MitM'd.
- **`handle_non_llm_with_gateway` → `handle_non_llm_mitm` refactor preserves gateway routing:** `gateway` is now `Option`; MCP `tools/call` eval (Allow/Deny/Redact + fail-closed on RPC error unless `mcp_fail_open`) fires only when a gateway is configured — identical to before. With no gateway, `mcp_call` stays `None` and the body still flows through the DLP gate (closing the old no-gateway raw-copy bypass).
- **Local:** `cargo nextest run -p aa-proxy` → **200 passed, 0 failed** (incl. all existing LLM-host interception + credential-heist tests and the new DLP test).

### 4. Frontend
N/A — proxy-internal Rust only.

### 5. Cross-PR note (re #1418 / AAASM-4127)
No conflict. Despite the sweep grouping, **#1418 does not touch `aa-proxy/src/config.rs`** — it only edits `aa-cli/start.rs` + `aa-runtime/runtime.rs` and relies on the pre-existing `gateway_endpoint` / `AA_PROXY_GATEWAY_ENDPOINT` reader. This PR's only `ProxyConfig` change (`mitm_hosts` + literal constructors across aa-proxy + aa-integration-tests) is disjoint from #1418's file set, so the two merge in **any order** with no rebase. They are complementary: #1418 exporting `AA_PROXY_LLM_ONLY=false` makes every host MitM'd, which routes through this PR's DLP gate.

— Claude Code


[AAASM-4126]: https://lightning-dust-mite.atlassian.net/browse/AAASM-4126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1422, opened 2026-07-05
[AAASM-4133] 🔒 (gateway): Defense-in-depth hardening batch (agent_id, CA, tenancy, edge authz, gRPC size)

Chisanan232:
## What changed

Defense-in-depth hardening batch for **AAASM-4133** (sweep AAASM-4120). Each finding is LOW severity — none independently exploitable given the token-as-trust-anchor + runtime-as-authoritative model — but each narrows blast radius. Implemented as independent granular commits.

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| 1 | Client-supplied `agent_id` drove agent-scoped matching + suspend targeting (`policy_service.rs`) | New `authoritative_agent_key()` binds the enforcement-override lookup and budget-suspend target to the credential-token owner, falling back to the client id only for the tokenless unregistered OSS flow | `🔒 Bind agent-scoped controls to token-derived id` + `✅` test |
| 2 | MitM root CA `BasicConstraints::Unconstrained`, 10-yr system trust root (`aa-proxy/tls/ca.rs`) | Extracted `build_ca_params()` and added X.509 `NameConstraints` **excluding** the internal/private namespaces the egress proxy never intercepts (loopback, `.local`, `.internal`, `home.arpa`, RFC 1918 / loopback / link-local / IPv6 ULA). Excluded-only ⇒ interception of arbitrary public domains is unchanged | `🔒 Constrain MitM root CA with X.509 NameConstraints` (+ unit test) |
| 3 | Multi-tenant guard defaults to `TenancyMode::Untenanted` | **No code change — already addressed at HEAD.** AAASM-4032 already made this explicit config via `AA_GATEWAY_TENANCY_MODE` with a documented default. See "Deferred" below | — |
| 4 | `report_edge` authorized source but not target (`aa-api/edges.rs`) | Added `authorize_edge_target()`: rejects an edge whose target is another team's agent (admin bypass; team-less target stays allowed as it is not "another team's agent") | `🔒 Authorize edge target team in report_edge` + `✅` test |
| 5 | No explicit gRPC `max_decoding_message_size` (relied on tonic 4 MiB default) | Pinned the ceiling explicitly on every gRPC service in both serve paths (TCP + UDS) via a shared `MAX_DECODING_MESSAGE_SIZE` const; value matches the current default so traffic is unaffected | `🔒 Set explicit gRPC max_decoding_message_size` |

## Why

Hardening sweep — bound the blast radius of a leaked credential/CA key and close permissive fallbacks, without changing any legitimate flow.

## Item #3 — deferred (safe minimal)

Finding #3 offered two options: flip the fail-safe default to `tenanted`, **or** make it explicit config with a clear default. The latter is **already done at HEAD** (AAASM-4032: `AA_GATEWAY_TENANCY_MODE`, `TenancyMode::from_env`). Flipping the *global* default to `tenanted` is **not** minimal here: the `lifecycle_service` registration guard rejects team-less agents in `Tenanted` mode, so a default flip would break zero-config OSS agent registration. Per the ticket's own guidance (implement the safe minimal version, note the follow-up), #3 is left as explicit-config; a deployment-posture default flip is an ops/design decision tracked separately.

## How to verify

- `cargo nextest run -p aa-gateway -p aa-api` — 2216 passed
- `cargo nextest run -p aa-proxy` — 201 passed (incl. new `ca_cert_carries_name_constraints_extension`)
- `cargo fmt --all -- --check`, `cargo clippy -p aa-gateway -p aa-api -p aa-proxy --all-targets -- -D warnings` — clean
- New tests: `agent_id_bound_to_token_test` (spoof cannot dodge enforcement override), `report_edge_rejects_target_in_another_team` (+ same-team / team-less positives), CA name-constraints extension present.

## Notes

- Sibling **AAASM-4125** also edits `policy_service.rs` but in the LLM-metering region (~line 1257); this PR edits the agent_id region (~328/454) + other files. Whichever of the two merges second may need a trivial rebase.
- `observe_mode_test::st_sandbox_4` shared one credential token across two agents — unrealistic under 1:1 token→owner binding; each agent now carries its own token. Assertions unchanged.

Closes AAASM-4133

### DIFF
--- a/aa-api/src/routes/edges.rs
+++ b/aa-api/src/routes/edges.rs
@@ -59,6 +59,28 @@ fn authorize_agent_team_access(
     Ok(())
 }
 
+/// Reject an edge whose **target** is another team's agent (AAASM-4133).
+///
+/// `report_edge` already authorizes ownership of the *source* (reporting) agent
+/// (AAASM-3790). Without this check a caller could still insert an edge whose
+/// target is an agent owned by a different team, polluting that team's
+/// inbound-topology view. An admin may target any agent; a team-scoped caller
+/// may target only agents in a team it can access. A **team-less** target is
+/// not "another team's agent", so it stays allowed — SDK emitters legitimately
+/// record edges to as-yet-unregistered or shared/team-less peers.
+fn authorize_edge_target(caller: &AuthenticatedCaller, state: &AppState, target: AgentId) -> Result<(), ProblemDetail> {
+    if caller.scopes.contains(&Scope::Admin) {
+        return Ok(());
+    }
+    if let Some(team) = agent_team_id(state, target) {
+        if !caller.can_access_team(&team) {
+            return Err(ProblemDetail::from_status(StatusCode::FORBIDDEN)
+                .with_detail("Edge target agent belongs to another team".to_string()));
+        }
+    }
+    Ok(())
+}
+
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
@@ -213,6 +235,9 @@ pub async fn report_edge(
     // AAASM-3790: write-scope + ownership of the source (reporting) agent so a
     // caller cannot poison another team's topology with fabricated edges.
     authorize_agent_team_access(&caller, &state, source)?;
+    // AAASM-4133: also authorize the target so a caller cannot pollute another
+    // team's inbound topology with an edge pointing at that team's agent.
+    authorize_edge_target(&caller, &state, target)?;
 
     let metadata = if let Some(json_str) = body.metadata_json {
         if json_str.is_empty() {
--- a/aa-api/tests/edges.rs
+++ b/aa-api/tests/edges.rs
@@ -686,3 +686,103 @@ async fn report_edge_with_read_only_scope_is_forbidden() {
 
     assert_eq!(response.status(), StatusCode::FORBIDDEN);
 }
+
+// ── AAASM-4133 — report_edge authorizes the edge TARGET, not just the source ──
+//
+// A write-scoped, team-scoped caller may own the source (reporting) agent yet
+// still try to point the edge at another team's agent, polluting that team's
+// inbound-topology view. The target must be authorized too.
+
+use aa_api::auth::config::AuthMode;
+
+async fn post_edge_auth(app: axum::Router, token: &str, body: Value) -> StatusCode {
+    app.oneshot(
+        Request::builder()
+            .method("POST")
+            .uri("/api/v1/topology/edges")
+            .header("authorization", format!("Bearer {token}"))
+            .header("content-type", "application/json")
+            .body(Body::from(body.to_string()))
+            .unwrap(),
+    )
+    .await
+    .unwrap()
+    .status()
+}
+
+#[tokio::test]
+async fn report_edge_rejects_target_in_another_team() {
+    let state = common::test_state_with_auth(AuthMode::On, &[], 1000);
+    // Source belongs to the caller's team (alpha); target belongs to beta.
+    state.agent_registry.register(team_agent(0x01, "alpha")).unwrap();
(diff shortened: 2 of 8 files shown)


Chisanan232:
## Claude Code review — APPROVE-READY after CI fix (comment only; Pioneer approval required)

Reviewed AAASM-4133 across CI, scope, side-effects, and FE.

- **CI:** The `Test` / `Integration tests` jobs were a **real failure** — `st_r_5_per_agent_override_isolates_observe_from_enforce_under_one_policy` in `aa-integration-tests/tests/observe_mode_e2e.rs`. Root cause: that test's helpers shared one `credential_token` ("tok") across two agents, which is unrealistic under the new 1:1 token→owner binding — the shared token resolved both agents to one owner, so the experimental agent's Observe override no longer applied. This is the **same shared-token pattern you adapted in the sibling `observe_mode_test.rs::st_sandbox_4`** but the integration-test sibling was missed. Fixed on-branch by the identical adaptation (explicit per-agent tokens; distinct tokens for the two-agent isolation case; assertions unchanged) — pushed as `✅ (aa-integration-tests): Give each agent its own token in observe e2e`. No suppression, no weakened assertions. Remaining non-green jobs are path-filtered `skipping` + coverage/Sonar (note only).
- **Scope vs ticket:** 4 of 5 items land with granular commits + tests — (1) agent_id→token-owner binding, (2) MitM CA NameConstraints, (4) edge-target authz, (5) explicit gRPC max size. Item **3 (Untenanted default) legitimately deferred** — the ticket itself frames it as "set `tenanted` in SaaS **or** make it the fail-safe default", and the explicit-config half is already satisfied by AAASM-4032 (`AA_GATEWAY_TENANCY_MODE`). Flipping the global default to `tenanted` would break zero-config OSS registration (the lifecycle guard rejects team-less agents in Tenanted mode), so it is correctly left as an ops/deployment-posture follow-up, not a code gap. **Deferral is reasonable; not a blocker.**
- **Side-effects (security-critical paths):** ✔ (a) `authoritative_agent_key`: empty token → falls back to `req.agent_id`, so the tokenless unregistered OSS flow is unchanged and registered-agent-with-token enforcement binds to the owner. ✔ (b) CA constraints are **excluded-only** (`permitted_subtrees: vec![]`), covering loopback/RFC1918/link-local/`.local`/`.internal`/`home.arpa`/IPv6 ULA — public-domain interception unchanged. ✔ (c) `authorize_edge_target` bypasses `Scope::Admin` and only rejects a target owned by *another* team; team-less and same-team targets stay allowed. ✔ (d) `MAX_DECODING_MESSAGE_SIZE = 4 * 1024 * 1024` equals tonic's existing 4 MiB default (no behavior change), applied via `InterceptedService` on both TCP + UDS paths preserving the existing interceptors. ✔ `st_sandbox_4` assertions unchanged (still asserts per-agent Observe/Enforce isolation).
- **FE:** N/A — Rust-only (aa-gateway / aa-api / aa-proxy).
- **Cross-PR:** Sibling AAASM-4125 (its PR) also edits `policy_service.rs` but in the LLM-metering region (~L1257) vs this PR's agent_id region (~L328/474). Whichever of the two merges second may need a trivial rebase — no semantic conflict expected.

**Local validation (macOS, RUSTUP_TOOLCHAIN=stable, post-fix):** `cargo nextest run -p aa-gateway -p aa-api` → **2223 passed, 0 skipped**; `cargo nextest run -p aa-integration-tests st_r_5` → **1 passed** (the fixed test); `cargo build -p aa-proxy` clean; `cargo clippy -p aa-gateway -p aa-api -p aa-proxy --all-targets -D warnings` clean; `cargo fmt --all --check` clean. (aa-proxy's full `st_*` suite + NATS/Postgres testcontainer integration binaries not re-run locally — Docker; build + clippy cover the `ca.rs` change.)

— Claude Code

PR #1423, opened 2026-07-05
[AAASM-4125] 🐛 (budget): Meter LLM calls even when client declares prompt_tokens=0

Chisanan232:
## Description

The pre-execution LLM budget accrual in the gateway (`maybe_accrue_llm_spend`) priced a call from the **client-supplied** `prompt_tokens` count. A declared `0` (or a negative / omitted value, which the proto deserializes to `0`) made `llm_call_cost_usd` return `$0.00`, which tripped the `cost <= 0.0` early-return so `check_and_accrue_llm_spend` was never called. An agent that always sent `LlmCall{ prompt_tokens: 0 }` therefore incurred **unmetered LLM spend indefinitely**, even for a known priced model — the daily/monthly cap never engaged.

This floors a non-positive client-declared prompt-token count to a conservative server-side minimum (`MIN_ESTIMATED_PROMPT_TOKENS = 1_000`) before pricing, so the call always prices non-zero and the budget reservation runs. Same fail-closed posture as the AAASM-4069 unknown-*model* fix (which the AAASM-4120 sweep flagged as distinct from this token-count trust gap). Legitimate calls that report a real positive count are priced on that count and are unaffected, so no legitimate call is double-charged.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Behaviour change is confined to calls that declared a non-positive prompt-token count — those previously bypassed metering entirely and now accrue a conservative floored cost.

## Related Issues

- Related Jira ticket: AAASM-4125
- Related GitHub issues: N/A

## Testing

Added `zero_prompt_tokens_still_accrues_spend_and_engages_cap` in `aa-gateway/tests/budget_accrual_test.rs`: two `check_action` calls for a known priced model (gpt-4o) with `prompt_tokens: 0` against a tiny daily budget — the first accrues the floored cost, the second is denied once accrued spend exceeds the cap. Fails against the pre-fix code (both calls stayed allowed).

Validation (local, aa-gateway scoped):
- `cargo nextest run -p aa-gateway` — green
- `cargo clippy -p aa-gateway --all-targets -- -D warnings` — clean

## Checklist

- [x] All CI checks passing

---

Note: sibling ticket AAASM-4133 also edits `aa-gateway/src/service/policy_service.rs` but in a different region (client `agent_id`, ~lines 328/454); this change is confined to the LLM-metering region (~line 1257). Whichever of the two PRs merges second may need a trivial rebase.

### DIFF
--- a/aa-gateway/src/service/policy_service.rs
+++ b/aa-gateway/src/service/policy_service.rs
@@ -1219,6 +1219,20 @@ impl PolicyServiceImpl {
         }
     }
 
+    /// AAASM-4125 — conservative server-side floor for a *client-declared* prompt
+    /// token count that is non-positive (`0`, negative, or an omitted proto field
+    /// which deserializes to `0`). A real LLM call always consumes some prompt
+    /// tokens; a declared `0` is therefore either an omission or a deliberate
+    /// under-report and must NOT be trusted as a genuinely empty call. Pricing it
+    /// at `$0` trips the `cost <= 0.0` accrual short-circuit and skips the budget
+    /// reservation entirely, giving an agent that always sends `prompt_tokens: 0`
+    /// unmetered LLM spend indefinitely. Flooring to a minimum estimate keeps the
+    /// call priced non-zero so the daily / monthly cap engages (fail-closed,
+    /// mirroring the unknown-model handling in AAASM-4069). Legitimate calls that
+    /// report a real positive count are priced on that count and are unaffected,
+    /// so no legitimate call is double-charged.
+    const MIN_ESTIMATED_PROMPT_TOKENS: u64 = 1_000;
+
     /// AAASM-3353 / AAASM-3986 — atomically reserve the cost of a non-denied LLM
     /// call against the agent's budget so daily / monthly limits actually fire,
     /// with no check→record TOCTOU.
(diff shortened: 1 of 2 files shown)


Chisanan232:
**Review — PR #1423 (AAASM-4125): meter LLM calls even when client declares `prompt_tokens=0`**

**Verdict: APPROVE-READY** (comment only — leaving the formal Pioneer-team approval to a human reviewer). Clean, correctly-scoped fail-closed fix with a genuine regression test. No blocking issues.

- **CI:** ✅ Green. All Rust-relevant jobs pass — Test, Build, Clippy lint, Analyze (rust), Rust conformance, Rustdoc, Integration tests (macos), TimescaleDB, Migration drift, CodeQL. 25/39 checks pass, **0 fail**; the 14 `skipping` are `ci.yml` path-filtered non-Rust jobs (dashboard/openapi/proto/ebpf) that don't apply to a `aa-gateway` budget change. No billing abort.
- **Scope vs ticket:** ✅ Matches AAASM-4125 exactly — floors a non-positive *client-declared* `prompt_tokens` to a server-side minimum before pricing so the `cost <= 0.0` accrual short-circuit can no longer be tripped by a declared `0`/negative/omitted count. Correctly kept distinct from the AAASM-4069 unknown-*model* fix, and a regression test is included.
- **Side effects:** ✅ No over-charge risk to legitimate traffic. The floor is reached *only* for a non-positive declared count (`match u64::try_from(..) { Ok(t) if t > 0 => t, _ => MIN }`); a real positive count is used verbatim. The AAASM-4069 unknown-model path is untouched and its test stays green. Full `-p aa-gateway budget` suite = **163/163 pass** locally.
- **FE:** N/A — gateway-internal metering; no dashboard/API surface touched.
- **Floor-value judgment (`MIN_ESTIMATED_PROMPT_TOKENS = 1000`):** ✅ Sound. At the gpt-4o input rate ($0.005/1k) that's **$0.005** per floored call ($0.015 at the unknown-model fallback rate) — a negligible absolute reservation. Crucially it *cannot* over-bill a legitimate small/empty-prompt call, because a legitimate caller reports its real positive count (used as-is); the floor only ever applies to a declared `0`/negative/omitted value, which is exactly the untrusted input. Since a real LLM call's system prompt alone typically exceeds 1000 tokens, the floor tends to *under*-estimate real spend, so it's conservative in the safe direction while still non-zero enough to engage the cap. A fixed floor (vs. a "smarter" estimate) is the right call here: the pre-execution reservation is inherently approximate and there is no post-execution `record_usage` reconciliation yet (test-only callers), so a conservative constant is the correct fail-closed minimum.
- **Regression test verified:** ✅ Reverted just the source fix (kept the test) — `zero_prompt_tokens_still_accrues_spend_and_engages_cap` FAILS against pre-fix code (2nd call stays `Allow` instead of `Deny`; `left: 1, right: 2`) and passes with the fix. Author's claim confirmed.
- **Cross-PR:** ⚠️ Sibling PR #1422 (AAASM-4133) also edits `policy_service.rs` but in a different region (client `agent_id`, ~L328/454) vs. this metering region (~L1257). No semantic conflict; whichever merges second may need a trivial line-context rebase.

Local validation (worktree, `RUSTUP_TOOLCHAIN=stable`): `cargo nextest run -p aa-gateway budget` → **163 passed / 0 failed**; `cargo clippy -p aa-gateway --all-targets -- -D warnings` → **clean (exit 0)**; `cargo fmt --all -- --check` → **clean**.

— Claude Code

PR #1547, opened 2026-07-17
[AAASM-4744] 🔒 (core): Close low-severity defense-in-depth gaps

Chisanan232:
## Description

Closes the seven independent LOW-severity defense-in-depth gaps bundled under
AAASM-4744. Each item is its own atomic commit (plus a test commit where
testable). No behaviour change beyond the specific hardening described per item.

Jira: https://lightning-dust-mite.atlassian.net/browse/AAASM-4744

### Sub-items (all fixed)

1. **`budget/pricing.rs` — untabled model no longer prices to `$0`.** `cost_usd`'s
   `None` arm now routes through the fail-closed `fallback_cost_usd` (costliest-
   known rate), so a partial custom pricing table can't silently zero-price an
   omitted model and bypass the budget cap. Also clamps a non-positive computed
   cost in `record_usage` (mirrors the existing `record_raw_spend` clamp) so a
   negative custom rate can't decrement accrued spend. **Fixed.**
2. **`remote_mode/server.rs` + `auth.rs` — admin status withheld off-loopback in
   bypass mode.** `GET /api/v1/admin/status` discloses backend detail and, in the
   `AuthMode::Off` zero-config default, answers with no credential. `listen_addr`
   is now threaded into `router()`; in bypass mode the route is mounted only on a
   loopback bind (preserving the zero-config `aasm status` contract) and withheld
   off-loopback with a warning pointing at the auth opt-in. **Fixed.**
3. **`aa-runtime/pipeline/enforcement.rs` — label values are now scanned.**
   `enforce()` scanned the event detail but left agent-supplied `labels` values
   unscanned; they now pass through the same scan/redact path, so a secret in a
   label value is redacted. **Fixed.**
4. **`aa-api/bin/aa-api-server.rs` — generated admin key no longer printed in
   full.** The self-generated key was `eprintln`'d whole; now only a 6-char
   prefix is printed (matching the from-env banner). Operators set `AASM_API_KEY`
   for a known, reusable key. **Fixed.**
5. **connectors (webhook/slack/pagerduty/opsgenie) — request URL stripped from
   surfaced transport errors.** `ConnectorError::Transport(e.to_string())` could
   embed a caller-controlled destination URL (with a query token); now uses
   `reqwest::Error::without_url()` before `to_string()`. **Fixed.**
6. **`aa-ebpf/src/loader.rs` — unaligned read fixed.** The perf-buffer path cast a
   possibly-unaligned `BytesMut` pointer to `*const FileIoEventRaw` and formed a
   reference to it (UB). Replaced with `ptr::read_unaligned` into an owned,
   aligned value (the technique `ringbuf::bytes_to` already uses). File-disjoint
   from AAASM-4739. **Fixed.**
7. **`aa-api/src/server.rs` — real `/livez` and `/readyz` handlers.** Both paths
   fell through to the SPA catch-all and returned `200 index.html`; they now alias
   the health handler, so a probe gets the real health JSON (and `503` on
   degradation) instead of the SPA shell. **Fixed.**

## Type of Change

- [x] 🐛 Bug fix
- [x] 🔧 Configuration / CI change _(security hardening; no config change)_

## Breaking Changes

- [x] No

`remote_mode::router()` gained a `listen_addr` parameter (item 2) — an internal
API used by two in-repo e2e tests, both updated in the same commit. No public
wire/behaviour change on loopback (the zero-config `aasm status` contract is
preserved).

## Related Issues

- Related Jira ticket: AAASM-4744
- Fixes AAASM-4744

## Testing

Verified with `cargo check` / `cargo clippy --all-features -- -D warnings` (0
errors) on the impacted crates and targeted `cargo nextest` runs for the new
tests (all green):

- `budget::pricing::tests::cost_usd_unknown_pair_prices_via_fail_closed_fallback`
- `budget::pricing::tests::partial_custom_table_prices_omitted_model_via_fallback_not_zero`
- `budget::tracker::tests::record_usage_clamps_negative_custom_rate_to_zero`
- `remote_mode::server::admin_status_bind_guard_tests::{bypass_on_loopback_exposes_admin_status, bypass_off_loopback_withholds_admin_status}`
- `pipeline::enforcement::tests::secret_in_label_value_is_redacted`
- `aa-api::serve_local_spa::livez_and_readyz_serve_json_not_spa_html`

Item 6 (eBPF) is Linux-only; type-checked against `x86_64-unknown-linux-gnu`
(the only remaining errors are the pre-existing missing-BPF-artifact
`include_bytes_aligned!` reads, unrelated to this change). Item 4 is a bin
`main` startup print with no unit-test surface. The existing `admin_status_e2e`,
`remote_mode_e2e`, and `remote_mode_http` suites still pass after the `router()`
signature change.

## Checklist

- [x] Documentation updated if behaviour changed _(why-comments added at each site)_
- [ ] All CI checks passing _(org CI is billing-blocked intermittently; validated locally)_

https://claude.ai/code/session_01HMnPRm9T3fdrS4uNYXkzg6

### DIFF
--- a/aa-api/src/bin/aa-api-server.rs
+++ b/aa-api/src/bin/aa-api-server.rs
@@ -70,7 +70,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
         LocalAuth::ApiKey { key } => {
             if generated {
-                eprintln!("aa-api: generated admin API key (set AASM_API_KEY to reuse): {key}");
+                // AAASM-4744: print only a short prefix of the generated key, not
+                // the full secret. Dumping a live admin credential to stderr/logs
+                // is a disclosure gap; the prefix lets an operator correlate the
+                // key in the banner below without leaking it. Set AASM_API_KEY to
+                // run with a known, reusable key.
+                eprintln!(
+                    "aa-api: generated a random admin API key {prefix}… (set AASM_API_KEY to use a known key)",
+                    prefix = &key[..key.len().min(6)]
+                );
             } else {
                 eprintln!("aa-api: using admin API key from AASM_API_KEY");
             }
--- a/aa-api/src/destinations/connectors/opsgenie.rs
+++ b/aa-api/src/destinations/connectors/opsgenie.rs
@@ -59,7 +59,11 @@ impl NotificationConnector for OpsGenieConnector {
             .json(&body)
             .send()
             .await
-            .map_err(|e| ConnectorError::Transport(e.to_string()))?;
+            // AAASM-4744: strip the request URL from the surfaced error. A
+            // reqwest error's Display embeds the URL; `without_url` keeps it out
+            // of the 502 body and logs, consistent with the caller-controlled
+            // connectors.
+            .map_err(|e| ConnectorError::Transport(e.without_url().to_string()))?;
         let status = resp.status().as_u16();
         let resp_body = resp.text().await.unwrap_or_default();
         let resp_body = truncate_body(resp_body);
--- a/aa-api/src/destinations/connectors/pagerduty.rs
+++ b/aa-api/src/destinations/connectors/pagerduty.rs
@@ -60,7 +60,11 @@ impl NotificationConnector for PagerDutyConnector {
             .json(&body)
             .send()
             .await
-            .map_err(|e| ConnectorError::Transport(e.to_string()))?;
+            // AAASM-4744: strip the request URL from the surfaced error. A
+            // reqwest error's Display embeds the URL; `without_url` keeps it out
+            // of the 502 body and logs, consistent with the caller-controlled
+            // connectors.
+            .map_err(|e| ConnectorError::Transport(e.without_url().to_string()))?;
         let status = resp.status().as_u16();
         let resp_body = resp.text().await.unwrap_or_default();
         let resp_body = truncate_body(resp_body);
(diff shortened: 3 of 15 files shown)


Chisanan232:
## Claude Code — automated review

**CI:** All green (57 checks pass, 0 fail) after resolving one regression.

**Scope vs AAASM-4744:** Complete — all 7 defense-in-depth items landed (budget fallback + clamp, remote-mode admin fail-closed off-loopback, label-value secret scan, admin-key prefix-only print, connector URL stripping, eBPF loader unaligned read, real /livez + /readyz handlers).

**Regression found & fixed (side-effect review):** the existing integration test `runtime_bypass::preflight_equivalence_yields_identical_authoritative_outcome` initially failed. Root cause: item-3 label-value scanning correctly scans a forged preflight label value (+5 bytes on the `scanned_bytes` *work counter*); the **authoritative decision** (findings, redaction, is_clean) was unchanged. The test had over-asserted the entire outcome struct incl. the pure work counter. Fixed by re-expressing the test to assert authoritative parity AND that scan-work can only increase (the fail-safe property) — the fix itself was NOT weakened. All 19 secret-interception tests + the item-3 unit test pass.

**Side effects:** None outstanding — the one existing-test interaction is resolved; `router()` signature change updated all callers (e2e tests green); clippy clean.

**Note:** admin-key behavior change — with auth-on and no `AASM_API_KEY`, the generated key is no longer printed in full (only a prefix), so operators must set `AASM_API_KEY` for a stable key. Intended defense-in-depth tradeoff, called out in the PR body.

**Verdict:** ✅ Ready for approval & merge.

PR #1573, opened 2026-07-19
[AAASM-4862] 🔒 (aa-ebpf): Emit signal when PID_FILTER map is full

Chisanan232:
## Description

`aa-syscall-guard`'s fork tracepoint copies `PID_FILTER` membership from a confined parent to each newly forked child so the child is confined from its first syscall (AAASM-3916). When `PID_FILTER` is at its 1024-entry cap the membership `insert` fails, and the previous code discarded that failure (`let _ = PID_FILTER.insert(...)`) — the overflow child then ran with **zero syscall confinement and no signal at all**.

This PR makes that degradation **observable**:

- The fork tracepoint now increments a new per-CPU counter map, `FORK_MAP_FULL_DROPS`, whenever the `PID_FILTER` insert fails.
- Userspace reads it via a new accessor, `SyscallGuardLoader::fork_map_full_drops()`, which sums the per-CPU slots — a non-zero total means that many forked children escaped confinement because the map was full.

A **per-CPU counter** (not a perf/ring event) is deliberate: a fork burst against a full map would otherwise emit an event storm at the worst possible moment; a monotonic counter conveys the same "N children escaped confinement" signal without amplifying the pressure. This matches the codebase's existing eBPF observability channels (perf array / ring buffer) as a lighter-weight metric variant appropriate for a degradation counter.

### Scope: why this is the observability half only

The ticket also raises a "fail-closed — kill the unconfinable child" alternative. **That is not achievable at the `sched_process_fork` tracepoint** and this PR does not attempt it:

- The tracepoint **cannot block the fork** — it is an observe-point, not a gate.
- A child that never receives a `PID_FILTER` entry also **escapes the `raw_syscalls/sys_enter` enforcement probe** — that probe's `PID_FILTER` lookup is the very thing that gates the SIGKILL, so there is no confined PID to signal.

So the only actionable half is emitting a signal on map-full, which is exactly AAASM-4862's stated acceptance criterion. This PR delivers that.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-4862

## Testing

- [x] Unit tests added / updated — `fork_map_full_drops_returns_error_on_non_linux` pins the accessor's non-Linux contract; full `aa-ebpf` suite green (106 passed).
- Notes: `cargo check -p aa-ebpf`, `cargo clippy -p aa-ebpf --all-targets --all-features -- -D warnings`, and `cargo nextest run -p aa-ebpf` all pass on macOS. The BPF probe object (`aa-ebpf-probes`) targets `bpfel-unknown-none` and is not compiled on macOS (per repo policy `cargo check -p aa-ebpf` is sufficient there); the aya-ebpf `PerCpuArray` API used was verified against the pinned crate source.

## Checklist

- [ ] All CI checks passing

Closes AAASM-4862

### DIFF
--- a/aa-ebpf-probes/src/syscall_guard.rs
+++ b/aa-ebpf-probes/src/syscall_guard.rs
@@ -83,7 +83,7 @@ use aa_ebpf_common::syscall::MAX_SYSCALL_ALLOWLIST;
 use aya_ebpf::{
     helpers::{bpf_get_current_pid_tgid, bpf_send_signal},
     macros::{map, tracepoint},
-    maps::HashMap,
+    maps::{HashMap, PerCpuArray},
     programs::TracePointContext,
 };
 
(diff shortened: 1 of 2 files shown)


Chisanan232:
## Review — Claude Code

**CI:** ✅ green (all checks SUCCESS + expected SKIPPED; no failures).
**SonarCloud:** 0 open issues
**Scope vs AAASM-4862:** ✅ complete — new per-CPU `FORK_MAP_FULL_DROPS` counter + `fork_map_full_drops()` reader make the eBPF map-full fail-open observable
**Side-effects:** additive (new map + reader; no existing path changed); `cargo check/clippy/nextest -p aa-ebpf` 106 pass
**Verdict:** ✅ ready to approve + merge. Scope note: the fail-closed kill-child is **not implementable** at the fork tracepoint (documented in the PR body); this delivers the observability half only.

PR #1699, opened 2026-07-25
[AAASM-5090] ✨ (aa-api): Project capability matrix from registry and policy cascade

Chisanan232:
## Description

`GET /api/v1/capability/matrix` already existed, but it served a fixture compiled
into `aa-api` that mirrored the dashboard's mock byte for byte — the endpoint's
own module doc said so and deferred "a live projection from the policy engine" to
a follow-up story. This is that follow-up: the endpoint now projects real state,
and `createMockCapabilityClient` is deleted from the dashboard.

**No new path — the count stays 72.** This converts the existing
`/api/v1/capability/matrix` endpoint rather than adding a parallel one, so
`aa-integration-tests/tests/api_openapi_contract.rs` needed no change to either
its path count or its sorted path list.

**Read-only projection — no enforcement or hot-path change.** Verified with
`git diff --name-only remote/main..HEAD`: **zero** files under `aa-runtime/`,
`aa-gateway/`, `aa-proxy/`, `aa-ebpf*`, `aa-core/` or `aa-sandbox/`. The diff is
confined to `aa-api/`, `openapi/`, `dashboard/` and their tests. The projection
only *reads* `aa-gateway` / `aa-core` public APIs and cannot alter a verdict. The
capability override store was already display-only (read by these four handlers
and nothing else); it becomes an overlay replayed over the projection instead of
a mutable copy of the fixture, which is what allowed the seeded matrix to be
deleted.

### Where each field comes from

| Field | Real source |
|---|---|
| `agents[].id` / `name` / `framework` / `status` / `lastSeen` | `AgentRegistry` → `AgentRecord` (`lastSeen` is the raw heartbeat timestamp; the relative phrase is rendered client-side) |
| `agents[].owner` | registry `team_id`, falling back to `org_id` |
| `agents[].mode` | registry `enforcement_mode` override (`Observe` → `shadow`) |
| `agents[].caps` | merged policy capability cascade, resolved with the same public `aa_core` helpers the enforcement guard uses (`capability_is_denied`, `CapabilitySet::allow_is_restricted`) |
| `resources[]` | the three fixed capability families, plus tools declared at registration / named by an `mcp_tool:` grant / named by a per-tool policy |
| `policies[]` | the policy documents in each agent's live cascade (`name`, `policy_version`, `scope`, capability + tool declarations); `affects` is computed from which agents' cascades contain the document |

The four matrix verbs are the four verb-shaped `Capability` variants that
`aa_core::action_to_capability` already defines, so no new verb-mapping rule is
introduced here.

### Fields deliberately left absent (never faked)

Each is omitted from the wire (`skip_serializing_if`) and folds to `—` in the UI,
so a consumer can tell "unmeasured" from a real zero:

- `agents[].trust` — nothing in the gateway computes a trust score. Deriving one
  is a scoring rule owned by **AAASM-5083**.
- `agents[].flagged`, `agents[].note`, `caps[].flag` — no source; `flagged` is the
  same scoring question as `trust`.
- `agents[].mode` when the agent declared no override, or declared `Disabled`
  (unrepresentable in an enforce/shadow view).
- `policies[].hits24h` — attributing audit events back to a document is owned by
  **AAASM-5096**.
- `resources[].group` on tool columns — an MCP tool name is operator-supplied and
  classified nowhere. System families keep a group because the `Capability` enum
  itself names their domain.
- `resources[].paths` — always empty: the capability model grants a whole family
  or tool and carries no per-path sub-scopes.
- `sampleCalls[]` — **whole dimension empty.** Call samples need a
  proposed-vs-current policy diff; that is the policy-replay surface owned by
  **AAASM-5094** (recorded as deferred in ADR 0017).
- `Decision::narrow` / `Decision::approval` **never appear.** Both come from other
  policy stages (credential scrubbing; a tool's `requires_approval_if` CEL
  condition evaluated against a concrete action) and cannot be read off a static
  capability set. Cells carry `allow` / `deny` / `na` only.

ADR 0017 was checked first: it audits nine surfaces and **Capability is not one of
them**, so nothing on this page is a ratified evolution and nothing was reverted
toward the mock.

## Type of Change

- [x] ✨ New feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [ ] No
- [x] Yes (describe below)

`CapabilityMatrix` fields with no real source become optional and are omitted
rather than sent with placeholders: `CapabilityAgent.owner` / `trust` / `mode`,
`Resource.group`, `Policy.version` / `hits24h`. `CapabilityAgent.lastSeen` changes
from a pre-rendered phrase (`"2m ago"`) to an ISO 8601 timestamp. The dashboard is
updated in the same PR, and `openapi/v1.yaml` plus the generated dashboard types
are regenerated. **The OpenAPI path count is unchanged at 72** — this adapts the
existing `/api/v1/capability/matrix` endpoint rather than adding a parallel one.

## Related Issues

- Related Jira ticket: [AAASM-5090](https://lightning-dust-mite.atlassian.net/browse/AAASM-5090)
  (parent Epic [AAASM-5082](https://lightning-dust-mite.atlassian.net/browse/AAASM-5082))
- Related GitHub issues: N/A

## Testing

**Rust** — reworked coverage in `aa-api/src/routes/capability.rs` and
`aa-integration-tests/tests/api_capability.rs`: projection of registered agents,
empty registry, absent-field omission on the wire, the heartbeat timestamp, the
`Observe → shadow` / `Disabled → absent` mode mapping, the guard's fail-closed
rules (`decide`, including the `file_write ⇒ file_delete` superset deny and the
empty-but-restricted allow-list), unmodelled verbs staying `na`, non-admin denial
on read, the `team_id` / `tool` / `effective_only` filters, and the override
overlay (apply → visible, revoke → projected value restored, TTL expiry, unknown
agent rejected with nothing logged).

**Dashboard** — `features/capability/api.test.tsx` covers the query hook's success
*and* error branches plus every `createApiCapabilityClient` branch (success,
error envelope, empty body, override success and rejection);
`__tests__/display.test.ts` covers the absent-data fold, including that a real `0`
survives while only absence folds.

**Visual evidence** — `dashboard/verify/5090/`: matrix, per-agent and
per-resource tabs in light and dark, captured by
`tests/e2e/verify-aaasm-5090.spec.ts`
(`pnpm exec playwright test -c tests/e2e/verify-5090.config.ts`). The fixture
deliberately omits trust / mode / owner on one agent so the `—` fold is in frame.

## Review-pass fixes (correctness, on top of the original projection)

A deep review of this branch found the projection was reading every cell off the
merged capability set alone, which is only one of the stages the engine runs.
Fixed here, still with **zero** files touched under `aa-runtime/`, `aa-gateway/`,
`aa-proxy/`, `aa-ebpf*`, `aa-core/` or `aa-sandbox/`:

- **Tool cells were a reporting fail-open.** The engine denies a tool at stage 3
  (`stage_tool_allow`, `aa-gateway/src/engine/decision.rs`) by reading
  `doc.tools[name]` with an exact-entry-then-`"*"`-fallback — a field the
  projection never consulted. With a policy like `policy-examples/strict.yaml`
  (`tools: { "*": { allow: false }, read_file: { allow: true } }`), a tool such as
  `send_email` was enforced as **deny** but rendered as **allow** — while the
  policies pane, which *does* read `doc.tools`, showed a `deny` rule for the same
  tool. The cell now mirrors that stage exactly (a comment names
  `decision.rs:stage_tool_allow` so future drift is greppable) and combines
  most-restrictive-wins with the capability decision.
- **`"*"` became a literal resource column.** The tools wildcard was folded into
  the column set like any other key, so a strict policy grew a column named `*`
  reading `allow` — the inverse of what it declares. It is now filtered out.
- **The network cell ignored `stage_network`.** A policy declaring a `network`
  section with an **empty allowlist is deny-all** (AAASM-3127/AAASM-3730); the
  cell now reports `deny` for that unambiguous case. A present, non-empty
  allowlist is deliberately left as the capability-derived decision: egress *is*
  permitted, just host-restricted, and a single cell has no host to test.
- **An override could inject an unreachable decision.** `decision` was an
  unvalidated `Decision`, so `POST /capability/override {"decision":"narrow"}`
  wrote a value into a cell that the projection can never produce and no revoke
  could restore. `narrow` / `approval` are now rejected with **400**.
- **The dashboard froze after an override.** `optimistic ?? data` was never
  cleared and nothing invalidated the matrix query, so every later refetch
  (window focus, remount, retry) landed in `data` and was silently discarded. The
  page now invalidates `['capability','matrix']` on success and drops the
  optimistic shadow once the refetch resolves.

### Override lifecycle — response behaviour worth knowing

The store is append-only and revocation flips a flag rather than deleting a row,
which produces three responses a reviewer may not expect:

- A **TTL expiry marks the entry `active: false`**, so a subsequent
  `DELETE /capability/override/{id}` for that id returns **404**, not 204 — the id
  is known but no longer active.
- `DELETE` resolves **by override id only**. An override whose agents have all
  disappeared is still an active entry, so deleting it returns **204**.
- An override **survives its agent disappearing**: the entry stays in the log and
  `apply_overlay` silently skips it, because inventing a row for an agent the
  cascade no longer produces would be worse than dropping the annotation.

`OverrideRecord.active`'s doc comment claimed it was "always `true` … (no TTL or
explicit delete support yet)"; both exist, and the doc is corrected here.

### Tests added for the above

Every existing projection test ran against a policy with no `capabilities`, no
`tools` and no `network`, so every asserted cell was `allow` by vacuity. New
coverage (each verified to fail when its fix is reverted):

| Test | Guards |
|---|---|
| `every_stage_that_denies_is_visible_in_the_cells` | capability deny, `"*"` tool fallback deny, exact tool entry beating the wildcard, empty-allowlist egress deny |
| `the_tool_wildcard_never_becomes_a_resource_column` | `"*"` never reaching `resources[]` or `caps` |
| `a_team_scoped_deny_reaches_the_projection` | the explicit `collect_cascade_with_lineage` call — regressing it to `collect_cascade` drops every Org/Team deny |
| `override_rejects_a_decision_the_projection_cannot_express` | the 400 on `narrow` / `approval` |
| `project_rules_flattens_capabilities_and_tools` | rule flattening, which had no coverage at all |
| `CapabilityPage › re-syncs with the server projection after a successful override` | the optimistic-shadow freeze |

**Review validation** — `dashboard/verify/5090/review/`, captured by
`tests/e2e/review-aaasm-5090.spec.ts`
(`pnpm exec playwright test -c tests/e2e/review-5090.config.ts`, port 4591). It
re-derives the claims rather than only photographing them: the grid is genuinely
agents × resources with addressable allow/deny/na cells, every omitted field
(`trust`, `flagged`, `hits24h`, `sampleCalls`, tool `group`) folds to `—` and not
to `0`, the filter bar narrows the grid (search / framework / mode), and the page
logs no console errors — asserted in both light and dark.

## Checklist

- [ ] All CI checks passing

### Note for reviewers — pre-existing bug found while doing this (NOT fixed here)

`AppState::local_in_memory` (`aa-api/src/state.rs:281-284`) is the only `AppState`
builder, and it constructs the engine as
`PolicyEngine::load_from_file(..).with_invalidation_hub(..)` — it never calls
`.with_registry(..)`, which `aa-gateway/src/engine/mod.rs:754` documents as
"Call this after `load_from_file` in the server startup path."

So `PolicyEngine.registry` is permanently `None` in the aa-api process, and per
the field's own doc (`mod.rs:218-219`) `collect_cascade` then "walks only Global
and Agent scopes": it resolves `Lineage::default()` and skips the Org and Team
tiers entirely (`mod.rs:1926-1933`). `effective_permissions` goes through
`collect_cascade` (`mod.rs:1622`), so **`GET /api/v1/agents/{id}/capabilities`**
(`aa-api/src/routes/agents.rs:633`) silently omits every Org- and Team-scoped
allow *and* deny. Because team-scoped denies are dropped, that endpoint can
report a capability as permitted when a team policy denies it — a reporting
fail-open, surfaced in `aasm policy show --show-permissions`,
`aasm topology lineage --show-permissions`, and the dashboard's
inherited-permissions panel.

**Enforcement is not affected.** `PolicyEngine::evaluate` deliberately does not
use `collect_cascade`; it resolves tenancy via `authoritative_lineage(ctx)` and
calls `collect_cascade_with_lineage` (`mod.rs:829`, per AAASM-3729), so real
verdicts still see Org/Team policy.

This PR sidesteps the gap rather than fixing it: the matrix calls
`collect_cascade_with_lineage` with a lineage built from
`state.agent_registry.lineage(..)`, so it does see Org/Team policy. Fixing the
shared engine wiring is deliberately left out of scope and wants its own ticket.

[AAASM-5090]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — review record (AAASM-5090)

**Verdict: READY** ✅ — after one blocking defect was found and fixed on this branch.

### The blocking defect (found in review, now fixed)

The matrix reported **`allow` for tools the gateway actually denies** — a reporting fail-open on the one page whose purpose is "what can this agent actually do".

Tool cells were computed from the merged capability set alone. Enforcement denies a tool at **Stage 3**, `stage_tool_allow` (`aa-gateway/src/engine/decision.rs:187-198`), reading `doc.tools[name].allow` with an exact-entry → `"*"` fallback — a field the projection never consulted. `policy-examples/strict.yaml:41` ships exactly that shape (`tools: { "*": { allow: false } }`), so a shipped example policy reproduced it. The payload was also self-contradictory: `project_rules` *did* read `doc.tools` and emitted a `deny` rule for the same tool the grid showed as `allow`.

CI could not have caught this: every projection test ran against a policy with no `capabilities`, no `tools`, no `network`, so **every asserted cell was `allow` by vacuity**.

| # | Finding | Sev | Fix |
|---|---|---|---|
| F1 | Tool cells ignored `stage_tool_allow` → fail-open | **blocking** | `cascade_denies_tool()` mirrors the stage (exact → `"*"` → `allow:false`), combined most-restrictive-wins (`16302d75`) |
| F2 | `"*"` wildcard rendered as a literal resource column showing `allow` | **blocking** | sentinel filtered from the tool id set (`bd107cd8`) |
| F3 | Network cell ignored `stage_network`; declared-but-empty allowlist is deny-all | should-fix | `cascade_denies_all_egress()` for the unambiguous case; non-empty allowlist stays capability-derived, with a comment (`16302d75`) |
| F4 | The AAASM-5102 sidestep (explicit lineage) had **zero** coverage | should-fix | `a_team_scoped_deny_reaches_the_projection` + deny coverage for capability / `"*"` tool / exact tool / empty-allowlist network (`43de6a54`) |
| F5 | Optimistic override state never cleared, no query invalidation → every later refetch silently discarded, page frozen until Retry | should-fix | invalidate `CAPABILITY_MATRIX_KEY` on success, then clear optimistic (`9b7c6571`) |
| F6 | `POST /capability/override` accepted `narrow`/`approval`, which the projection cannot express | should-fix | rejected with 400 + test (`f5524369`) |
| F7 | Stale docs on `OverrideRecord.active` and `project_matrix` tenant scope | nit | corrected (`52ae5421`) |

Each fix was verified by reverting it and confirming the named test fails.

### Checklist

| Item | Result |
|---|---|
| CI | **33 pass / 0 fail** (8 skipped) — verified against the PR head SHA |
| SonarCloud | gate **OK**, **0 open issues** (4 were open at review start: 2 CRITICAL cognitive-complexity + 2 MINOR duplicate import — all fixed as pure refactors, `ea225a34` `d37ce27f` `27aa6dd5`) |
| CodeCov | `codecov/patch` **success** |
| Scope vs AAASM-5090 | **complete** — `createMockCapabilityClient` and the server-side `seeded_matrix()` fixture both removed. Not over-scoped: `CAPABILITY_MATRIX_FIXTURE` correctly survives as test data for two unrelated e2e specs |
| OpenAPI | path count **stays 72** — the endpoint already existed serving a fixture; it was converted, not added. Contract test untouched, drift check green |
| Enforcement isolation | **0 files** under `aa-gateway/ aa-runtime/ aa-proxy/ aa-ebpf* aa-core/ aa-sandbox/` — verified by diff, not by assertion |
| Side effects | none found. The override store's move from mutable-copy to append-only overlay is **strictly safer** — the old restore-to-snapshot destroyed a still-active later override when revoking an earlier one; replay composes correctly. `lastSeen` ISO-8601 change has no stranded consumer |
| Playwright (light + dark) | **2/2 passed**, zero console errors — `dashboard/verify/5090/review/` |

### Front-end validation evidence

Repo-local `@playwright/test` on an isolated port (not the MCP server — cross-session lock). Asserted the grid renders 2 agents × 4 resources with `allow`/`deny`/`na` addressable via `data-decision`; verb switching flips cell counts; **absent fields fold to `—` and never to `0`** (`trust`, owner, tool group, `sampleCalls` → "no recent calls"); filter bar narrows by search / framework / mode; zero `undefined`/`NaN` on the page; console errors and pageerrors asserted `[]`.

### Honest caveats

- Absent by design, each owned by a ticket, never faked: `sampleCalls[]` → AAASM-5094 · `trust` → AAASM-5083 · `policies[].hits24h` → AAASM-5096 · `flagged`/`note`/`flag` · tool `group` · `paths`.
- `narrow`/`approval` remain structurally unreachable from a static capability set (they come from credential scrubbing and `requires_approval_if` CEL against a concrete action).
- **AAASM-5102** — aa-api's `PolicyEngine` is never `.with_registry(..)`-wired, so `GET /agents/{id}/capabilities` drops all Org+Team scoped policy including denies. Found here, filed separately, deliberately **not** fixed in this PR. This PR sidesteps it by passing an explicit lineage, now guarded by a test.
- One judgement call left as-is: `project_rules` still emits a rule row with `resource: "*"` for a wildcard tool entry. It can never match a cell and it accurately describes the policy.

Reviewed by Claude Code. Findings were verified against the source directly, not taken from the implementing agent's report.

PR #1700, opened 2026-07-26
[AAASM-5099] ✨ (aa-api): Widen the topology graph to all six edge kinds + effective perms

Chisanan232:
## Description

Widens the `GET /api/v1/topology` graph projection — no new route, no new data
source, no enforcement change.

**Edges.** The edge store and `/topology/edges` have carried all six relation
kinds since AAASM-985, but the graph projection walked only `DelegatesTo` and
`Calls`. It now walks `EdgeType::ALL`. The two structural kinds keep the wire
vocabulary AAASM-5040 shipped (`delegates_to` → `delegation`, `calls` → `call`)
so nothing the frontend already renders breaks; the other four pass their
canonical wire string through (`reads`, `writes`, `approves`, `messages`). Each
edge additionally reports `cross_team`, computed with the same rule
`edges::compute_cross_team` uses for `is_cross_team`, so the graph and the edge
list cannot disagree about one edge.

**Effective permissions.** Each node now carries `effective_permissions`: one
row per cascade tier that applies to the agent (`Global → Org → Team → Agent`)
with the policy documents loaded there, plus the merged capability set that walk
produces. This unblocks the Topology edge-kind checkboxes and the
policy-inheritance chain from AAASM-5071 — both shipped inert waiting on this
data.

**Dashboard.** All six edge-kind checkboxes are live and actually filter the
canvas (the four that shipped disabled with a "soon" tag are now toggleable and
styled); the cross-team badge, canvas curves, and toggle read the server's
`cross_team` flag instead of re-deriving it; the node-detail panel renders the
real Policy-Inheritance block the design calls for.

### Read-only — no enforcement or hot-path change

`git diff --name-only remote/main..HEAD` touches `aa-api/`, `openapi/`,
`dashboard/`, `aa-integration-tests/tests/api_topology.rs`, and
`docs/src/adr/0017-dashboard-design-parity-ratified-evolutions.md` — nothing else.
Grep-verified untouched: `aa-runtime/`, `aa-gateway/`, `aa-proxy/`, `aa-ebpf*`,
`aa-core/`, `aa-sandbox/`, `aa-sdk-client/`, `aa-security/`. The projection reads the same registry,
edge store, policy engine, and budget tracker its neighbours already read, under
the same `RequireRead` + tenant-scoped, deny-by-default guard.

### Path count unchanged

`openapi/v1.yaml` still declares **72** paths — this widens an existing
projection and adds no route, so
`aa-integration-tests/tests/api_openapi_contract.rs` needed no edit (verified, 7/7 pass).

### One existing test intentionally changed

`api_topology.rs::topology_graph_returns_nodes_with_badges_and_delegation_call_edges`
posted a `messages` edge and asserted it was **not** graphed — it pinned exactly
the narrowing this ticket removes. That assertion described the old deliberate
behaviour, so it was updated to the new one (post one edge of every stored kind,
require all six back) and renamed, rather than worked around. A second case was
added for the two new wire fields on the live HTTP surface: a cross-team edge
reporting `cross_team=true`, and each node carrying the inheritance chain with
its **team tier present** — the tier that silently vanishes under the AAASM-5102
lineage trap. `api_topology` is 17/17.

### How the AAASM-5102 lineage trap was avoided

`AppState::local_in_memory` never calls `.with_registry(..)`, so
`PolicyEngine.registry` is permanently `None` in the aa-api process.
`collect_cascade()` — and `effective_permissions()` built on it — therefore
resolve `Lineage::default()` and silently walk only Global and Agent, dropping
every Org- and Team-scoped allow **and** deny.

This PR does **not** use those. `project_graph_nodes` resolves the lineage
explicitly from `state.agent_registry.lineage(..)` and calls
`collect_cascade_with_lineage(..)` — the same guard `capability::project_matrix`
uses (AAASM-5090). `a_team_scoped_policy_reaches_the_permission_chain` fails, and
is the *only* failure, if someone regresses that call to `collect_cascade`
(verified by temporarily making the change).

**Do not delete `a_team_scoped_policy_reaches_the_permission_chain` thinking the
integration test covers it.** The HTTP-level `tiers.contains("team")` assertion in
`api_topology.rs` is *not* sufficient on its own: `build_policy_chain` takes
`lineage` as its own argument, so a partial regression — one that keeps resolving
the lineage for the chain but drops it from the cascade — would still emit a team
row, just an empty one, and the integration assertion would pass. The unit test is
the real regression guard because it asserts on the merged capability set the
cascade produces, not merely on the presence of the tier.

The pre-existing `policy_count` in this same handler had the identical bug — it
issued its own lineage-less `collect_cascade`. It now reads off the one corrected
cascade rather than leaving two disagreeing resolutions in one function.
AAASM-5102 itself is untouched; it is separately owned.

`allow` / `deny` / `allow_restricted` mirror **three** of the four stages
`evaluate_single_doc` (`aa-gateway/src/engine/decision.rs`) runs — see the review
pass below. `stage_approval` is not mirrored (it needs a concrete action) and the
doc comment says so explicitly rather than overclaiming. Every mirrored `stage_*`
is named in the code so future drift is greppable.

### Fields deliberately left absent

Nothing is fabricated. Everything emitted projects state that already exists.

| Left absent | Why |
|---|---|
| **Parent-agent tier** in the inheritance chain (the mock's "③ parent" row) | `PolicyScope` is `Global \| Org \| Team \| Agent \| Tool` — there is no parent tier, and a parent agent's own `agent:`-scoped policies are **not** inherited by its children. Emitting the row would mean inventing a derivation rule. Documented on `NodeEffectivePermissions`. |
| **Tool tier** | Selected per *action* (`aa_gateway::engine::action_tool_name`), not per agent — there is no agent-level answer to project. |
| **Org / Team tier** for an agent with no `org_id` / `team_id` | The tier genuinely does not apply to that agent. The chain lists only the tiers it has; the UI renders what it receives. |
| **`effective_permissions` on `AgentNode`** from the list / tree / team endpoints | Those projections do no policy-engine lookup. `null`, exactly like `policy_count` / `budget` — the panel folds to `—` rather than showing an empty-but-authoritative chain. |
| **A `narrowed` / `baseline` verdict on the wire** | The design's "→ effective" verdict has no server-side source. The panel derives its wording from the real `allow` / `deny` / `allowRestricted` fields instead, so no product decision is embedded in the contract. Recorded as a deviation in the ADR-0017 addendum. |

A tier that *applies* but carries no document renders `none`, kept deliberately
distinct from the `—` no-data affordance: "no team policy" and "we don't know"
are different claims.

ADR-0017 was checked first — its three ratified Topology entries (force-directed
layout, team budget bars, 5s polling) are left untouched. This PR does introduce
three *new* deviations from `design/v1/hi-fi/topology.jsx`, now recorded in an
**addendum** to that ADR: no `③ parent` row (`PolicyScope` has no parent tier),
`① org baseline` rendered as a `global` tier, and the `→ effective` verdict derived
client-side rather than hardcoded to `narrowed (pending)` / `baseline`. The
addendum is appended, so the reconciliation program's item numbering and count
note are unchanged.

## Review pass (post-first-review)

Four findings, each fixed with a test that fails without the fix (verified by
reverting the fix and confirming the named test — and only that test — goes red).

**S1 — the permission projection mirrored only the capability stage.**
`evaluate_single_doc` runs `stage_network` → `stage_tool_allow` → `stage_capability`
→ `stage_approval` and returns on the first `Deny`, so anything the first two block
never reaches the merged capability set. The projection read only the last of them,
yet its doc comment claimed the panel "cannot drift from what the gateway
enforces" — false as written. Concretely: a cascade with
`tools: { "*": { allow: false } }` and no `capabilities` block denies every tool
call, but the node emitted `allow: [], deny: [], allow_restricted: false` and the
panel read "baseline — no capability restriction"; same for
`network: { allowlist: [] }`. Worse in the permissive direction, a document
granting `mcp_tool:search` in `capabilities.allow` while another denied it via
`tools: { search: { allow: false } }` put `mcp_tool:search` in the wire `allow`
array — advertising a permission the engine refuses.

Fixed by folding the earlier stages in, exactly as `project_matrix` does.
`cascade_denies_tool`, `cascade_denies_all_egress` and `agent_tool_ids` were
**promoted out of `routes/capability.rs` into `routes/enforcement_mirror.rs`** so
both routes call one implementation rather than two copies that can drift. The
promotion is behaviour-preserving for the capability route: its tests are
untouched and all 19 still pass, including
`every_stage_that_denies_is_visible_in_the_cells` and
`the_tool_wildcard_never_becomes_a_resource_column`. The doc comment now names
each mirrored `stage_*` and states plainly that `stage_approval` is *not* mirrored
and why that cannot make the projection permissive. It also states the one honest
limit: a `"*"` wildcard deny also blocks tools nobody has named, so absence from
`deny` is not a grant.

New tests: `a_wildcard_tool_deny_reaches_the_permission_set`,
`a_tool_stage_deny_beats_a_capability_grant`,
`an_empty_network_allowlist_denies_outbound_access`.

**S2 — `deny` under-stated the engine by one capability.**
`aa_core::capability_is_denied` treats `deny(file_write)` as implying
`deny(file_delete)` (the fail-closed migration from AAASM-4103). The projection
emitted the raw merged deny set, so a cascade denying `file_write` reported
`deny: ["file_write"]` while the engine denies two. Under-reporting a denial is
the permissive direction. Fixed by reusing `capability_is_denied` rather than
restating the rule. New test: `a_write_deny_reports_the_delete_it_also_blocks`.

**N2 — TypeScript name collision.** `features/topology/types.ts` exported
`EffectivePermissions` while `features/agents/api.ts` already exports a *different*
type under that exact name (an alias of `EffectivePermissionsResponse` — `merged` +
`sources`). Renamed to `NodeEffectivePermissions`, mirroring the Rust type and the
generated schema name, and its consumers updated.

**S3 — new design deviations recorded.** ADR-0017 addendum, as described above.

## Type of Change

- [x] ✨ New feature

## Breaking Changes

- [x] No

`TopologyGraphEdge` gains a field and emits more rows; `AgentNode` gains an
optional field. The two existing `kind` values are unchanged, and every new
field is additive.

## Related Issues

- Related Jira ticket: [AAASM-5099](https://lightning-dust-mite.atlassian.net/browse/AAASM-5099)
- Parent Epic: [AAASM-5082](https://lightning-dust-mite.atlassian.net/browse/AAASM-5082)
- Unblocks: [AAASM-5071](https://lightning-dust-mite.atlassian.net/browse/AAASM-5071)

## Testing

**Rust** — 15 new tests in `routes/topology.rs::graph_tests` (11 from the original
pass, 4 from the review pass above): every one of the
six kinds reaching the graph, the structural kinds keeping their vocabulary,
cross-team true/false/team-less, the team-scoped-lineage guard, tier omission for
an agent without an org, the `allow_restricted` carry-through, authz denial for
an unscoped caller, an out-of-tenant peer dropped across *all six* kinds, and an
empty registry.

```
cargo nextest run -p aa-api --lib
     Summary [   2.822s] 437 tests run: 437 passed (2 leaky), 0 skipped

cargo nextest run -p aa-integration-tests --test api_topology
     Summary [   0.994s] 17 tests run: 17 passed, 0 skipped

cargo nextest run -p aa-integration-tests --test api_openapi_contract
     Summary [   0.261s] 7 tests run: 7 passed, 0 skipped

cargo fmt --all --check     → clean
cargo clippy -p aa-api --all-targets -- -D warnings
cargo clippy: 0 errors, 6 warnings   (all 6 are workspace Cargo.toml
                                      `default-features is ignored` notices,
                                      pre-existing and unrelated)
```

**Dashboard** — new coverage on the hook's success *and* error branches, the
mapper (six kinds, unknown-kind drop, chain mapping, absent-chain → `null`), the
sidebar (all six enabled, a newly-emitted kind toggling), the canvas (six kinds
drawn, filtered, server flag beating the client derivation, and the fallback),
the page's cross-team counter, and the panel's tier rows / effective summary /
deny-all case / `—` fold.

```
pnpm test        → Test Files 228 passed (228) | Tests 1973 passed (1973)
pnpm type-check  → clean
pnpm generate:api && git status --porcelain → empty
```

**Playwright** — `tests/e2e/review-aaasm-5099.spec.ts` on its own port (4599),
repo-local `@playwright/test`, both themes: `PASS (2) FAIL (0)`. Asserts the
checkboxes genuinely filter, the cross-team toggle removes exactly the flagged
edges, the inheritance chain renders, an absent chain shows `—`, and console
errors are `[]`. Screenshots committed under `dashboard/verify/5099/`.

## Checklist

[AAASM-5099]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — review record (AAASM-5099)

**Verdict: READY** ✅ — no blocking issues. Four review findings were applied on this branch before sign-off.

### The question this review was built around

The previous PR (#1699, AAASM-5090) shipped a real fail-open: a projection read only the capability stage and reported `allow` for tools the gateway denies at `stage_tool_allow`. The same question was asked of this PR.

**It does not repeat as user-visible harm.** The permission panel renders only counts and policy *document* names — never a per-tool verdict — so no cell on screen can contradict the engine the way #1699's did. Two things actively reduce the risk: `merge_capabilities` **intersects** allow-lists rather than unioning them (`aa-core/src/capability.rs:174-179`), and `allow_is_restricted()` is carried through verbatim, so the AAASM-4154 deny-all-with-empty-allow case cannot read as permissive.

**But there was a residual on the wire contract, and it has been fixed** — see S1.

### Findings applied

| # | Finding | Sev | Fix |
|---|---|---|---|
| S1 | `project_effective_permissions` mirrored **only** stage 3.5 (`stage_capability`). Stages 2 (`stage_network`) and 3 (`stage_tool_allow`) were not folded in — while the sibling `project_matrix` folds both, precisely because a permission projection that omits them is wrong. A team policy with `tools: {"*": {allow:false}}` emitted `allow: [], deny: [], allow_restricted: false` and the panel read "baseline — no capability restriction" for an agent that could not invoke a single tool; a `capabilities.allow` grant contradicted by a per-tool `allow:false` put that tool in the public `allow` array. The function's own doc comment claimed "what the panel shows cannot drift from what the gateway enforces" — false as written. | should-fix | `cascade_denies_tool`, `cascade_denies_all_egress` and `agent_tool_ids` promoted out of `capability.rs` into a shared `pub(crate)` module `routes/enforcement_mirror.rs` — **one implementation, not a copy**. `deny` is now built first (network deny-all, tool-stage denies), then `allow` is filtered so a blocked capability can never appear as granted. Doc comment rewritten to name each mirrored `stage_*` and to state the two things it **cannot** do (`stage_approval` needs a concrete action; a `"*"` deny also blocks unnamed tools, so absence from `deny` is not a grant) rather than overclaim. |
| S2 | `deny` under-stated the engine: `capability_is_denied` treats `deny(file_write)` as implying `deny(file_delete)`, but the raw merged set was emitted — under-reporting a denial, the permissive direction. | should-fix | reuses `aa_core::capability_is_denied` rather than restating the rule. |
| N2 | Two different `EffectivePermissions` types in one app (`features/topology/types.ts` vs `features/agents/api.ts`). The Rust side was renamed for exactly this reason; the FE had kept the collision. | nit | renamed to `NodeEffectivePermissions`, mirroring the Rust intent. |
| S3 | Three **new** deviations from `design/v1/hi-fi/topology.jsx` were unrecorded: the mock's `③ parent` row dropped (no parent tier exists in `PolicyScope`), `① org baseline` → `global` tier, `→ effective` wording derived client-side. | should-fix | recorded as an addendum to ADR-0017, appended so the AAASM-5077 item numbering and the three existing ratified Topology entries are untouched. |

Tests that now fail without the S1/S2 fixes: `a_wildcard_tool_deny_reaches_the_permission_set`, `a_tool_stage_deny_beats_a_capability_grant`, `an_empty_network_allowlist_denies_outbound_access`, `a_write_deny_reports_the_delete_it_also_blocks`. Verified by reverting the fix and observing exactly those four go red while all 11 pre-existing `graph_tests` stayed green.

**The shared-helper promotion is a pure move** — independently verified: `git diff --stat` on `capability.rs` is **2 insertions / 67 deletions**, the insertions being `use` lines and the deletions the three moved definitions verbatim. No logic line changed, no capability test touched; `every_stage_that_denies_is_visible_in_the_cells` and `the_tool_wildcard_never_becomes_a_resource_column` still pass.

### Checklist

| Item | Result |
|---|---|
| CI | **37 pass / 0 fail** (9 skipped) — verified against head `6f2d7015` |
| SonarCloud | gate **OK**, **0 open issues**, `new_coverage` **93.3%**, duplication 0.0%, hotspots 100% reviewed, all ratings A. Scan confirmed post-push, not stale |
| CodeCov | `codecov/patch` **success** |
| Scope vs AAASM-5099 | **complete** — all 6 kinds, `cross_team`, the effective-permission chain; AAASM-5071's edge-kind checkboxes and inheritance chain are unblocked |
| OpenAPI | path count **72**, unchanged — same `edge_repo`/`policy_engine`/`agent_registry`, no new data source |
| Enforcement isolation | **0 files** under `aa-gateway/ aa-core/ aa-runtime/ aa-proxy/ aa-ebpf* aa-security/ aa-sandbox/` — verified by diff |
| Lineage correctness | the chain resolves `agent_registry.lineage(..)` → `collect_cascade_with_lineage` on **every** path; it is the only construction site for `NodeEffectivePermissions`, so there is no unlineaged path reintroducing the AAASM-5102 fail-open. The PR also fixed `policy_count`, which had the same lineage-less bug and under-counted |
| Cross-tenant leak | impossible — `teams_by_id` is built from visible records only and doubles as the visibility set; a test inserts an out-of-tenant edge of **all six** kinds and asserts zero survive. `cross_team: true` reaches only callers who already see both endpoints |
| Playwright | light + dark, port-pinned to 4599; fixture deliberately includes a node with **no** `effective_permissions` key to exercise the absent path → `dashboard/verify/5099/` |

### The modified existing test — strengthened, not weakened

An integration test asserted `messages` edges are NOT graphed — it pinned the exact narrowing this ticket removes, and CI caught it. The rewrite posts one edge of **all six** kinds and asserts an **exact sorted equality** on the kind set (strictly stronger than the old membership checks), plus `cross_team` present-and-false per edge, plus a wholly new cross-team + permission-chain case over live HTTP. Every other removed assertion in the diff was audited for the same pattern; `mapGraph.test.ts` even **gained** a guard that an unknown kind is still dropped, preserving soundness the rename could have quietly lost. Nothing was loosened to make CI pass.

⚠️ **Do not delete `a_team_scoped_policy_reaches_the_permission_chain`.** It is the real lineage regression guard. The HTTP-level `tiers.contains("team")` check is **not** sufficient on its own — `build_policy_chain` takes `lineage` as its own argument, so a partial regression would leave the team row present-but-empty and still pass.

### Known, tracked, not fixed here

**AAASM-5102** — `aa-api/src/routes/agents.rs:633` still calls the unlineaged `effective_permissions()`, so `GET /agents/{id}/capabilities` still drops Org/Team tiers. Correctly out of scope, but it means **two aa-api surfaces disagree about the same agent's permissions** until 5102 lands.

Reviewed by Claude Code. Findings were verified against the source directly, not taken from the implementing agent's report.

PR #1701, opened 2026-07-26
[AAASM-5102] 🐛 (aa-api): Wire the agent registry into the policy engine

Chisanan232:
## Description

`AppState` never called `PolicyEngine::with_registry`, so `PolicyEngine.registry` was permanently `None` in the aa-api process. `collect_cascade` then fell back to `Lineage::default()` and walked **only the Global and Agent tiers**, so `GET /api/v1/agents/{id}/capabilities` (which goes through `effective_permissions` → `collect_cascade`) silently omitted every **Org- and Team-scoped allow *and* deny**. Because the denies were dropped too, the endpoint could report a capability as permitted while a team policy denied it — a **reporting fail-open**. It surfaced in `aasm policy show --show-permissions`, `aasm topology lineage --show-permissions`, and the dashboard's inherited-permissions panel, and it made `/agents/{id}/capabilities` disagree with `/topology` about the same agent.

**Enforcement was never affected.** `PolicyEngine::evaluate` deliberately avoids `collect_cascade`: it resolves tenancy itself via `authoritative_lineage(ctx)` and calls `collect_cascade_with_lineage` (AAASM-3729). Real verdicts always saw Org/Team policy. This PR changes reporting only — no response schema changed, only the data inside the existing shape.

### The fix

`AppState::local_in_memory` now attaches the registry: `.with_registry(Arc::clone(&agent_registry))`.

**A second, load-bearing detail:** the shipped `serve_local` entrypoint goes through `local_hardened_at`, which *replaces* `state.agent_registry` with a durable SQLite-backed registry **after** the base wiring. Wiring `local_in_memory` alone would have left the engine holding the discarded throwaway registry — an always-empty registry, i.e. exactly the `Lineage::default()` fallback this ticket removes — so the fix would never have reached production. The durable registry is therefore built *first* and threaded into the base wiring via a new private `local_in_memory_with_registry`, and the trailing `state.agent_registry = registry` assignment is gone. Nothing else about how state is assembled changed.

### Blast radius

aa-api calls exactly four `policy_engine` methods: `effective_permissions` (1×, `agents.rs:633` — the bug), `collect_cascade_with_lineage` (2×, already lineage-explicit), `active_policy_info` (2×), `simulate` (1×). Only the first changes behaviour. `simulate` shares the engine's registry, so a dry-run of a *registered* agent now resolves its cascade from the registered owner instead of the client-supplied context — that is the designed behaviour (AAASM-3729/3138) and matches what a real evaluation does; unregistered agents still fall back to the request context, unchanged.

No enforcement crate was touched — `git diff --name-only remote/main..HEAD` is four files, all under `aa-api/src/`.

**No existing test asserted the old (broken, Org/Team-dropped) numbers** — the full `aa-api` package (1014 tests, lib + integration) is green with no expectation edits. The only pre-existing tests in this area (`capability::project_matrix`, `topology::project_graph_nodes` and their guards) deliberately pass an *explicit* lineage to sidestep this bug, so they were already asserting the corrected numbers and keep passing untouched. Their doc comments, which asserted "the engine `aa-api` builds is not registry-wired", were refreshed in a separate comment-only commit — the explicit lineage stays as the guard that keeps those projections correct under any engine.

### Other construction sites checked

- `aa-gateway/src/server.rs` `serve_tcp` / `serve_uds` also build the engine without `.with_registry(..)` even though a registry is in scope. **Left as-is, out of scope:** aa-gateway has no production caller of `effective_permissions` / unlineaged `collect_cascade`, and enforcement resolves lineage itself, so it is latent rather than live. Worth a follow-up ticket.
- `aa-api/tests/common/mod.rs` builds its own unwired engine, but it constructs `AppState` by hand rather than through `local_in_memory`, so it is unaffected by this change. Left alone.
- `aa-cli/src/commands/policy/simulate.rs` builds a standalone engine for offline simulation with no registry available — correct as-is.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No

No schema change. `GET /api/v1/agents/{id}/capabilities` keeps its exact response shape; it now returns the Org- and Team-scoped rows it was dropping. A consumer that had (incorrectly) treated the endpoint as authoritative will now see additional `deny` entries and additional `sources` rows — which is the endpoint finally agreeing with enforcement and with `/topology`.

## Related Issues

- Related Jira ticket: [AAASM-5102](https://lightning-dust-mite.atlassian.net/browse/AAASM-5102)
- Related GitHub issues: N/A

## Testing

Two regression guards, both verified red without the wiring by reverting `.with_registry(..)` and re-running:

`aa-api routes::agents::tests::a_team_scoped_deny_reaches_the_capabilities_response`

```
thread 'routes::agents::tests::a_team_scoped_deny_reaches_the_capabilities_response' panicked at aa-api/src/routes/agents.rs:1502:9:
a Team-scoped deny must reach the merged set: []
```

`aa-api state::tests::local_hardened_engine_resolves_lineage_from_the_durable_registry`

```
thread 'state::tests::local_hardened_engine_resolves_lineage_from_the_durable_registry' panicked at aa-api/src/state.rs:716:9:
the engine must resolve the durable registry's lineage: {}
```

Both pass with the fix restored. Gates run locally:

```
cargo nextest run -p aa-api --lib   → 439 tests run: 439 passed, 0 skipped
cargo nextest run -p aa-api         → 1014 tests run: 1014 passed (1 leaky), 0 skipped
cargo fmt --all                     → clean
cargo clippy -p aa-api --all-targets -- -D warnings → exit 0
```

## Checklist

- [ ] All CI checks passing

[AAASM-5102]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aa-api/src/routes/agents.rs
+++ b/aa-api/src/routes/agents.rs
@@ -1378,4 +1378,136 @@ mod tests {
         let err = parse_agent_id("aabb").unwrap_err();
         assert_eq!(err.status, StatusCode::BAD_REQUEST.as_u16());
     }
+
+    // ── AAASM-5102: the cascade this endpoint reports must be tenancy-aware ──
+
+    fn admin() -> RequireRead {
+        RequireRead(AuthenticatedCaller {
+            key_id: "k".to_string(),
+            scopes: vec![Scope::Admin],
+            tenant: crate::auth::Tenant {
+                org_id: None,
+                team_id: None,
+            },
+        })
+    }
+
+    /// Minimal registered agent owned by `team_id`. Only the fields the
+    /// capabilities path reads (id, tenancy) carry meaningful values.
+    fn record(id_byte: u8, team_id: Option<&str>) -> aa_gateway::registry::AgentRecord {
+        aa_gateway::registry::AgentRecord {
+            agent_id: [id_byte; 16],
+            name: "a".to_string(),
+            framework: "langgraph".to_string(),
+            version: "0.1.0".to_string(),
+            risk_tier: 1,
+            tool_names: Vec::new(),
+            public_key: "pk".to_string(),
+            credential_token: "tok".to_string(),
+            metadata: BTreeMap::new(),
+            registered_at: chrono::Utc::now(),
+            last_heartbeat: chrono::Utc::now(),
+            status: AgentStatus::Active,
+            pid: None,
+            session_count: 0,
+            last_event: None,
+            policy_violations_count: 0,
+            active_sessions: Vec::new(),
+            recent_events: std::collections::VecDeque::new(),
+            recent_traces: Vec::new(),
+            layer: None,
+            governance_level: aa_core::GovernanceLevel::default(),
+            parent_agent_id: None,
+            team_id: team_id.map(str::to_string),
+            org_id: None,
+            depth: 0,
+            delegation_reason: None,
+            spawned_by_tool: None,
+            root_agent_id: Some([id_byte; 16]),
+            children: Vec::new(),
+            parent_key: None,
+            enforcement_mode: None,
+        }
+    }
+
+    /// A policy document carrying only the capability block the cascade reads.
+    fn policy_doc(
+        name: &str,
+        scope: aa_gateway::policy::PolicyScope,
+        capabilities: aa_core::CapabilitySet,
+    ) -> aa_gateway::policy::PolicyDocument {
+        aa_gateway::policy::PolicyDocument {
+            name: Some(name.to_string()),
+            policy_version: Some("1".to_string()),
+            version: None,
+            scope,
+            network: None,
+            schedule: None,
+            budget: None,
+            data: None,
+            approval_timeout_secs: 300,
+            approval_policy: None,
+            tools: Default::default(),
+            capabilities: Some(capabilities),
+        }
+    }
+
+    /// `AppState::local_in_memory` loads a budget-only policy file, so the
+    /// documents under test have to be pushed into the engine directly.
+    fn state_with_policies(
+        records: Vec<aa_gateway::registry::AgentRecord>,
+        docs: Vec<aa_gateway::policy::PolicyDocument>,
+    ) -> AppState {
+        let mut state = AppState::local_in_memory().expect("state builds");
+        for r in records {
+            state.agent_registry.register(r).expect("register");
+        }
+        let engine =
+            std::sync::Arc::get_mut(&mut state.policy_engine).expect("engine is unshared until the state is cloned");
+        for doc in docs {
+            engine.load_policy(doc);
+        }
+        state
+    }
+
+    /// Regression guard for AAASM-5102. `AppState::local_in_memory` never called
+    /// `PolicyEngine::with_registry`, so `effective_permissions` resolved
+    /// `Lineage::default()` and walked only Global and Agent — this endpoint
+    /// dropped every Org- and Team-scoped allow *and* deny, reporting a
+    /// team-denied capability as permitted (a reporting fail-open). Enforcement
+    /// was never affected; it resolves tenancy itself (AAASM-3729).
+    #[tokio::test]
+    async fn a_team_scoped_deny_reaches_the_capabilities_response() {
+        let mut caps = aa_core::CapabilitySet::default();
+        caps.deny.insert(aa_core::Capability::TerminalExec);
+
+        let state = state_with_policies(
+            vec![record(0x01, Some("team-alpha"))],
+            vec![policy_doc(
+                "team-rules",
+                aa_gateway::policy::PolicyScope::Team("team-alpha".to_string()),
+                caps,
+            )],
+        );
+
+        let (status, Json(body)) = get_agent_capabilities(
+            admin(),
+            Extension(state),
+            axum::extract::Path(hex::encode([0x01u8; 16])),
+        )
+        .await
+        .expect("admin may read a registered agent's capabilities");
+
+        assert_eq!(status, StatusCode::OK);
+        assert!(
+            body.deny.iter().any(|c| c == "terminal_exec"),
+            "a Team-scoped deny must reach the merged set: {:?}",
+            body.deny
+        );
+        assert!(
+            body.sources.iter().any(|s| s.scope == "team:team-alpha"),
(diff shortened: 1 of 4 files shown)


Chisanan232:
## Claude Code — review record (AAASM-5102)

**Verdict: READY** ✅ — no blocking findings.

### What this fixes
`AppState`'s policy engine was never `.with_registry(..)`-wired, so `PolicyEngine.registry` was `None`, `collect_cascade()` resolved `Lineage::default()`, and `GET /api/v1/agents/{id}/capabilities` silently dropped every Org- and Team-scoped allow **and deny** — a reporting fail-open. Enforcement was never affected (`evaluate` uses `authoritative_lineage` + `collect_cascade_with_lineage`, per AAASM-3729).

### The part that mattered — the obvious fix would not have worked
The naive fix (wire only `AppState::local_in_memory`) would have shipped green and left production broken. The shipped `serve_local` entrypoint (`aa-api/src/server.rs:133`) uses `local_hardened_at`, which calls `local_in_memory` and then **replaces** `state.agent_registry` with a durable SQLite-backed registry — `state.rs` calls the first one *"a throwaway in-memory `AgentRegistry`"* in its own comment. Wiring only the base builder would have left the engine holding the **discarded** registry: permanently empty, `Lineage::default()`, same bug by another route.

The PR instead threads the registry into the base wiring (`local_in_memory_with_registry`), builds the durable registry first, and deletes the trailing reassignment. Verified: engine and `state.agent_registry` hold the **same `Arc`** on every production path, and no other `state.agent_registry =` exists in `aa-api/src`.

### Verification

| Item | Result |
|---|---|
| CI | **29 pass / 0 fail** |
| SonarCloud | gate **OK** · new-code coverage **96.5%** · **0 open issues** |
| CodeCov | `codecov/patch` ✅ · `Coverage` ✅ |
| Files | 4, all `aa-api/src/` — **0** under `aa-gateway/ aa-core/ aa-runtime/ aa-proxy/ aa-ebpf* aa-sandbox/ aa-security/` |
| Schema | unchanged — this changes the DATA inside an existing shape, not the shape |
| Regression guards | two, complementary; one specifically catches the naive `local_in_memory`-only fix. Revert evidence: `a Team-scoped deny must reach the merged set: []` and `the engine must resolve the durable registry's lineage: {}` |
| Assertions weakened | **none** — full removed-line audit: the only deletions are four doc-comment blocks and the verbatim move of the durable-registry block. Zero `assert*` lines removed or altered |

**Blast radius, checked not assumed.** `self.registry` is read at five engine sites, not one. aa-api never calls `evaluate` (its embedded gRPC is `AgentLifecycleService` only), so only the `simulate` path is otherwise reachable, and it is inert (below). Cascade growth is bounded to +2 tiers; nothing hashes or caches a cascade (`DecisionCache` keys on `(agent_id, policy_epoch, action)` and isn't on this path). Downstream fixed-count consumers checked: `aa-cli/src/commands/permissions.rs:230` asserts a hardcoded fixture, `InheritedPermissionsPanel.tsx:94` renders `sources.length` dynamically. Cost: one extra DashMap read per request, adjacent to one the handler already does.

**Authz unchanged.** `authorize_agent_access` still gates on admin-or-own-team. The newly-returned rows are the agent's **own** Org/Team policy — the same data class `/topology` already returns for the same agent behind the same guard. No new cross-tenant exposure.

### ⚠️ Two corrections to the PR body — the code is right, the justification isn't

1. **The `simulate` behaviour change described in the body cannot happen.** `policies.rs:428` derives the id as `hash_to_16(bare_id)` (`aa-gateway/src/service/convert.rs:40`), while the registry is keyed on `SHA256("{org}/{team}/{agent}")[..16]` (`registry/convert.rs:9-14`). These key shapes can never coincide — the engine documents this itself at `aa-gateway/src/engine/mod.rs:1942-1943`. So `authoritative_lineage` always misses and falls back to ctx metadata exactly as before; `POST /policies/simulate` and the dashboard Simulate panel are byte-identical before and after. **Correct statement: simulate is unchanged.** Left uncorrected, a future ticket could "fix" simulate on a false premise.

2. **The proposed aa-gateway follow-up, as framed, would be a no-op.** `serve_tcp`/`serve_uds` do build the engine unwired, but adding `.with_registry(..)` there fixes nothing for the same key-shape reason: `ctx.agent_id` is `hash_to_16(bare agent DID)` and can never match the registry's composite key. The gateway is safe today because `apply_authoritative_tenancy` (`service/policy_service.rs:398-439`) overwrites `ctx.org_id`/`ctx.team_id` from its own registry before `engine.evaluate`. Any follow-up must be scoped as *"gateway lineage / `ProductionPolicyContext` lookups are keyed on a shape the registry doesn't hold"*.

### Should-fix (non-blocking)
- `aa-api/tests/common/mod.rs:84` still builds an **unwired** engine, so `tests/agents_detail.rs:73` and `tests/agents.rs:531` exercise the pre-fix path. They only assert `is_array()` today, so nothing pins broken data — but if someone later adds a content assertion there, they will codify the Global+Agent-only numbers, i.e. the exact regression class this ticket exists to kill. One `.with_registry(Arc::clone(&agent_registry))` fixes it.

### Nit
- Moving the durable-registry block to the top means an `aa-api-server` boot that fails later (rate-limit resolution, API-key parse, audit backend) now leaves an empty `~/.aasm/local.db` behind where it previously wouldn't. Harmless, worth knowing.

### Note on the Sonar verdict
This PR originally carried **no** Sonar or CodeCov result — `ci.yml:1229` gates those jobs on a `run-coverage` label, a dashboard change, or a push, and a Rust-only PR gets none of them. Because the `pull_request` trigger has no `types: [labeled]`, adding the label alone cannot re-fire the workflow, so an empty commit was pushed to obtain a real verdict rather than report an absence as a pass. **Standing gap: any Rust-only PR in this repo ships with no static-analysis verdict by default.**

Reviewed by Claude Code. Findings verified against the source directly, not taken from the implementing agent's report.

PR #1703, opened 2026-07-26
[AAASM-5096] ✨ (aa-api): Enrich PolicyResponse with affects[] and add the team→policy mapping

Chisanan232:
## Description

Enriches the policy contract so the two dashboard surfaces blocked on it can render real data instead of placeholders.

**`GET /api/v1/policies` — `PolicyResponse.affects[]`.** Per policy version, the agents it is *in force for*: the agents whose live policy cascade carries that version's document. Derived by re-parsing each stored snapshot into a `PolicyDocument` and matching it against a reach map built by walking every visible agent's cascade — grouped by `(scope, metadata.name)`, then confirmed by whole-document equality (see *Archived-version conservatism* below). That grouping key is the same one `capability::collect_policy_rows` uses, so the Policy list, the capability matrix and the new team view all name one document the same way.

**`GET /api/v1/policies/team/{team_id}` — the team→policy mapping.** The read-only inverse: the union of a team's agents' cascades, deduplicated by document, each carrying the cascade tier it comes from. Backs the Teams Active-policies card (AAASM-5080).

**Dashboard.** Policy list row gets the hi-fi affects chips (tail beyond three collapsed into `+N`) and the 24h hits block; Teams gets a fourth detail card. Both fold absent data to `—`.

Read-only throughout: no enforcement or hot-path change. Grep-verified — `git diff --name-only remote/main..HEAD` touches **zero** files under `aa-runtime/`, `aa-gateway/`, `aa-proxy/`, `aa-ebpf*/`, `aa-core/`, `aa-sandbox/`, `aa-security/`. `aa-api/src/state.rs` and `docs/src/adr/` are untouched.

> ### ⚠️ These fields have no production data source yet — AAASM-5106
>
> In **every shipped aa-api deployment the policy cascade is empty**, so both new surfaces resolve nothing today. `AppState::local_in_memory` loads the policy through `load_from_file`, which fills the engine's *primary policy slot* and leaves `scope_index` empty; only `read_cascade_dir` (reached via `load_cascade_from_dir*`) ever inserts into that index, and nothing in aa-api calls it. `apply_yaml` likewise stores to the primary slot only.
>
> **Enforcement is unaffected** — `PolicyEngine::evaluate` falls back to `evaluate_primary` when an agent's cascade is empty, so policy is still enforced. Only the *reporting* surfaces go blank.
>
> This is a pre-existing inherited defect, not introduced here, and is filed as **[AAASM-5106](https://lightning-dust-mite.atlassian.net/browse/AAASM-5106)**. This PR deliberately ships a **correct contract over a currently-empty source**: every field is designed so that "we cannot resolve this" is a distinguishable answer rather than a false statement. Wiring the cascade is 5106's job.

### The unknown-vs-none distinction, and why it is load-bearing

The original team endpoint returned `{"team_id": "...", "policies": []}` whenever it resolved nothing, and `TeamActivePoliciesCard` rendered that as **"No policy is in force for this team."** Under AAASM-5106 that sentence is *false*: the team's agents fall back to the primary policy slot, so a policy **is** in force — this projection simply cannot name it. A blank governance claim that reads as "you are ungoverned" is worse than no claim at all.

`TeamPoliciesResponse.policies` is therefore **required-but-nullable**, applying the same absent-vs-empty discipline `affects[]` already uses:

| Wire | Meaning | Card renders |
|---|---|---|
| `policies: [ … ]` | These documents are in force. | The list, with a scope chip per tier |
| `policies: []` | The team has no agent this caller can see, so there is genuinely nothing for a policy to be in force over. | "No policy is in force for this team." |
| `policies: null` | The team has agents but no resolvable cascade — the primary slot is enforcing and cannot be named (AAASM-5106). | "Policy data unavailable — the policy cascade is not currently loaded", count folds to `—` |

Required rather than optional so a client reads an explicit `null` it must handle instead of a missing key it can shrug off with `?? []` — which is exactly how "unknown" would decay back into "none". The query hook preserves `null` end-to-end, and the card resolves `loading | error | unknown | none | list` once instead of chaining guards. Both empty states are asserted **textually distinct** in unit tests and in the Playwright run.

### Archived-version conservatism

`(scope, name)` is a grouping handle, not an identity — `policy_version` is excluded from it, and from `PolicyDocument`'s identity. With `include_archived=true` that made every historical snapshot of e.g. `global/baseline` key onto the live document and **inherit its full `affects[]`**, asserting that a superseded version is in force; in the flat/non-envelope format, where `name` is `None`, every unnamed document at a scope collapsed into one key.

`affects` is now reported only when the parsed snapshot **equals** the one document the cascade carries (`PolicyDocument: PartialEq` over whole validated content — the strongest identity available, since a runtime document carries no id and no digest). Where identity cannot be established — an unreadable snapshot, or a key shared by two distinct live documents — the answer is the conservative **absent**, never an inherited reach. The same rule drops `TeamPolicyResponse.version` on a colliding key rather than reporting whichever document the registry walk reached first.

Relatedly, `PolicyValidator::from_yaml("")` *succeeds* — every section is optional — so an unretrievable snapshot used to parse into a permissive default document keying onto `("global", None)`. Blank snapshots are now rejected before they can key onto anything.

### The `tool:` tier is invisible to both surfaces

`collect_cascade_with_lineage` walks Global → Org → Team → Agent; `evaluate` appends the `tool:`-scoped tier itself, selected from the tool name of the action being evaluated (AAASM-3981). Tool selection is per-action, so no projection built from an agent alone can name it: a `tool:`-scoped document that **is** enforced never appears in `affects[]` or on the team mapping. Documented next to the existing lineage rationale on `cascade_for`. Reporting gap, not an enforcement one; closing it needs a per-`(agent, tool)` surface.

### Why a new path rather than a field on `GET /api/v1/policies`

`/api/v1/policies` is Admin-only — it discloses every tenant's raw policy YAML — but a team's own operator must be able to read the policies governing its team. The new path discloses no YAML, only document identities, so `RequireRead` plus membership in the requested team is the correct gate. **Path count 72 → 73**; `aa-integration-tests/tests/api_openapi_contract.rs` bumped (both the exact count and the sorted path list).

### Fields left absent, and who owns them

| Field | State | Why |
|---|---|---|
| `PolicyResponse.hits24h` / `TeamPolicyResponse.hits24h` | **Always absent** | No audit record attributes a decision to a policy *document*. `AuditEntry` has no policy field (`aa-core/src/audit.rs`); the payload's `policy_rule` is the free-text deny **reason** (`policy_service::evaluate_one`), empty on every allow, and `analytics.rs` groups by it at rule/reason granularity, not document. `PolicyDecision::Deny` carries a `source_scope`, but `into_policy_result` drops it before the audit write — and it is scope-granular, so it could not name a document even if it survived. Capturing the deciding document at the enforcement boundary is now owned by **[AAASM-5107](https://lightning-dust-mite.atlassian.net/browse/AAASM-5107)** (opened during this review pass, under Epic AAASM-5082). Emitting `0` would be indistinguishable from "fired zero times". |
| `PolicyResponse.affects` on a version the engine does not carry, or whose identity cannot be proven | Absent, never `[]` | `[]` would assert "this policy targets nobody", a stronger and different claim than "not currently loaded". See *Archived-version conservatism*. |
| `PolicyResponse.affects` from `POST /api/v1/policies` | Absent | `apply_yaml` swaps the primary slot only, never the scope index, so the just-applied version is in no cascade yet. |
| `TeamPoliciesResponse.policies` when the cascade resolves nothing for a team that has agents | `null`, never `[]` | See *The unknown-vs-none distinction*. |
| `affects[]` frameworks | Not emitted | The ticket says "agents/frameworks". `AgentRecord.framework` exists, but a policy scope targets agents, never frameworks — there is no `framework:` `PolicyScope`. Emitting frameworks would have meant inventing a derivation rule, so only the real agent set ships. |

### How the two known traps were avoided

**Lineage trap (AAASM-5102).** Every cascade is resolved as `state.agent_registry.lineage(..)` → `collect_cascade_with_lineage(..)`, never plain `collect_cascade` / `effective_permissions`. aa-api's engine holds no registry handle, so the unlineaged path yields a default `Lineage` and walks Global+Agent only — every Org- and Team-scoped document would silently vanish, which is exactly the failure mode `affects[]` and a team mapping would exhibit. This is correct whether or not the parallel 5102 fix lands. Two tests pin it with a **Team-scoped** document; swapping `collect_cascade` back in was verified to turn both red.

**Single-stage trap.** `affects[]` is a *membership* statement — this document sits in that agent's cascade — not a verdict. It makes no allow/deny claim, so no enforcement stage needs mirroring. The rationale is recorded in a doc comment on `policy_reach` naming `evaluate_single_doc` and its `stage_network → stage_tool_allow → stage_capability → stage_approval` order, so a future contributor who adds a verdict to this surface finds the evaluator and `routes/enforcement_mirror.rs` in the same grep.

## Type of Change

- [x] ✨ New feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Both new `PolicyResponse` fields are optional and omitted when absent, so existing clients are unaffected. `TeamPoliciesResponse.policies` is required-but-nullable, but that endpoint is introduced by this PR and has no released consumers. `topology::record_visible_to` became `pub(crate)` — crate-internal, no wire change.

## Related Issues

- Related Jira ticket: [AAASM-5096](https://lightning-dust-mite.atlassian.net/browse/AAASM-5096)
- Unblocks AAASM-5072 (Policy list-row affects/hits) and AAASM-5080 (Teams Active-policies card), under Epic [AAASM-5082](https://lightning-dust-mite.atlassian.net/browse/AAASM-5082).
- Depends on **[AAASM-5106](https://lightning-dust-mite.atlassian.net/browse/AAASM-5106)** for a live data source (pre-existing empty-cascade defect; enforcement unaffected).
- Opened during review: **[AAASM-5107](https://lightning-dust-mite.atlassian.net/browse/AAASM-5107)** — capture the deciding policy document at the enforcement boundary so `hits24h` becomes sourceable.

## Testing

**Rust — 12 tests in `aa-api/src/routes/policies.rs`:** the Team-scoped-lineage guard on both `affects[]` and the team mapping; absent-not-empty reach for a version the engine does not carry; a **superseded version of the same `metadata.name`** must not inherit the live document's reach; the team `version` is dropped when two documents share one key; an unreadable snapshot never parses into a permissive default document; `policies: null` (not `[]`) when the engine carries no cascade for a team that has agents; `policies: []` only for a team with no visible agent; `hits24h` omitted from the wire; a non-admin caller cannot enumerate policy versions; a caller scoped to another team is denied; an out-of-tenant member contributes no cascade.

Each fix was verified by reverting it and confirming its guard test turns red.

**Dashboard — 21 tests across the hook, card and row:** hook success / error / `[]`-preserved / `null`-preserved / missing-key-is-unresolved / not-fetched-until-selected; card loading / error / unknown / empty / populated-with-tiers / absent-hits fold / real-hits render / count folds to `—` / unknown-and-empty are textually distinct; row chips / `+N` collapse / absent-vs-empty reach / absent-hits fold / real-hits render. Error, unknown and empty are asserted as *distinct* states so neither a failed load nor an unresolvable cascade ever reads as "no policy is in force".

**Playwright evidence** (`dashboard/tests/e2e/review-aaasm-5096.spec.ts`, own port 4596): 4/4 passing, light and dark, console errors `[]`. Now drives **both** empty answers and asserts they differ. Screenshots committed under `dashboard/verify/5096/`, including the new `teams-active-policies-unknown-{light,dark}.png`.

Gate output:

```
cargo nextest run -p aa-api --lib              449 tests run: 449 passed, 0 skipped
cargo nextest run -p aa-integration-tests openapi_spec
                                                 6 tests run: 6 passed, 610 skipped
cargo fmt --all                                clean
cargo clippy -p aa-api --all-targets -- -D warnings   exit 0 (no lint warnings)
pnpm test                                      Test Files 229 passed | Tests 1994 passed
pnpm type-check                                clean
pnpm lint                                      ESLint: No issues found
pnpm generate:api && git status --porcelain    (empty — spec and generated types in sync)
playwright --config tests/e2e/review-5096.config.ts   PASS (4) FAIL (0)
```

## Checklist

- [ ] All CI checks passing

[AAASM-5096]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AAASM-5106]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — review record (AAASM-5096)

**Verdict: READY** ✅ — after one blocking finding and five others were fixed on this branch. But read the AAASM-5106 note below before deciding what merging this actually buys you.

### The blocking finding (fixed)

The PR was meticulous that `affects` is **absent, never `[]`** — "absent = we don't know" vs "`[]` = targets nobody" — and then shipped the exact opposite on the team endpoint: `{"policies": []}`, rendered by the Teams card as **"No policy is in force for this team."** A positive, false governance claim, made while a policy *is* in force through the engine's primary slot.

Fixed with the same discipline, applied properly:

| Wire | Meaning |
|---|---|
| `[ … ]` | these documents are in force |
| `[]` | no agent this caller can see — nothing for a policy to be in force over |
| `null` | agents exist, cascade resolves nothing → **unknown** |

`policies` is **required-but-nullable** (`required: [team_id, policies]`, `type: [array, 'null']`), surfacing in TypeScript as a **non-optional** `| null`. That choice is load-bearing: an *absent* key invites `?? []`, which is precisely how "unknown" decays back into "none". The unknown case never emits an array at all, so there is no `.length === 0` to misread. Card now renders *"Policy data unavailable — the policy cascade is not currently loaded"* with the count folded to `—`.

### All findings

| # | Finding | Sev | Fix |
|---|---|---|---|
| A | Team endpoint turned an unresolvable cascade into "no policy is in force" | **blocking** | required-but-nullable `policies`; card says *unavailable*, not *none* |
| B | `(scope, name)` is a grouping handle, not an identity — with `include_archived=true`, every historical snapshot of a name inherited the live document's full `affects[]`, asserting a superseded version is in force. Worse for the flat format, where `name: None` collapses every unnamed document at a scope to one key | should-fix | `affects` now requires the parsed snapshot to **equal** the document the cascade carries; ambiguous or unreadable → conservative absent |
| C | `hits24h` ownership was **circular** — `policies.rs` pointed at AAASM-5100 (whose scope is verdict/latency/trace_id, not document attribution) while `capability.rs` pointed back at 5096. After merge nobody would own it, and the pointer ships in the **published OpenAPI description** | should-fix | new ticket **AAASM-5107** created; both sites and the regenerated spec repointed |
| D | `tool:`-scoped documents are invisible to both surfaces — `collect_cascade_with_lineage` walks Global→Org→Team→Agent; the tool tier is appended per-action by `evaluate` | should-fix | documented as a reporting gap next to the lineage rationale; no behaviour change |
| F | `or_insert` picked an arbitrary document on key collision, so the card showed whichever version was seen first | nit | version reported only when the key resolves to exactly one document |
| G | **A real defect, not a confirmation.** `PolicyValidator::from_yaml("")` *succeeds* — every section is optional — yielding a permissive default document keyed `("global", None)`, colliding with any unnamed global policy | nit | blank/unparseable snapshot guarded before parsing |

Each fix verified by mechanically reverting it and confirming the named test turns red.

### Verification

| Item | Result |
|---|---|
| CI | **33 pass / 0 fail** (8 skipped) |
| SonarCloud | gate **OK** · new-code coverage **100%** · **0 open issues** · duplication 0.0% · hotspots 100% |
| CodeCov | `codecov/patch` ✅ |
| Files | `aa-api/`, `dashboard/`, `openapi/` only — **0** under `aa-gateway/ aa-core/ aa-runtime/ aa-proxy/ aa-ebpf* aa-security/ aa-sandbox/`. `aa-api/src/state.rs` untouched (PR #1701's branch unaffected) |
| OpenAPI | 72 → **73**; contract test bumped (exact count **and** sorted list). Only `-` lines are the old count and its comment — nothing loosened |
| Lineage trap | single cascade entry point, `lineage(..)` → `collect_cascade_with_lineage(..)`; no bare `collect_cascade` anywhere in the diff. Both guards use a `team:`-scoped document that the unlineaged path would drop |
| Playwright | 4/4 on port 4596, light + dark, driving **both** empty answers (`[]` and `null`) and asserting they differ |
| Assertions weakened | **none** — the only modified test file is the OpenAPI contract (count bump + list addition); all 8 Rust and 16 dashboard tests are new |

**Authz — I could not construct a leak.** `affects[]` is reachable only from `list_policies` / `get_active_policy`, both hard-gated on `Scope::Admin` before the reach walk, so it discloses nothing a caller could not already dump as raw YAML. The team path is deny-by-default: `can_access_team` fails closed for a non-admin whose `tenant.team_id != team_id`, including org-only and untenanted callers; each member is re-filtered through `record_visible_to`, which also fails closed. Identity only — no YAML, no rule bodies. The `record_visible_to` widening is **visibility-only**: `fn` → `pub(crate) fn`, body byte-identical.

**Frameworks correctly not invented.** `PolicyScope` is `Global | Org | Team | Agent | Tool` — no `framework:` variant exists, so emitting them would have required a fabricated derivation. The ticket says "agents/frameworks"; shipping agents only is the honest reading.

**`hits24h` escalation was honest.** Independently confirmed there is no per-document 24h signal: `AuditEntry` carries no policy field; `PolicyDecision::Deny { source_scope }` is dropped by `into_policy_result` and is scope-granular anyway; `analytics.rs` groups by the free-text `policy_rule`, which is rule/reason granularity. Shipping `0` would have been fabrication. Field stays absent, now owned by AAASM-5107.

### ⚠️ What merging this actually buys you — AAASM-5106

The **contract** is correct and complete. The **data source is not connected**. Filed as AAASM-5106 (High), pre-existing and inherited, not introduced here:

`PolicyEngine::load_from_file_with_budget` builds the engine with an empty `scope_index` (`aa-gateway/src/engine/mod.rs:695`); the only code that populates it is `read_cascade_dir` (`mod.rs:520-586`), reachable only via `load_cascade_from_dir*`, which **aa-api never calls** (`aa-api/src/state.rs:281` uses `load_from_file`, and its own doc says that is "the only public loader"). `apply_yaml` stores to the primary slot only (`mod.rs:796`) and never inserts into the index. Every `load_policy` call in aa-api is inside `#[cfg(test)]`, which is why every unit test passes.

So `collect_cascade_with_lineage` returns `vec![]` for every agent in production. **Enforcement is unaffected** — `evaluate` falls back to `evaluate_primary` when the index is empty (`mod.rs:813-816`). Only the reporting surfaces go blank. That asymmetry is the danger: the gateway denies, and the dashboard has nothing to show.

Same root cause makes AAASM-5090's capability matrix and AAASM-5099's topology chain inert in production today.

This PR is worth merging on its own terms — it is the first surface to make "we cannot resolve this" a *distinguishable answer* rather than a false statement — but it will render `—` and "unavailable" until 5106 is resolved.

### Process notes, disclosed
The implementing agent's first commit attempt used `git -c core.hooksPath=/dev/null` (equivalent to `--no-verify`, which this project forbids). It caught the mistake, reset, and recommitted through the hooks — which then rejected it on a formatting issue, confirming they ran. No commit in the pushed history bypassed hooks. Gates were run on the working tree and the content then rebuilt as 8 granular commits, byte-compared against the validated snapshot — so the gate output describes the pushed content, not each intermediate commit.

Reviewed by Claude Code. The wire shape, ticket repointing, OpenAPI required/nullable declaration, and the AAASM-5106 root-cause chain were verified against the source directly, not taken from the implementing agent's report.

PR #1705, opened 2026-07-26
[AAASM-5104] 🐛 (aa-api): Unify the trust field type and null contract

Chisanan232:
## Description

`trust` was declared two different ways across three schemas, with opposite
null-serialization contracts:

| Schema | Was | Null contract (was) |
|---|---|---|
| `CapabilityAgent` (`aa-api/src/models/capability.rs:205`) | `Option<u8>` | `skip_serializing_if` — **key absent** |
| `AgentNode` (`aa-api/src/models/topology.rs:279`) | `Option<f64>` | explicit `null`, never omitted |
| `AgentTree` (`aa-api/src/models/topology.rs:447`) | `Option<f64>` | explicit `null`, never omitted |

Today the value is `null` everywhere (AAASM-5083 has not landed), so nothing is
visibly broken — which is exactly why this is the cheapest moment to fix it.
This is a **prerequisite** of AAASM-5083, not a follow-up: the moment `trust`
carries a real number, the same agent serializes as an integer on one endpoint
and a float on another, and the generated TypeScript client carries two
incompatible shapes for one concept.

**`trust` remains `null` on every construction site. No score is computed here.**

### Representation chosen: integer 0–100

Evidence, in the priority order the ticket sets out:

1. **The ratified design mock renders a whole number.** `design/v1/hi-fi/fleet.jsx:90`
   prints `{score}` verbatim into a `width: ${score}%` bar; `agent-detail.jsx:27-43`
   prints it into the gauge's `<text>` and bands it at `<50` / `<75`;
   `capability.jsx:131-139` renders `trust {a.trust}` with `width: ${a.trust}%`;
   `topology.jsx:900` clamps drift with `Math.max(0, Math.min(100, n.trust + evt.delta))`
   over integer deltas. Every fixture value in `data.jsx` is a whole number
   (42, 78, 88, 71, …). Nothing in the mock is a 0–1 fraction, a percentage
   string, or a letter grade.
2. **ADR 0019 (Proposed, AAASM-5083) proposes the same thing** — line 275-278:
   *"The representation is reconciled first … Proposed: `Option<u8>`, emitted as
   explicit `null` (the `AgentNode` contract), since the score is an integer
   0–100 and `f64` implies a precision the formula does not have."* Cited as
   corroboration only — the ADR is **not treated as decided**, and this PR
   implements no part of its scoring proposal.
3. **The shipped dashboard already renders 0–100 integers.**
   `TrustBar.tsx:20-35` clamps to `[0,100]` and bands at ≥80/≥60;
   `AgentDetailPage.tsx:28-40` and `PerResourceTab.tsx:17-20` do the same.

All three agree, so there was nothing to break the tie on. The schemas now also
declare `minimum: 0` / `maximum: 100`, so the 0–100 range is machine-checkable
rather than a doc-comment claim (`u8` alone would admit 200).

### Null contract chosen: required-but-nullable

Following the precedent `AAASM-5096` just set on `main` for
`TeamPoliciesResponse.policies` (`aa-api/src/routes/policies.rs:632`,
`#[schema(required = true)]` on an `Option`): an **absent** key invites
`?? []` / `?? 0`, which silently turns "unknown" into a real answer, while an
explicit `null` on an always-present key forces the consumer to handle it and
surfaces in TypeScript as a non-optional `| null`.

For a trust score that reasoning is sharper than it was for policies: `?? 0`
does not just lose information, it produces *the worst possible value* —
"this agent scored zero" is the strongest negative claim the field can make,
and it would be manufactured out of no data at all. So `CapabilityAgent` moves
onto the `AgentNode` contract (explicit `null`, key required) rather than the
other way round. No concrete reason was found to depart from the precedent.

### Fallout the null contract exposed (and this PR fixes)

Making the key always present turned three dashboard guards into live bugs.
`CapabilityMatrixGrid.tsx:165`, `PerResourceTab.tsx:133` and `filters.ts:30`
tested `trust !== undefined`; with `null` on the wire they would take the
"has a score" branch — drawing `width: null%` with the danger tone (a
zero-width bar that reads as *scored zero*) and letting unmeasured agents
through every `trustMax` ceiling. `tsc` catches the first two once
`trust: number | null` lands; all three are now `!= null`.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes (describe below)

`trust` is now a nullable **integer** rather than a nullable **number/double**
on `AgentNode` and `AgentTree`, and is **required** (always present, possibly
`null`) on all three schemas rather than optional. Any consumer that treated an
absent `trust` as "no key" must now handle an explicit `null`. In practice the
wire value is `null` today and was `null` before, so no deployed response body
changes value — only its shape and its declared type. `openapi/v1.yaml` and
`dashboard/src/api/generated/schema.d.ts` are regenerated accordingly; the path
count is unchanged at **73**.

**Second-break risk, accepted:** pinning a scale now means that if AAASM-5083 /
ADR 0019 later settles on a *different* representation (a 0–1 fraction, say),
this becomes a second breaking schema change to the same field rather than one.
The trade-off is judged worth it — ADR 0019 proposes this exact 0–100 integer,
the ratified mock renders it, and the ADR's own Consequences section routes "the
`trust` type/serialization inconsistency" to a separate ticket (this one), so
reconciling first is the sequencing it asks for. Leaving the field divergent
until 5083 lands would instead guarantee a break *and* ship a period where one
agent serializes two incompatible shapes.

## Related Issues

- Related Jira ticket: [AAASM-5104](https://lightning-dust-mite.atlassian.net/browse/AAASM-5104)
- Prerequisite of: [AAASM-5083](https://lightning-dust-mite.atlassian.net/browse/AAASM-5083)
- Follows the null-contract precedent set by [AAASM-5096](https://lightning-dust-mite.atlassian.net/browse/AAASM-5096)

## Testing

### Changed test expectations — and why none of them is a weakening

Exactly **one** existing assertion changed side, plus one additive
strengthening. Both are listed with their justification.

1. `aa-api/src/routes/capability.rs` — `fields_without_a_real_source_are_absent_never_zero`
   (renamed `…_are_never_faked`, since `trust` is no longer absent).
   - **Was:** `assert!(json.get("trust").is_none())` — i.e. "no consumer can
     read a value from `trust`, because the key is not there".
   - **Now:** `is_some()` + `is_null()` + `!is_number()` + `assert_ne!(…, 0)`.
   - **Not a weakening:** the old assertion guaranteed exactly one thing — that
     `trust` is unreadable as a score. The new four assertions guarantee the
     same property against the new contract *and* add two claims the old one
     never made (not a number, not zero). The `mode` omission assertion in the
     same test is untouched. Nothing was deleted to make it pass.
2. `aa-api/src/models/topology.rs` — `agent_node_emits_trust_null_not_omitted`.
   - Purely **additive**: `!is_number()` and `assert_ne!(…, 0)` were added
     alongside the existing `is_some()` / `is_null()` assertions, which are
     unchanged. Strictly stronger.

3. `dashboard/src/features/capability/api.test.tsx` —
   `surfaces absent optional columns as undefined, never as 0`. Its
   `SPARSE_MATRIX` fixture had **no `trust` key** and the test asserted
   `expect(agent.trust).toBeUndefined()`, both self-described as "the live
   endpoint" shape. After this PR the endpoint can never omit `trust`, so both
   were false statements about the wire — and the next reader would reasonably
   conclude `trust` may be absent and write `agent.trust ?? 0`. Fixed:
   `trust: null` in the fixture, and `trust` moved **out** of the absent-columns
   test into its own `toBeNull()` / `not.toBeUndefined()` / `not.toBe(0)` /
   `'trust' in agent` assertion. **Not a weakening:** the old assertion made one
   claim (trust is not a number); the new one makes that claim plus three more.
   The `mode` / `flagged` / `hits24h` / `group` assertions — genuinely still
   absent — are untouched.
   The fixture was also cast `as unknown as CapabilityMatrix`, which suppressed
   the missing-required-property error and is why the first sweep missed it (the
   TypeScript equivalent of how `api_capability.rs:178` slipped through). The
   cast is now a plain `: CapabilityMatrix` annotation, restoring the
   type-checker as a net: deleting `trust: null` now fails with
   `TS2741 Property 'trust' is missing`.

No test was deleted, skipped, or relaxed. Fixture objects in the Teams
tests/stories gained an explicit `trust: null` — that is the contract's forcing
function (the key is now required), not a loosened assertion.

### CI-found miss, and the fix

The first CI run failed one test on both the Linux and macOS runners:
`aa-integration-tests/tests/api_capability.rs` —
`capability_matrix_omits_fields_with_no_real_source` asserted `trust` was
**absent** end-to-end. That file was outside the enumeration sweep and is the
only caller of the old contract outside `aa-api`. It is now
`…_never_fakes_a_field_with_no_real_source`, with the same deliberate
strengthening as (1) above: `is_some()` + `is_null()` + `!is_number()` +
`assert_ne!(…, 0)` replacing a single `is_none()`. The `flagged` / `note` /
`mode` / `hits24h` omission assertions in that test are untouched. Verified
locally: `cargo nextest run -p aa-integration-tests --test api_capability`
14 passed, `--test api_topology` 17 passed.

### Tests added

- `aa-api/src/openapi.rs` — one test over the generated OpenAPI asserting all
  three schemas declare `["integer","null"]`, `minimum: 0`, `maximum: 100`, and
  list `trust` as `required`. This is the cross-schema guard: the divergence
  this ticket fixes could previously only be caught by reading three files.
- `aa-api/src/models/capability.rs` — `CapabilityAgent` emits present-and-`null`
  (not a number, not `0`); a *missing* key still deserializes to "no score",
  proving dropping `skip_serializing_if` did not make the key mandatory inbound.
- `aa-api/src/models/topology.rs` — `AgentTree` carries the same null contract,
  and both `AgentNode` and `AgentTree` serialize a real score as a **whole
  number** (`is_u64()`), so one agent cannot be `78` on one projection and
  `78.0` on the other.
- `dashboard` vitest — an unmeasured agent is excluded by every `trustMax`
  ceiling (it is unscored, not "below"); the matrix grid and the per-resource
  table fold `null` to `—` with **no bar** and nothing matching
  `0` / `NaN` / `undefined` / `null`.
- `dashboard/tests/e2e/review-aaasm-5104.spec.ts` (+ `review-5104.config.ts`,
  port 4594) — drives all four surfaces that render trust (capability
  matrix / per-resource / per-agent, Fleet, agent-detail, topology) in **light
  and dark** against a payload shaped like the new contract. Screenshots in
  `dashboard/verify/5104/`.
  Verified **non-vacuous**: re-running the same spec with `trust: 0` on the wire
  fails the capability and topology assertions (`trust 0` renders, the topology
  badge appears).

### Gate output

```
cargo nextest run -p aa-api --lib
  Summary [3.208s] 456 tests run: 456 passed, 0 skipped

cargo nextest run -p aa-integration-tests openapi_spec
  Summary [0.166s] 6 tests run: 6 passed, 610 skipped
  (openapi_spec_loads_without_errors pins the path count — still 73)

cargo fmt --all                                        clean
cargo clippy -p aa-api --all-targets -- -D warnings     exit 0

cd dashboard
  pnpm test        Test Files 229 passed (229) / Tests 1997 passed (1997)
  pnpm type-check  clean
  pnpm lint        ESLint: No issues found
  pnpm generate:api → git status --porcelain shows no regeneration drift

pnpm exec playwright test --config tests/e2e/review-5104.config.ts
  PASS (6) FAIL (0)
```

Also run, since both directly assert the changed contract:
`cargo nextest run -p aa-api --test topology --test capability` — 44 passed.

## Checklist

- [x] All CI checks passing — 33 pass / 8 skipped / 0 fail (incl. `Test`, `Integration tests (macos-latest)`, `OpenAPI spec drift check`, `Dashboard codegen drift check`, SonarCloud Code Analysis, codecov/patch)

### Read-only scope

`git diff --name-only remote/main..HEAD` touches **zero** files under
`aa-runtime/`, `aa-gateway/`, `aa-proxy/`, `aa-ebpf*`, `aa-core/`,
`aa-sandbox/`, `aa-security/` — no enforcement or hot-path change. It also
touches no file under `docs/src/adr/`.

[AAASM-5104]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5104)

**Verdict: APPROVE** ✅ — after one MAJOR was found and fixed on this branch.

### Scope
`trust` was `Option<u8>` on `models/capability.rs` and `Option<f64>` on two `models/topology.rs` schemas, with **opposite null contracts** (skip-when-absent vs explicit null), both pinned by tests. Now **integer 0–100, required-but-nullable**, uniformly. Value stays `null` — the score itself is AAASM-5083. Full scope, no creep.

### The MAJOR (fixed, `f36569bb`)
`api.test.tsx` still pinned the **pre-5104** contract: a fixture documented *"as the live endpoint sends"* with no `trust` key, asserting `expect(agent.trust).toBeUndefined()`. After this PR the endpoint can never send an absent `trust`.

It was green **only because of an `as unknown as CapabilityMatrix` double-cast** suppressing the missing-property error. That is the same escape hatch that let `aa-integration-tests/tests/api_capability.rs:178` reach CI red — except TypeScript had no net to catch it.

Concrete harm: a contributor reads a test *named for the live-endpoint shape*, concludes trust may be absent, writes `agent.trust ?? 0` — the exact misread this ticket exists to prevent.

Fixed properly: fixture carries `trust: null`, assertion is `toBeNull()`, `trust` moved into its own test, and **the cast was dropped entirely** (`: CapabilityMatrix`). Negative control confirms the net is back — deleting `trust: null` now raises `TS2741`, previously suppressed.

**The fix round also corrected the review**: there are **five** `as unknown as CapabilityMatrix` fixtures, not three. The two the review missed (`useAgentPolicies.test.tsx`, `AgentDetailPage.test.tsx`) are clean, so the conclusion held — but the sweep was two short.

### Verified clean
**No test weakened** — raw-diff sweep of every removed line: exactly two assertions removed, each replaced by **four** (`is_some` + `is_null` + `!is_number` + `assert_ne!(…,0)`); the `mode`/`flagged`/`note`/`hits24h` omission assertions are untouched; fixtures gaining `trust: null` are forced by the type, not loosened.

**`u8` narrowing has no producer risk** — all construction sites are `None`; `78.5` → serde `invalid type: floating point`, `300` → `out of range`, never truncation; and nothing outside `aa-api` deserializes these structs (`aa-cli` defines its **own** `AgentNode`/`AgentTree` with no `trust` field and no `deny_unknown_fields`).

**Guard sweep complete** — zero remaining `!== undefined` / `typeof === 'number'` / truthiness / `'trust' in obj` guards on trust.

**Path count 73, unchanged.** OpenAPI ↔ `schema.d.ts` consistent; both drift checks green.

**Latent bug fixed as a side effect:** `api/capability.ts:31` casts the wire body asserting generated and hand-written types are "structurally identical". On `main` they were **not** (`number | null` vs `number`) and the cast hid it. This PR makes the claim true.

### One framing correction
The three `trust !== undefined` guards were **not** pre-existing bugs. On `main` the key was omitted, so the guards were **correct**; they become bugs only because this PR makes the key present — introduced and fixed in the same change. The PR body states this accurately.

### The Proposed-ADR-0019 question: **non-issue**
ADR 0019 is `Proposed`, and its lines 275-278 do propose `Option<u8>` + explicit null. But its own **Consequences** section says the trust type/serialization inconsistency *"should be separate tickets"* — the ADR **explicitly disclaims ownership and routes it to a separate ticket**, which is AAASM-5104. Authority was delegated, not front-run.

All five design-mock citations verified literally — `fleet.jsx:90`, `agent-detail.jsx:27-43`, `capability.jsx:131-139`, `topology.jsx:900`, and **22 fixture values in `data.jsx`, every one a whole number**. Zero fractions, zero 0–1 values, zero letter grades. ADR-0017 ratified design parity, so the mock is a decided input.

Residual risk is precision-only and now stated in Breaking Changes: if 5083 later picks a different scale this is a second breaking change — but that requires 0019 to be decided *against its own text*, and doing nothing guarantees a break the moment 5083 lands.

### Gates — stated exactly
**33 success / 8 skipped / 0 failing / 0 pending.** All 8 skips verified by design against `ci.yml`. Sonar gate **OK**, new coverage 100.0%, duplication 0.0%, **0 open issues**. `codecov/patch` pass.

⚠️ **`Coverage` (Rust) is label-gated off on PRs** (`ci.yml:378`), so **Rust patch coverage was never measured**. Sonar's 100% cannot be attributed between the Rust and dashboard diffs. The new Rust tests plainly cover the new lines, but "codecov green" is weaker evidence here than it looks.

### Merge order
`git merge-tree` is clean with **#1707 in both directions**, yet the merged product previously failed `tsc` — #1707 deleted `trust: 50` from two fixtures that become required under this PR. **Fixed on #1707's side.** Whichever merges second should still rebase and re-run `type-check`; a stale green check would hide a recurrence. No other shared contract: test ids disjoint, no CSS collisions, and #1707 **re-exports** `NO_DATA` rather than redeclaring it.

### Follow-ups, deliberately not fixed here
Three stale doc comments (`capability/api.ts:9-12`, `capability/display.ts:4-5` — also edited by #1707 — and `features/topology/types.ts:104`) · `minimum`/`maximum` are declarative only, so a future `Some(200)` from 5083 would serialize happily; a constructor guard belongs with that ticket · `TrustBar.tsx:21` / `AgentDetailPage.tsx:42` use `=== null` where this PR standardises on `!= null`.

Reviewed by Claude Code. Findings verified against the source directly, not taken from the implementing agent's report.

PR #1706, opened 2026-07-26
[AAASM-5082] 📝 (docs): Correct ADR-0017, re-anchor the visual spec to v2, and propose the open truthfulness decisions

Chisanan232:
> ## ⚠️ Read this first
>
> **Every new record in this PR is a `Status: Proposed` decision *proposal* awaiting
> sign-off. Merging it authorises no implementation, schedules no ticket, and changes
> no code behaviour.** The diff is Markdown only — three new ADRs, a correction
> appended to an existing ADR, `design/README.md`, and two index files. No `.rs`, no
> `dashboard/src/`, no `openapi/v1.yaml`, no generated client.
>
> Each new ADR carries an explicit sign-off gate naming who must sign (following the
> ADR 0018 precedent). Until those are signed, nothing here may be implemented.
>
> **Revision 2** — an adversarial review returned REQUEST CHANGES on six load-bearing
> claims. All six were independently re-verified in source and corrected in commits
> `bae6022`…`3ebc579`; see **"Review round 2"** at the bottom for the diff per finding.
> The ADR 0017 correction addendum remains **additive-only: 101 insertions, 0
> deletions**, appended at EOF.

## Description

Documentation and governance lane of the dashboard-truthfulness remediation programme
(Epic AAASM-5082). Three other lanes are changing code in parallel; this one changes
none.

**1. ADR 0024 — Semantics of an Empty or Unavailable Policy Cascade** (AAASM-5106).
ADR 0023 (#1704) asks *whether* `aa-api` should load a cascade. This is the separate,
orthogonal question that survives every answer to it: *what does an empty one mean?*
Today the answer is "allow everything", by accident — `decide`
(`aa-api/src/routes/capability.rs:480-488`) falls through both guards on an empty
`CapabilitySet` and returns `Decision::Allow`, so every cell of the Capability Matrix
renders `allow` — which is the grid's **neutral** treatment (`var(--paper-2)`, the plain
page surface), not an alarming one. A uniformly unremarkable grid reads as "nothing to
see here", which stops an operator looking more effectively than a loud colour would.

The ADR is precise about scope in both directions. **It is not a runtime enforcement
bypass:** `decide` is private to `aa-api`, and `aa-api` appears in none of
`aa-gateway/Cargo.toml`, `aa-runtime/Cargo.toml`, `aa-proxy/Cargo.toml` (verified —
zero matches); enforcement runs `PolicyEngine::evaluate` → `evaluate_primary` in
[...]

PR #1707, opened 2026-07-26
[AAASM-5173] ✨ (dashboard): Add shared truthfulness status and verdict primitives

Chisanan232:
## Description

Foundation lane for the dashboard truthfulness remediation programme. The design-QA re-audit found the dashboard **asserts things it does not know**: `0` rendered from failed requests, `<1ms` from absent latency, and — most seriously — a capability grid that asserts `allow` for every cell purely because an *empty* policy cascade constrained nothing.

This PR defines the shared vocabulary **once**, with tests, so the page-level lanes that follow **import rather than invent**. No two lanes may independently redefine the same status or verdict semantics — that is precisely why this lane exists.

**No page-level fixes were made.** Every other page keeps its current behaviour and gets its own ticket.

---

### 1. The status / unavailability vocabulary — `src/lib/truthfulness/absence.ts`

`TruthState` — the six explicit states the dashboard may display when it has no production value. **This enum is the contract; do not extend or alias it locally.**

| State | Means | Tone | Announcement |
|---|---|---|---|
| `unknown` | We asked and could not determine the value | caution | "Unknown — this value could not be determined." |
| `unavailable` | The request failed; we have no answer at all | **fault** | "Unavailable — the request for this value failed." |
| `unconfigured` | The capability exists but nothing is configured for it | caution | "Unconfigured — nothing is configured to produce this value." |
| `not-evaluated` | No evaluation has been performed | caution | "Not evaluated — no evaluation has been performed." |
| `not-supported` | The backend genuinely cannot provide this | neutral | "Not supported — the backend cannot provide this value." |
| `demo` | Fixture/demo data — illustrative, never a production fact | caution | "Demo data — illustrative sample, not a production value." |

They are deliberately **not** interchangeable: collapsing `unavailable` into `unconfigured` hides an outage behind a setup prompt; collapsing either into `not-supported` tells the operator to stop looking.

The carrier is a discriminated union:

```ts
type Certain<T> =
  | { known: true;  value: T }
  | { known: false; state: TruthState; detail?: string; sample?: T }
```

- `.value` is unreachable without narrowing through `isKnown`. **That is what makes it impossible to pass a number that is really an absence.**
- `demo` sits on the `known: false` side on purpose: its `sample` renders, but no aggregate anywhere can total demo rows into a production figure.
- `certain(value, whenMissing)` requires the caller to *state why* a value could be missing — the part a default value throws away. `null` / `undefined` / `''` are missing; `0`, `false`, `[]` are real values.
- `NO_DATA = '—'` is the one canonical inline affordance. `features/capability/display.ts` now re-exports it instead of declaring a second copy.

**Renderers** (`src/components/truthfulness/`):
- `TruthfulValue` takes a `Certain<T>`. There is no value-plus-fallback overload, so "just render 0" is not writable.
- `AbsenceMarker` gives every state the same learnable `—` glyph, with the state recoverable by colour tone, `title`, `data-truth-state`, **and always a screen-reader sentence** — a governance surface must never read as a clean bill of health.
- `StatusState` is the block-level surface. `role="alert"` is reserved for `unavailable`; everything else is a polite `status`. `state={null}` means a *genuinely empty* result (query succeeded, zero rows) — a known answer, not an absence.

### 2. Capability / verdict semantics — `src/lib/truthfulness/verdict.ts`

Grounded in what the backend actually emits. `aa-api`'s `decide()` (`aa-api/src/routes/capability.rs:480`) ends with:

```rust
Decision::Allow   // "Anything else is allowed because no capability rule constrains it."
```

`Allow` is the **fallback**, not a measurement. Under **AAASM-5106** the policy cascade is empty in every shipped deployment — `caps.deny` is empty, `allow_is_restricted()` is false — so **every cell resolves to `Allow`**: the grid asserts `allow` for every capability, an unbroken wall of `ALLOW`, while the gateway is denying those calls.

The defect is *data*, not logic: `decide()` is a line-for-line mirror of the gateway's `stage_capability` (`aa-gateway/src/engine/decision.rs:207-221`) — same deny check, same fail-closed `allow_is_restricted()` — so it is not drifting from enforcement, it is being handed a cascade nothing populated.

> **This is display-only.** `decide` is private to `aa-api` and no enforcement crate depends on `aa-api`. It is **not** a runtime bypass; the enforcement path is untouched. What is wrong is the claim the UI makes.

`resolveVerdict(decision, cascade)` is the single implementation of the rule. Precedence, strongest disqualifier first:

1. **Cascade absent** → propagate that absence. No verdict is trustworthy, not even a `deny`.
2. **No decision for the cell** → `not-evaluated`, never a default `allow`.
3. **`na`** → `not-supported` (the backend models no such verb for that resource).
4. **`allow` on an empty cascade** → `unconfigured`. ← **the AAASM-5106 rule**
5. Otherwise, a real rule-backed verdict.

**Never infer permission from missing policy data.** Step 4 is deliberately narrow: a `deny` / `narrow` / `approval` is a *positive restriction* and survives an empty cascade untouched — suppressing those would weaken what the operator sees, the opposite failure.

`tallyVerdicts` extends the rule to counts: on an empty cascade **all** counts are `unconfigured`, including the zeroes. "0 denied" reads as *we checked and found no denials*; nothing was checked.

`verdict.test.ts` guards this so no page lane can reintroduce it.

### 3. API error normalisation — `src/lib/truthfulness/query.ts`

`certainFromQuery(outcome)` is the one place a fetch/query outcome becomes the vocabulary. Error → `unavailable` (message kept as `detail`); pending → `unknown` (a slow request is not a broken one — a fault tone while data is in flight trains operators to ignore fault tones); empty payload → `whenEmpty` (default `unknown`); otherwise known, or `demo` if flagged.

The feature hooks under `src/features/*/api.ts` already `throw` at the *fetch* boundary. The gap this closes is the *render* boundary, where each consumer otherwise re-decides what to show — and where "render 0" keeps creeping back. Structurally typed against `UseQueryResult`, so a hook result passes straight in.

Documented gotcha: `openapi-fetch` captures `globalThis.fetch` at module load, so failures must be intercepted at the network layer (in tests, `page.route`), never by swapping `fetch`. This helper only ever sees the outcome.

---

### The competing state family, and how it was converged

Two families existed:

- `src/components/{Empty,Loading,Error}State.tsx` — the faithful port of the hi-fi `states.jsx`, with per-page copy tables.
- `src/components/states/{Empty,Error}State.tsx` — a generic, prop-driven pair that **bypassed the ported family on four call sites**: `PoliciesPage`, `LiveOpsPage`, `TraceViewPage`, and `features/analytics/PanelErrorBoundary`. Different roles, different tone, different markup for the same idea.

**Converged onto a new canonical `StatusState`**, which the generic pair now delegates to. **Props and test ids are preserved, so no call site needed editing — but the rendered output did change**: `<output>` became `<div role="status">`, the `.state*` classes became `.truth-state*`, and `ErrorState` now renders a visible `UNAVAILABLE` badge plus a screen-reader announcement (visible in `dashboard/verify/5173/status-state-unavailable-*.png`). Reviewers of `PoliciesPage`, `LiveOpsPage`, `TraceViewPage` and `PanelErrorBoundary` should expect that visual delta. Both are marked `@deprecated` for new work, which should use `StatusState` directly and say *why* a value is missing.

The ported family was **not** rewritten in this lane: its copy is page-visible, so changing it is a page-level behaviour change. Noted for the page lanes below.

### Surfaces wired as proof

**One surface, end to end: the Capability matrix summary row** (`CapabilitySummary` + `CapabilityPage` + `features/capability/{summary,api}.ts`).

With the empty cascade every shipped deployment resolves, the grid renders an unbroken wall of `ALLOW` — and the summary row now reports **Unconfigured / Unconfigured / Unconfigured / Not evaluated** instead of a permission total beside a reassuring `0 denied`. See `dashboard/verify/5173/summary-unconfigured-*.png`. With a real policy document loaded, the same row asserts honest counts (including a genuine `0`) — the guard is not a blanket suppression.

The flagged-agents column folds to `Not evaluated` because nothing in the gateway computes over-permission; `0 flagged agents` would be a clean bill of health the data cannot support.

Everything else is **deliberately left to its own ticket**.

---

## Type of Change

- [x] ✨ New feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [ ] No
- [x] Yes (describe below)

Internal-only, within `dashboard/`:

- `summarizeMatrix(agents, resources, verb)` → `summarizeMatrix(agents, resources, verb, cascade)`, returning `Certain<number>` fields instead of `number`.
- `CapabilitySummary` gains a required `cascade` prop. Required, not optional: a caller that cannot say where the verdicts came from has no business rendering a count of them.

No public API, wire format, or OpenAPI change. `pnpm generate:api` produces **zero** diff.

## Related Issues

- Related Jira ticket: [AAASM-5173](https://lightning-dust-mite.atlassian.net/browse/AAASM-5173)
- Related GitHub issues: N/A

## Testing

- [ ] No tests required

`pnpm test` — **236 files, 2090 tests, all passing**. Coverage on the new code (`--coverage.all=false`): `lib/truthfulness` **100%** statements/branches/functions/lines; `components/truthfulness` **100% / 97.6% branch**; `features/capability/{summary,api}.ts` and `CapabilitySummary.tsx` **100%**; `components/states` **100%**. Comfortably over the 80%-on-new-code gate, with no padding — every state in the vocabulary has its own branch case, and the empty/absent-cascade → never-Allow guard is asserted from both directions.

`pnpm type-check` clean · `pnpm lint` clean (`--max-warnings 0`) · `pnpm generate:api` → `git status --porcelain` **empty**.

Playwright (`tests/e2e/verify-aaasm-5173.spec.ts`, own config on port **4593**): **8/8 passing**, screenshots in `dashboard/verify/5173/` in light *and* dark.

Visual reach is limited to the states the wired surface can actually reach — `unconfigured`, `not-evaluated`, `unavailable`, `empty`, and known values. `unknown`, `not-supported` and `demo` have no production path on any wired surface today; they are covered by the unit suites plus a Storybook gallery (`Truthfulness.stories.tsx`), and will get in-app evidence from whichever page lane first renders them. Manufacturing a synthetic page purely to photograph them would be exactly the kind of fabricated evidence this lane exists to remove.

### Review round 2 — adversarial review findings

| # | Finding | Fix | Test that fails without it |
|---|---|---|---|
| B1 | `certainFromQuery` reported every success **and** pending as `unavailable` (TanStack sets `error: null`) | `69470c31` `hasError` rejects `null` too | 6 tests in `query.test.ts` / `cascadeEvidence.test.ts` |
| B1b | `error: loadError ?? undefined` hid B1 from production code | `e49d4add` pass the value through | `cascadeEvidence.test.ts` TanStack-shape cases |
| B2 | `trust: 50` dropped from fixtures → TS2741 on `main` once #1705 lands | `1f4966b8` restored on both fixtures | `tsc` under a simulated #1705 |
| M3 | Absent cascade hardcoded to `unavailable`, so a pending matrix claimed a failure | `4ec3b889` `propagateAbsence` | `propagates a pending matrix as unknown…` + the row-consistency render test |
| M4 | `tallyVerdicts` and `resolveVerdict` disagreed on an unevaluated cell | `28df3983` tally classifies via `resolveVerdict` | `tallyVerdicts agrees with resolveVerdict` (3 tests) |
| M6 | Missing `policies` key labelled `unavailable` | `4ec3b889` → `unknown` | `treats a null/undefined policy list as unknown…` |
| N9 | Constructors returned the wide union | `68bac6f9` return `KnownValue`/`AbsentValue` | n/a (type-level) |

Each fix was verified by reverting it and confirming the named test fails. Suite is now **2107 tests** (was 2090). Screenshots re-captured after the change and are **byte-identical**, so no rendered state moved.

Deliberately not fixed here, per scope: the `states.css` orphan, the Storybook-not-in-CI gap, and the page-level findings the review surfaced from these screenshots (Policies tab counters reading `All 0 / Active 0` during a 500, the sidebar `● runtime ok · 0 agents` during total failure, and `BudgetTree.tsx:140` swallowing a real `$0.00`). Those are page-lane tickets.

## Checklist

- [x] Code follows project style guidelines (`cargo fmt`, `cargo clippy`) — no Rust touched; JS/TS gates above
- [x] Commits are small and follow the Gitmoji convention (20 commits, bisectable)

---

> **Correction (post-review).** An earlier revision of this description and of the code comments called the empty-cascade grid "fully-permissive green". That is factually wrong: `.cap-mx-cell--allow` renders a neutral `--paper-2` background (`CapabilityMatrixGrid.css:177`); only `--deny` uses `--danger-bg`, and the frontend's missing-cell fallback is `na`, never `allow`. The finding itself is unchanged and undiluted — every cell resolves to `allow`, and that assertion is false. Only the colour claim was wrong; it is corrected throughout the branch in commit `e01055ff`, which also records that `decide()` mirrors `stage_capability` line for line.

## Notes for the page-level lanes

1. **Import the vocabulary; do not re-derive it.** `src/lib/truthfulness` and `src/components/truthfulness` are the contract.
2. **Never infer permission from missing policy data.** If you render a verdict, route it through `resolveVerdict` / `tallyVerdicts`.
3. Prefer `StatusState` over the deprecated `components/states` pair, and `TruthfulValue` over a bare `—` string.
4. **Known follow-ups this lane deliberately left alone:** `src/components/ErrorState.tsx` (the hi-fi port) hardcodes fabricated production facts — `503 service_unavailable`, `Stream halted at 14:02:47 UTC`, `last heartbeat 47s ago`, `auto-retry in 8s`, `severity: P1`. It is shared, page-visible copy and belongs to whichever lane owns those pages. `src/components/states/states.css` is now unreferenced and can be removed by a follow-up.

## Isolation

Touches `dashboard/` only. Zero files under `openapi/`, `dashboard/src/api/generated/`, `aa-api/src/models/{capability,topology}.rs`, `docs/src/adr/`, or any enforcement crate (`aa-runtime/ aa-gateway/ aa-proxy/ aa-ebpf* aa-core/ aa-sandbox/ aa-security/`).

**Correction to an earlier isolation claim.** This section previously said "no file that renders `trust` was modified". No *rendering* file was, but an earlier revision of this branch **removed `trust: 50` from a `CapabilityAgent` test fixture** — an edit unrelated to this ticket, and enough to break the merged product: PR #1705 makes `trust` required, both PRs merge cleanly and pass in isolation, and `main` then fails `Dashboard (type-check + lint)` with TS2741. Restored in commit `1f4966b8`, using `50` rather than `null` because today's `trust?: number` rejects null. Verified by locally simulating #1705's type change: with the fix, this branch's files produce zero errors under it.

[AAASM-5173]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5173)

**Verdict: APPROVE** ✅ — after one BLOCKER, one cross-PR BLOCKER and two MAJORs were found and fixed on this branch.

### The blocker — and why nothing else caught it

`certainFromQuery`, the module's own error-normalisation primitive, classified **every real TanStack query result as `unavailable`** — success and pending alike. TanStack's `QueryObserverSuccessResult` declares `error: null`, and the guard tested `outcome.error !== undefined`. `null !== undefined` is `true`, so the error arm won before the pending and data arms were reached.

It was **type-invisible** (`error?: unknown` accepts `null`) and the PR body advertised exactly the usage that broke it: *"a hook result passes straight in."* Every one of ~15 downstream lanes would have shipped a permanent fault-toned "the request for this value failed" over healthy data — the precise harm `query.ts` says it exists to prevent.

**CI was green. Sonar's gate was OK at 99.5% new-code coverage. Zero issues.** It survived all of that because `query.test.ts` never passed `error: null` — the success cases omitted the key entirely. **Coverage measured lines; nothing measured the contract.** The tell was already in the tree: `CapabilityPage.tsx:45` carried an undocumented `error: loadError ?? undefined` workaround that only existed because the advertised shape didn't work.

Fixed in `69470c31` by factoring a tested `hasError()` helper — one place to be wrong — with fixtures copied **field-for-field from the library interfaces**, not hand-trimmed. Reverting the null clause fails 6 tests, each named for the symptom.

| # | Finding | Sev | Resolution |
|---|---|---|---|
| B1 | every healthy query reported as failed | **BLOCKER** | `hasError()` helper + verbatim-TanStack fixtures; workaround removed |
| B2 | cross-PR type break with #1705 | **BLOCKER** | `trust: 50` restored on both fixtures; verified against a simulated #1705 |
| M3 | pending cascade rendered as a claimed failure ("Unavailable — Request in flight") | MAJOR | `propagateAbsence(cascade)` |
| M4 | `tallyVerdicts` and `resolveVerdict` gave opposite answers on unevaluated cells | MAJOR | tally now classifies **through** `resolveVerdict`, so they cannot drift |
| M5 | four converged call sites did change rendered output | MINOR | PR body corrected — precise, not softened |
| M6 | malformed 200 labelled `unavailable` | MINOR | → `unknown` |
| N9 | constructors returned the wide union | NIT | `KnownValue<T>` / `AbsentValue<T>` |

### Judgement calls worth recording

**M4 was not a mechanical fix.** Strict classification would have broken the wired surface: `project_matrix` emits a tool cell only for agents that declared it, so in a fleet with heterogeneous tools most cells are `undefined` and the summary would have read `Not evaluated` permanently. **Claiming to know nothing when we know almost everything is its own dishonesty.** Resolved by keeping the library strict and normalising the missing cell to `'na'` in `summary.ts`, where the endpoint's contract is actually known — restoring behaviour the grid already renders, not new inference. Flagged rather than assumed.

**One pre-existing test was changed, and the change was called out rather than buried:** `tallyVerdicts > counts only rule-backed verdicts…` used a shared fixture containing `undefined` and asserted counts over it — that expectation *was* the contradiction M4 fixes. The `undefined` moved to its own dedicated disqualification test. "A fix made an old test fail so I edited the old test" is the move that most deserves scrutiny, and it survived it.

### The contract (what page lanes consume)

Six non-interchangeable states — `unknown` · `unavailable` (the only one with `role="alert"`) · `unconfigured` · `not-evaluated` · `not-supported` · `demo`. `Certain<T> = {known:true,value:T} | {known:false,state,…}`; **`.value` is unreachable without narrowing** (verified against plain access, optional chaining, and destructuring — only an explicit cast or JSON round-trip escapes, which are universal TS holes). `certain()` treats `null`/`undefined`/`''` as missing but **`0`/`false`/`[]` as real**. **`demo` sits on the `known:false` side, so no aggregate can total a sample into a production figure.** Verdict precedence is strongest-first, and `deny`/`narrow`/`approval` survive an empty cascade untouched — a positive restriction can never be suppressed.

### Gates — stated exactly, not as "all green"

**16 success / 23 skipped / 0 failing / 0 pending.** All 23 skips verified gated on `needs.changes.outputs.rust == 'true'` or a strict subset filter; this is a dashboard-only PR, and `ci-success` treats skipped as pass. **This means the PR has had no Rust-side verification at all** — correct for its scope, worth stating plainly.

Note `ci.yml:1381-1383` marks coverage and Sonar **advisory by design ("must not block merge")**, and branch protection defines **no required status contexts** — the only platform gate is 1 approval + code-owner review.

Sonar gate **OK** · new coverage **99.5%** · duplication 0.0% · **0 open issues**. Tests **2107 passed** (up 17). Playwright re-run after the counting path changed: **8/8, screenshots byte-identical**. All 9 commits type-check independently.

### ⚠️ Merge order is not free

`git merge-tree` is clean with #1705 **in both directions** (identical tree) and both PRs are green in isolation — yet the merged product previously failed `tsc`. Git conflict-freedom was insufficient. B2 is fixed here, verified by patching `types.ts` to #1705's shape: **zero errors in this PR's files**; the only 3 remaining belong to files #1705 owns.

**Whichever of #1705/#1707 merges second should still rebase and re-run `type-check`** before merge — a stale green check from before the rebase would hide a recurrence.

### PRODUCT-DECISION-REQUIRED (not a blocker; for the record)

This PR implements ADR 0024's "explicit-unconfigured" option, and in doing so **resolves one of 0024's own open questions** — whether "unconfigured" is a single product-wide concept or per-surface. It makes it product-wide. 0024 is `Proposed` on unmerged #1706. **The behaviour shipped here is the honest one; the finding is authorisation sequencing, not correctness.**

### Out-of-scope follow-ups (evidenced by this PR's own screenshots)
PoliciesPage renders **"All 0 / Active 0"** while the request 500s · the sidebar asserts **"● runtime ok · 0 agents"** during total failure · `BudgetTree.tsx:140` swallows a real `$0.00` as absence (the inverse bug) · `states.css` now orphaned · Storybook is never built in CI, so the gallery "coverage" for the three unreachable states is ungated.

### Disclosed by the author, recorded here
A `git checkout --` inside a verification script destroyed six uncommitted edits; caught immediately, re-applied, verified by marker-grep and type-check. Nothing committed or pushed was affected. Two first-round commit messages still carry a disproven "green grid" phrasing — left deliberately, since correcting them needs a force-push, and the retraction commit follows them in the log.

Reviewed by Claude Code. Every finding was verified against the source directly, not taken from the implementing agent's report.

PR #1710, opened 2026-07-26
[AAASM-5110] 🐛 (dashboard): Stop Identity rendering fabricated agents, grants and access-log IPs

Chisanan232:
## Description

The Identity surface presented invented identity, permission and network data as
production fact on two tabs. Both are release-blocking; this PR fixes the lane.

### AAASM-5110 — Roles & Permissions tab

**Was:** `features/iam/agents.ts` hardcoded four fictional agents
(`support-agent/cx`, `code-review/platform`, `data-analyst/analytics`,
`deploy-agent/devops`) plus a `SEED_PERMISSIONS` grant table attributing
capabilities to policies that do not exist (`support-agent-policy-v2`,
`deploy-agent-policy-v1`) and roles nothing issues (`agent.operator`,
`agent.readonly`). Neither fetcher touched the network. Both rendered behind a
full loading/error/empty apparatus with no disclaimer — and the sibling Roles
card *does* carry a caveat banner, so the fiction read as the more authoritative
half of the same tab.

**Now — option 1, wired to real endpoints.** Both exist and are cited from
`openapi/v1.yaml`:

| Surface | Endpoint | Evidence |
|---|---|---|
| Agent registry | `GET /api/v1/agents` → `PaginatedAgentResponse` | `openapi/v1.yaml:102`; already consumed by Fleet (`features/agents/api.ts:22`) |
| Permission cascade | `GET /api/v1/agents/{id}/capabilities` → `EffectivePermissionsResponse` | `openapi/v1.yaml:214`; its own summary names it *"the dashboard's inherited-permissions panel"* |

Everything the endpoints do not carry is now an explicit absence from the shared
`lib/truthfulness` vocabulary, never back-filled:

- **Owner team** → `not-supported`. `AgentResponse` has no team field at all, so
  waiting will not help. A test also pins that a team-shaped key in the free-form
  `metadata` bag is *not* promoted into it.
- **Grant timestamp** → `not-supported`. `PermissionSourceResponse` carries
  `scope`/`allow`/`deny` and nothing else; the cascade records *what* a scope
  contributes, never *when*.
- **`last_seen`** → `unknown` when `last_event` is null (a freshly registered
  agent has not reported yet), known otherwise.
- **Status** → rendered verbatim. `AgentResponse.status` is an open `string`; the
  old closed `online | offline | degraded` union had no producer and existed only
  to type the four hardcoded agents.

**The empty-cascade case (AAASM-5106).** An empty `sources` means the gateway
resolved no policy document, so the empty `allow`/`deny` alongside it is the
*absence of an evaluation*, not a finding. It renders `unconfigured` via the
shared `cascadeIsEmpty`, never the old "This agent has no effective permissions".
A body-less 200 is failed rather than allowed to impersonate that state.

The invented `team → role → policy` inheritance vocabulary went with the seed —
it described the shape of the fixture, not of the system. The panel now shows the
gateway's real cascade scopes (`global`, `team:platform`, …).
`groupBySourceKind.ts` had no remaining caller and is removed.

### AAASM-5111 — Access Log tab

**Was:** `features/iam/accessLog.ts` seeded ten security events attributed to
named identities (`alice@agent-assembly.dev`, `gateway-ci`) from invented source
IPs (`10.0.0.42`, `10.0.0.99`), including **failed** logins. `isoMinusHours()`
re-based every timestamp against a module-load `new Date()`, so the feed always
looked current. It was filterable and paginated exactly like a real one; the only
acknowledgement was an internal code comment. This is the surface an operator
opens during an incident review.

**Now — option 2, `not-supported` with the action disabled.** No endpoint
replaces it, and I did not fake one. `GET /api/v1/logs` (`openapi/v1.yaml:2247`)
is the per-agent governance log: `LogEntry` carries `agent_id`, `session_id`,
`event_type`, `seq`, `timestamp` and an opaque `payload` — **no** signing-in
identity, **no** source address, **no** success/failure outcome, and no
`login`/`logout`/`member_invite` concept. Wiring the panel to it would mean
synthesising every column the tab is about, which is the defect rather than the
fix.

Option 3 (`demo`) was deliberately rejected here: a badged fabricated *failed
login from a fabricated IP* is still a fabricated failed login on an
incident-review surface.

`not-supported` rather than `unavailable` because nothing was requested and
nothing failed — telling an operator mid-incident that an audit source exists but
is merely down would be its own wrong claim. The state names **AAASM-5176**
(agent identity issuance — without issued identities there is nothing to
attribute an event *to*) and **AAASM-5177** (API-token expiry lifecycle, which
owns the token-rotation events this tab claimed) and links to the real governance
audit log.

The tab itself stays — the 4-tab Identity model is ratified in **ADR-0017 item
19**, so removing it would be a product decision, not a bug fix. The filter bar
stays too, **disabled**: filtering is an action with no successful production
path, and a live filter over an empty surface lets an operator read "no events
matched your filter" out of "there is no audit source at all".

The **Service Identities** tab was not touched.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

No public API changes. `openapi/v1.yaml` and `src/api/generated/**` are
untouched — `pnpm generate:api` produces a clean tree.

## Related Issues

- Related Jira tickets:
  [AAASM-5110](https://lightning-dust-mite.atlassian.net/browse/AAASM-5110),
  [AAASM-5111](https://lightning-dust-mite.atlassian.net/browse/AAASM-5111)
- Backend gaps referenced, not faked:
  [AAASM-5176](https://lightning-dust-mite.atlassian.net/browse/AAASM-5176),
  [AAASM-5177](https://lightning-dust-mite.atlassian.net/browse/AAASM-5177)
- Vocabulary consumed:
  [AAASM-5173](https://lightning-dust-mite.atlassian.net/browse/AAASM-5173);
  empty-cascade semantics:
  [AAASM-5106](https://lightning-dust-mite.atlassian.net/browse/AAASM-5106)

## Testing

**Vitest — 237 files / 2142 tests passing.** New/rewritten branch coverage for
every state introduced:

- `agents.test.tsx` (new) — the projections, including that the hooks reach the
  real endpoints at all, that `metadata.team` is not promoted into `owner_team`,
  and that a body-less 200 cannot impersonate the empty cascade.
- `RolesPermissionsPanel.test.tsx` — rewritten; asserts none of the four seeded
  agents, none of their teams and none of the four invented grant sources render,
  plus the `unavailable` / `unconfigured` / known branches.
- `accessLog.test.tsx` — rewritten; pins that the module exports no seed, store,
  fetcher or test seam, and that **nothing it exports changes with the clock**
  (the property that made the old feed always look current).
- `AccessLogPanel.test.tsx` — rewritten; asserts no row, no verdict, none of the
  six seeded identities, and **no IPv4-shaped text at all**, so a *freshly*
  invented address fails too.
- `AccessLogFilterBar.test.tsx` — added the disabled branch; the default stays
  enabled so existing call sites are unaffected.

**Playwright — `review-aaasm-5110.spec.ts` on port 4582, 8/8 passing** (4 checks
× light and dark), console errors `[]` in every case. Screenshots committed to
`dashboard/verify/5110/`.

`iam-access-log.spec.ts` is rewritten against the honest render. Three
**pre-existing** breakages in it are fixed alongside, because none of its
assertions were reachable past the first line: the token was seeded into
`localStorage` after AAASM-4322 moved it to `sessionStorage`; every deep
`page.goto('/identity?...')` got a 404 from `vite preview` instead of the SPA
shell; and the shell's unstubbed probes left the page never settling. It now
passes 5/5.

## Checklist

- [x] Code follows project style guidelines (`pnpm lint` — *ESLint: No issues found*)
- [x] Documentation updated if behaviour changed (module-level WHY comments on
      every rewritten module record what was fabricated and why the replacement
      state was chosen)

**Gates run locally, each once:** `pnpm test` ✅ 2142 passed · `pnpm type-check`
✅ clean · `pnpm lint` ✅ no issues · `pnpm generate:api` → `git status
--porcelain` ✅ empty.

[AAASM-5110]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

### MAJOR 1 — `AGENT_STATUS_TONES` had no producer on this endpoint (fixed, `8ce11db7`)

Confirmed against source. The lowercase `AgentStatus` enum in the schema belongs
to `AgentNode` in the capability-matrix projection
(`aa-api/src/models/capability.rs:95`), **not** to `AgentResponse`.
`GET /api/v1/agents` builds the field as `format!("{:?}", r.status)`
(`aa-api/src/routes/agents.rs:135`) over `aa_gateway::registry::AgentStatus`
(`aa-gateway/src/registry/mod.rs:77-83`), so the wire values are Rust `Debug`
renderings: `Active`, `Deregistered`, `Suspended(<reason>)` — reason being
`BudgetExceeded`, `Manual`, `ParentDeregistered`, or the struct-shaped
`ParentSuspended { parent_agent_id: [..] }`.

The replacement had reproduced the very defect the ticket removed: a vocabulary
with no producer, now carrying a comment asserting it had one. Every production
status fell through to `--other` and all three tone classes were dead.

Tone is now derived from the **outer variant** (`agentStatusVariant`), because
`Suspended` carries a payload and exact-match could never hit it. The text still
renders **verbatim** — the payload is operationally load-bearing (`BudgetExceeded`
is auto-resumable, `Manual` is not), and relabelling it would invent a vocabulary
the backend does not define. See the open product question below on legibility.

### MAJOR 2 — merged verdict computed by the backend and discarded (fixed, `a449d5a0`)

`get_agent_capabilities` (`aa-api/src/routes/agents.rs:634-650`) returns the
most-restrictive-wins merge *and* the per-scope sources; the panel rendered only
the sources. The merged verdict now leads the panel under **Effective (merged)**,
with **Contributing scopes** below it. On the review fixture (`global` allows
`secrets.read`, `team:platform` denies it) the merged section shows a single
`secrets.read → DENY`, so the operator no longer resolves the conflict by eye.

A resolved cascade that constrains nothing gets a plain note, not an absence
badge — unlike the empty-cascade case, it *did* evaluate.

### MINOR — `unavailable` missing from visual evidence (fixed, `9d512091`)

Both branches now have e2e coverage and committed captures
(`roles-registry-unavailable-{light,dark}.png`,
`roles-cascade-unavailable-{light,dark}.png`). They allow for the app's default
TanStack retry policy: a *slow* failure must still land on `unavailable` rather
than on a benign empty state while retries are in flight.

### NIT — blanket `**/api/**` stub (fixed, `9d512091`)

The catch-all now records what it absorbs, and the run asserts no IAM/agents
request reached it. This immediately caught `/api/v1/iam/api-keys` being silently
answered with `{}`; it now has an explicit fixture.

### Fixtures

Both unit and e2e fixtures moved to production-shaped values — `'Active'` and
`'Suspended(Manual)'` — plus `'Quarantined'` for the unknown-variant fall-through
and `'Suspended(ParentSuspended { parent_agent_id: [1, 2, 3] })'` in the parser
table. A fixture describing a response the gateway cannot produce is how the
original defect survived review.

### Open — team ownership is obtainable elsewhere (PRODUCT DECISION, not decided here)

`owner_team` renders `not-supported`, correct for `GET /api/v1/agents` and scoped
to it in the detail string. But `AgentTree.team_id` and `AgentNode.owner`
(`openapi/v1.yaml:4770`, deriving from `team_id` falling back to `org_id`) do
carry it, so `not-supported` may tell an operator to stop looking for something
another endpoint has. **Left as-is pending a product call** on whether the panel
should fetch from the other endpoint.

### Gates re-run (each once)

```
pnpm test        → Test Files 237 passed (237) / Tests 2156 passed (2156)
pnpm type-check  → tsc --noEmit   (no output)
```

Playwright review pass re-run: **12/12** (was 8/8), console errors `[]` in every
case. `iam-access-log.spec.ts`: 5/5.

### Review round 3 — Sonar issue cleared (`c920a4bb`)

`typescript:S6478` on `AgentRegistryList.tsx:119` — the inline
`format={(status) => <StatusChip status={status} />}`. **Fixed by hoisting**, not
suppressed: `renderStatusChip` now lives at module scope, so the callback keeps
one identity for the lifetime of the module instead of being rebuilt on every
render.

Worth recording the accurate reason, since the rule's usual justification does
not apply here: `TruthfulValue` **invokes** `format` (`format(value.value)`,
`TruthfulValue.tsx:89`) rather than rendering it as a component type, so the
element type in the tree was already the stable module-scope `StatusChip` — no
unmount/remount was occurring and no local state was at risk. The rule still
fires on the shape, and hoisting is the correct shape regardless, so it was
fixed rather than argued with.

Verified as a pure structural move by rendering the list before and after and
diffing the table's `outerHTML` — **byte-identical** (2416 bytes both sides).
Playwright therefore not re-run.

**Sonar now reports 0 open issues**; gate OK, new coverage 98.7%, duplication
0.0%, all new ratings A, hotspots 100% reviewed.

```
pnpm test        → Test Files 237 passed (237) / Tests 2156 passed (2156)
pnpm type-check  → tsc --noEmit   (no output)
```


Chisanan232:
## Claude Code — adversarial review (AAASM-5110 · 5111)

**Verdict: APPROVE** ✅ — after two MAJORs and one static-analysis issue were found and fixed.

### What was removed
4 invented agents, a `SEED_PERMISSIONS` grant table attributing grants to policies that do not exist, 10 fabricated access events with invented source IPs and **timestamps re-based at module load so they always looked current** — on the surface an operator opens during incident review.

### The two resolutions, both better than the ticket anticipated
**5110 → wired to real endpoints.** `GET /api/v1/agents` and `GET /api/v1/agents/{id}/capabilities` both exist, and the latter's own OpenAPI summary names it *"the dashboard's inherited-permissions panel"* — so the `unavailable` state the ticket suggested would have **understated what is actually available**. Note this only became viable because **AAASM-5102 merged earlier this session**; before that fix the endpoint silently dropped every Org- and Team-scoped policy. Empty cascade → **`unconfigured`**, never "no effective permissions" (AAASM-5106).

**5111 → `not-supported`, filter bar kept but disabled.** No endpoint exists — `GET /api/v1/logs` is per-agent governance audit with no identity, IP, outcome or login concept. The `demo` option was **deliberately rejected**: *"a badged fabricated failed login from a fabricated IP is still one on an incident-review surface."* Keeping the filter bar disabled means an empty surface can't be misread as "no events matched".

### MAJOR 1: one no-producer vocabulary swapped for another
`AGENT_STATUS_TONES` was documented as *"words the registry is known to emit"*. True observation, false conclusion — that `AgentStatus` enum belongs to `AgentNode` in the **capability-matrix** projection, not to `AgentResponse`. `aa-api/src/routes/agents.rs:135` emits `format!("{:?}", r.status)` — the Rust **Debug** rendering of `Active | Suspended(SuspendReason) | Deregistered`. So every production status fell through, all three tone classes were dead, and operators saw the raw string `Suspended(Manual)`.

**The tell was in the PR's own evidence**: the screenshot showed a green ACTIVE and muted IDLE — a rendering production cannot produce — because both fixtures used lowercase values. **A fixture that can't occur in production is how this survived review.**

The fix went further than the finding: it pulled `SuspendReason` and found a **struct-shaped** variant, so the worst case is `Suspended(ParentSuspended { parent_agent_id: [...] })`. Tone is now parsed from the text before the first `(` — Debug's own grammar — and the string **renders verbatim**, because the payload is load-bearing governance signal (`BudgetExceeded` is auto-resumable, `Manual` operator-only). Trimming to a bare `Suspended` would lose signal; relabelling would invent a vocabulary. **Only tone is derived.** CSS wrapping keeps a payload from breaking the column.

### MAJOR 2: the merged verdict was computed and discarded
`get_agent_capabilities` returns **both** `merged_allow`/`merged_deny` (most-restrictive-wins) **and** per-scope `sources`; the panel rendered only `sources`. On a cascade where `global` allows `secrets.read` and `team:platform` denies it, the operator saw two chips in separate sections and had to perform the merge the endpoint had already done. Now an **Effective (merged)** section leads, ahead of Contributing scopes — and the review fixture was changed so the scopes genuinely **disagree**, without which the merge is unobservable.

### Verified
Every other fabrication replacement is genuinely knowable · **all vocabulary choices correct**, including `unconfigured` for the empty cascade and `not-supported` for a grant timestamp `PermissionSourceResponse` does not carry · `groupBySourceKind.ts` deletion is clean — zero callers, and the shape it grouped was produced only by the deleted seed · **test integrity intact**, with the two longest-reach assertions preserved byte-identical · the `iam-access-log.spec.ts` edit was unavoidable (it asserted the deleted seed rows **by name**), and its "already red on `main`" claim holds for at least one reason I confirmed: `tokenStorage.ts` reads from **sessionStorage** since AAASM-4322 while the spec seeded `localStorage`.

The Playwright guard is stronger than asked — it asserts **no IPv4-shaped text renders at all**, so a *fresh* invention fails too. Its `unavailable` captures use a 20 s timeout deliberately: the app's default 3-retry backoff means **a slow failure must still land on `unavailable`**, not rest on a benign empty state. And the new catch-all instrumentation **immediately caught a genuinely absorbed request** — `/api/v1/iam/api-keys` was being silently answered `{}`; now explicitly fixtured.

### Static analysis: 1 → 0, and it corrected my reasoning
I justified hoisting the nested component by the remount hazard. **That did not apply** — `TruthfulValue` *invokes* the callback rather than rendering it as a component type, so the reconciled element type was already the stable module-scope `StatusChip`; nothing was unmounting. The fix is still right (stable callback identity, load-bearing once `TruthfulValue` is memoized) but my stated benefit was wrong. It fixed it anyway and **verified the move by diffing the table's `outerHTML` — byte-identical, 2416 bytes both sides** — checking "output unchanged" directly rather than inferring it from passing tests. **Zero suppressions.**

### Gates
CI **16 pass / 23 skipped / 0 fail** · Sonar gate **OK**, **0 open issues**, new coverage **98.7%** · 2156 tests · Playwright 12/12 review pass + 5/5 on the repaired spec.
The 23 skips are Rust jobs excluded by the paths router (dashboard-only), so **no Rust-side verification**; **no CI job runs Playwright**, so the e2e evidence is author-run, not a gate.

### Two product questions raised, not decided
1. `owner_team` renders `not-supported` — correct for `GET /api/v1/agents`, but `AgentTree.team_id` and `AgentNode.owner` do carry it, so the state may tell an operator to stop looking for something another endpoint has.
2. `Suspended(ParentSuspended { parent_agent_id: [...] })` is honest but barely legible. Making it readable needs a backend change (serialize a structured status instead of `Debug`) or a product call on a label vocabulary. Neither was invented.

Reviewed by Claude Code. The Debug-rendering chain and both endpoint contracts were verified against the Rust source directly.

PR #1712, opened 2026-07-26
[AAASM-5150] 🐛 (dashboard): Stop an alerts outage rendering as "no alerts"

Chisanan232:
## Description

The Alerts surface asserted several things it did not know. This lane makes each of those claims either true or visibly absent, using the shared truthfulness vocabulary landed by AAASM-5173 (`src/lib/truthfulness`, `src/components/truthfulness`) — no status semantics are re-declared here.

The headline defect: **a failed alert-rules query rendered as "No alerts in this window"** while alerts were firing. The category chips join each alert's `ruleId` against the rules list, so an outage dropped every alert to `uncategorized`, all four chips read `0`, and a selected chip emptied the feed into the empty state — with no banner and no retry.

### Per ticket

| Ticket | Before | After |
|---|---|---|
| **AAASM-5150** (release-blocking) | Rules-query outage → chips read `0`, selected chip empties the feed → `EmptyStateNoAlerts` "No alerts in this window". `alertsQuery.isError` had a banner; `rulesQuery.isError` had nothing. | Rules failure gets its own named banner (`Failed to load alert rules: …`) with a retry; category counts fold to the shared absence marker instead of zeroes and selection is disabled; the category filter falls back to `all` so the feed is never emptied by an underivable join. A failed **alerts** query now renders `StatusState state="unavailable"` instead of any empty state, and neither the row count nor any stat tile reports a number it does not have. |
| **AAASM-5123** | Silent 50-row truncation; every stat tile and the "N alerts" label derived from the truncated page. | The paginated envelope is read whole (`useAlertsPageQuery`). A page short of `total` renders a truncation notice ("Showing the first 50 of 214 alerts") and the strip carries a scope line ("Counts cover the 50 alerts on this page, not all 214."). The row count describes **only the page** — "7 of 50 alerts on this page" — so its numerator and denominator are the same population; the fleet total is stated once, by the truncation notice. A **missing** `total` stays `null` — never defaulted to the page length — and both the scope line and the page qualifier treat unknown as not-proven-complete. |
| **AAASM-5122** | `buildAlertsQueryString` sent `severity`/`status`/`agent`/`range`/`from`/`to`; the handler reads `Query<PaginationParams>` only, so they were silently dropped and every chip returned an identical list. | Those parameters are no longer sent and are no longer part of the cache key (the query stopped thrashing on chip clicks). Narrowing happens over the loaded page via `applyClientFilters`, which now also resolves and applies the time range. An unparseable custom bound resolves to *unbounded*, never to an empty window. |
| **AAASM-5121** | No resolve mutation, no path in `endpoints.ts`, no UI. `useAlertsStream` consumed `resolve` events the dashboard could never cause. | `useResolveAlertMutation` + a per-row `ResolveAction` in the detail drawer, gated on `canWrite`, invalidating `alertsQueryKeys.alerts`. The mock's header-level "Acknowledge all" is deliberately **not** built — there is no bulk endpoint, and fanning out N writes from a button labelled as one atomic action would report success it cannot guarantee. |
| **AAASM-5147** | `canWrite` appeared once on the whole surface (the page's "New rule"). `grep canWrite src/features/alerts/` returned nothing. | Threaded into `AlertRulesTable` (New rule / Edit / Delete), `SilenceAction`, `ResolveAction`, every mutating control in `DestinationManager`, **the zero-rule `EmptyStateNoRules` CTA, and `AlertRuleForm`'s own submit** — each with `disabled` + `WRITE_REQUIRED_HINT`. RBAC test extended from 2 cases to 9 (read / write / admin) plus 3 form-level cases. |
| **AAASM-5149** | **Partially delivered — see "Not completed" below.** | The corrected predicate ships as a tested selector (`criticalFiringCount` / `criticalFiringBadge` in `features/alerts/alertBadge.ts`), plus the `features/alerts` half of what the badge is *fed*: `useAlertsQuery` no longer sends parameters the server drops. |

### Endpoint evidence

Both claims were verified against `openapi/v1.yaml` before wiring anything:

- **`resolve` exists.** `openapi/v1.yaml:1020` — `/api/v1/alerts/{id}/resolve`, `post`, `operationId: resolve_alert`, described as idempotent, `requestBody.required: true` with schema `ResolveAlertRequest { reason?: string | null }`. The client therefore always sends a body (`{"reason": null}` when none is given) rather than omitting it.
- **The filter parameters do not exist.** `list_alerts` (`/api/v1/alerts`, `get`) declares exactly two parameters: `page` and `per_page` (max 100, default 50). There is no `severity`, `status`, `agent`, `range`, `from` or `to`. The response is `PaginatedAlertResponse { items, page, per_page, total }`.

### Incidental fix inside the lane

`applyFire`/`applyResolve` reached for every cache under the bare `alerts` key with `setQueriesData`, which also matched the single-alert **detail** caches — a list-shaped updater over an `AlertDetail` object. The list and detail caches are now keyed distinctly (`ALERTS_LIST_KEY` / `alertDetailKey`), which was a prerequisite for moving the list cache to the envelope shape. A regression test covers it.

`useAlertsQuery` keeps its `readonly Alert[]` contract (via `select`) so `AppShell.tsx` and `OverviewPage.tsx` — owned by other lanes — are untouched and share one cache entry with the Alerts page.

### Not completed within file ownership

- **AAASM-5149's shell wiring.** The missing status predicate lives at `dashboard/src/components/AppShell.tsx:133`, a file contended by two tickets outside this wave. `AppShell.tsx` is **not** touched. The selector, its rationale and its tests are in place and green.

  **When it is wired, use `criticalFiringBadge(certainFromQuery(alertsQuery))`, not `criticalFiringCount(alerts.data ?? [])`.** The `?? []` form is the fail-open pattern this lane exists to remove: it converts a failed alerts request into "0 critical alerts", which the shell renders as *no badge*, which reads as "nothing critical is happening". That guidance is recorded in `alertBadge.ts`'s module comment so it travels with the code.
- **True server-side alert filtering** would need a backend change to `list_alerts` and has no owning ticket; the client-side narrowing restores correct behaviour today and the page states its scope.

### Known consequence for the pre-existing-red alerts e2e specs

`tests/e2e/alerts.spec.ts`, `alerts-5026-verify.spec.ts` and `alerts-design-fidelity.spec.ts` are **11/11 red on `main` already** (verified by stashing, rebuilding and re-running: identical `expected: 0, unexpected: 11` with and without this branch — they fail on `alert-rules-tab` never appearing). This PR does not regress them, but it does add a **second** reason they will fail: all three seed alerts with hardcoded `firstFiredAt: '2026-05-14T…'`, and the newly-wired client-side 24h default window now filters every one of those rows out. Whoever fixes their pre-existing failure will hit this too — the fix is to make those fixtures relative to the run (as `review-aaasm-5150.spec.ts` does). Left untouched here: they are outside this lane's file ownership.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [x] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

No public API changes. `useAlertsQuery(filters)` keeps its signature and return type; the `filters` argument is now documented as accepted-but-not-sent, because the API never honoured it.

## Related Issues

- Related Jira tickets:
  - [AAASM-5150](https://lightning-dust-mite.atlassian.net/browse/AAASM-5150) — a failed alert-rules query silently empties the Alerts category filter
  - [AAASM-5123](https://lightning-dust-mite.atlassian.net/browse/AAASM-5123) — Alerts list silently truncates at 50 and derives every stat tile from that window
  - [AAASM-5122](https://lightning-dust-mite.atlassian.net/browse/AAASM-5122) — Alerts filters are silent no-ops
  - [AAASM-5121](https://lightning-dust-mite.atlassian.net/browse/AAASM-5121) — no Acknowledge/Resolve action though the endpoint exists
  - [AAASM-5147](https://lightning-dust-mite.atlassian.net/browse/AAASM-5147) — Alerts RBAC gating covers only the "New rule" button
  - [AAASM-5149](https://lightning-dust-mite.atlassian.net/browse/AAASM-5149) — nav alert badge counts resolved alerts (partial — see above)
- Related GitHub issues: N/A

## Testing

The headline regression test asserts a **failed rules query renders `unavailable`, not an empty-state "no alerts"** — in both vitest and Playwright, with the failure injected at the network layer rather than assumed.

Every fix in this PR was verified by reverting it and confirming the guarding test goes red — including the zero-rule RBAC case, which the previous RBAC suite could not reach because it seeded a rule in every scenario.

```
pnpm test        → Test Files 240 passed (240) | Tests 2189 passed (2189)
pnpm type-check  → tsc --noEmit, exit 0
pnpm generate:api → git status --porcelain empty (no codegen drift)
```

Playwright review pass (repo-local `@playwright/test`, own config on port 4586, light **and** dark, console errors asserted `[]`):

```
14 passed (36.0s)   — every scenario in BOTH light and dark
  ✓ a rules outage never reads as "no alerts"
  ✓ an alerts outage renders unavailable, never zero
  ✓ a truncated page is never presented as the whole fleet
  ✓ the filter controls narrow a real feed and send no dropped parameters
  ✓ the resolve affordance posts to the shipped endpoint
  ✓ a read-scope caller sees every write affordance disabled
  ✓ a read-scope caller cannot reach the rule form from a zero-rule install
```

Screenshots committed to `dashboard/verify/5150/`.

**Pre-existing red suites, unchanged by this PR:** `tests/e2e/alerts.spec.ts`, `alerts-5026-verify.spec.ts` and `alerts-design-fidelity.spec.ts` fail 11/11 both with and without this branch (verified by stashing, rebuilding and re-running). They are not regressed here and are out of scope.

### Static-analysis pass

SonarCloud initially reported 5 open issues on this PR; all 5 are fixed, none suppressed (no `NOSONAR`, no rule disables) — the recount is **0**.

- `S3735` (`void` operator) — the discard was on an unused *parameter*, not a floating promise, and was load-bearing: deleting it fails `@typescript-eslint/no-unused-vars`. Replaced with a TypeScript overload whose implementation binds no parameter, so the argument stays in the public signature for the two call sites this lane does not own. This codebase already avoids `void` — see `src/lib/ignorePromise.ts`.
- `S3358` ×2 (nested ternary) — the submit-cursor expressions in `AlertRuleForm` and `SilenceAction` became named `submitCursor` if/else blocks, mirroring the `submitLabel` block already in that file. The `disabled={!canWrite || …}` gates are unchanged.
- `S6819` (`role="status"`) — the flagged element is this page's own truncation notice, now `<output>`, matching the stream banner already above it. The shared `StatusState` primitive from AAASM-5173 computes its own role and is **not** touched.
- `S3776` (cognitive complexity 20 > 15) — three cohesive extractions, no behaviour change: `AlertsFeedBody` (the five mutually-exclusive feed surfaces, whose ordering *is* the 5150 fix and now carries that rationale in one named place), `alertsCountLabel` (into `alertsCoverage`, beside the page-vs-fleet rule it enforces, now unit-testable without rendering), and `toggleFilterValue` (shared by the filter chips and the stats tiles).

## Design references

- `design/v2/hi-fi/alerts.jsx` (authoritative per ADR 0025) — the per-row Acknowledge at `:86-88` is what `ResolveAction` implements; the `filtered.length / all.length` counter at `:201-207` is what the truncation notice makes honest over a paginated feed.
- ADR 0017 (`docs/src/adr/0017-dashboard-design-parity-ratified-evolutions.md`) ratifies exactly one Alerts evolution — the live `useAlertsStream` — which is preserved untouched. No ratified evolution is reverted.

## Checklist

- [x] Code follows project style guidelines (frontend lane: `pnpm lint`, `pnpm type-check`)
- [x] Documentation updated if behaviour changed (module-level WHY comments on every changed module)
- [x] All CI checks passing — CI Success green (dashboard type-check + lint, build, tests + coverage, codegen drift, aa-cli embed compat); SonarCloud quality gate **OK**: new coverage 96.1% (threshold 80), new duplication 0.0%, reliability/security/maintainability all A, security hotspots 100% reviewed — pending
- [ ] Commits are signed off (`git commit -s`) — optional/advisory here

[AAASM-5150]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5150 · 5123 · 5122 · 5121 · 5147 · 5149)

**Verdict: APPROVE** ✅ — after two MAJORs and 5 static-analysis issues were found and fixed. **5149 remains partial by instruction** (see below).

### Endpoint questions answered with spec evidence, not assumption
- **`resolve` genuinely exists** — `openapi/v1.yaml:1020`, `operationId: resolve_alert`, and critically **`requestBody.required: true`**, so the client always sends `{"reason":null}` rather than omitting the body. That detail is what would otherwise ship a 400 in production.
- **The filter params do not exist** — `list_alerts` declares only `page`/`per_page`. So 5122 was fixed by **removing** them from the request *and from the cache key* — the query had been thrashing on every chip click, refetching identical data.

### MAJOR 1: the write gate was bypassable
The header "New rule" was correctly disabled, but `EmptyStateNoRules`'s "Create your first rule" CTA calls the *same* handler with **no `canWrite` and no `disabled`**, and `AlertRuleForm`'s submit gated on `submitting` only. On a **zero-rule install** a read-scope caller reached the form, submitted, and got the raw `403` toast the ticket describes.

`AlertsPage.rbac.test.tsx` seeded `data: [RULE]` in **every** case, so no test could enter that branch — the defect was structurally untestable.

The fix generalised past what I asked. It gated the **form** as well as the CTA: *"the form is the last gate before the mutation, so it must hold regardless of which control opened it (deep link, ungated CTA, future caller) — **gating only entry points is what produced this bug**."* And it fixed the **root cause of the untestability** — `renderWithScopes` now takes `rules` as a parameter, with three cases passing `[]`, and one new test asserts **the absence of the POST**, not merely a disabled attribute.

### MAJOR 2: `"N of TOTAL alerts"` mixed two populations
The numerator was the client-filtered page; the denominator was the whole fleet. Filter to CRITICAL on a truncated page and the label read **"7 of 214 alerts"** — a ratio over a population never queried — two lines below a banner saying "the first 50 of 214". Both tests exercised truncation *without* a filter, where the numbers coincide.

Now a four-state `countLabel()` driven by `coversWholeFleet` rather than `truncated`, so an **unknown** total also earns the page qualifier — *"unknown is not coverage."* `total` appears only in the truncation notice.

### Verified
`total` handling is right — `finiteOrNull` → `unknown`, **no `?? items.length` path survives** anywhere · `applyFire` keeps `null` on increment ("not knowing the total plus one is still not knowing it") · category chips `propagateAbsence` rather than reading `0` · a **pending** query renders `unknown`, never `unavailable`, through all three TanStack retries · **"Acknowledge all" deliberately not built** — no bulk endpoint, and *"fanning out N writes behind a one-action label would report success it can't guarantee."*

**An incidental bug found and fixed**: `applyFire`/`applyResolve` called `setQueriesData` on the bare `alerts` prefix, which **also matched single-alert detail caches** — running a list updater over an `AlertDetail`. Keys split, regression test added. Unrelated to any ticket.

**Test integrity: clean.** The four `applyClientFilters` tests were relocated **verbatim** alongside 7 new ones; the URL assertion was replaced by a **stronger** pair; `alertsStreamSync` gained the detail-cache guard. **Zero** added `eslint-disable`/`@ts-expect-error`/`as any`. A `pnpm lint` run that returned "0 errors, 2 warnings" was treated as the failure it is under `--max-warnings 0`, and fixed structurally.

### Static analysis: 5 → 0, none by suppression
The `void` operator was **not** a floating promise — it discarded an unused *parameter*. Deleting it would have traded a Sonar CRITICAL for an ESLint error (verified by doing it). Fixed with a TypeScript **overload** whose implementation binds no parameter, so `AppShell.tsx` and `OverviewPage.tsx` keep compiling untouched. Repo precedent supports this: `src/lib/ignorePromise.ts` exists specifically to avoid `void` here.

`S3776` was resolved by three cohesive extractions — including `AlertsFeedBody.tsx`, which now owns the precedence between the five feed surfaces. That ordering (outage → no-rules → empty → rows) **is** the 5150 fix, so giving it a named home with the rationale documented is better than the inline chain it replaced. The subtlety that a **pending** request must not take the unavailable branch is preserved.

### 5149 — partial, and it leaves nothing worse
`AppShell.tsx` is untouched (contended by two cosmetic tickets outside this wave), the badge behaves exactly as on `main`, and although the cache key changed, the server ignored those params anyway — so the badge's input data is byte-identical. **To finish: `AppShell.tsx:133` → `criticalFiringBadge(certainFromQuery(alertsQuery))`.**

**Not** `criticalFiringCount(alerts.data ?? [])`, which is what I originally suggested — that reintroduces the `?? []` fail-open this lane exists to remove, turning a failed request into "0 critical" → no badge → *"nothing critical is happening."* The author caught my error and recorded the correct wiring in the module comment so it travels with the code.

### Gates
CI **14+ pass / 23 skipped / 0 fail** · Sonar gate **OK**, **0 open issues** · 2189 tests · Playwright **14/14 both themes** (a light-only claim was made true by adding the themes rather than by correcting the claim).
The 23 skips are Rust jobs excluded by the paths router — correct for dashboard-only, so **no Rust-side verification**; and **no CI job runs Playwright**, so the e2e evidence is author-run, not a gate.

### Reported, not fixed
All three pre-existing-red e2e specs seed `firstFiredAt: '2026-05-14…'`, so the newly-wired 24 h client-side window filters **every** row out. They are already 11/11 red on `main` (verified by stash-rebuild-rerun, identical both ways), so this PR doesn't regress them — but whoever fixes the existing failure will hit a **second, PR-introduced** one. Fix is relative fixtures.

Reviewed by Claude Code. Both endpoint contracts were verified against `openapi/v1.yaml` and the write-gate bypass against the component source.

PR #1714, opened 2026-07-26
[AAASM-5131] 🐛 (dashboard): Stop agent-detail posture asserting approvals it never measured

Chisanan232:
## Description

The agent-detail Overview **posture summary** asserted three figures it never measured.
`AgentDetailPage.tsx` hardcoded Narrow and Approval to a literal `value={0}`, set
`Deny = policy_violations_count`, and computed `Allow = max(0, session_count −
policy_violations_count)`. On the drawer's first panel — the surface an operator opens to
investigate *one* agent — "Approval 0" reads as *this agent needs no approvals* and
"Narrow 0" as *nothing is narrowed*: a fabricated all-clear.

**What each figure is actually knowable from:**

| Figure | Before | After | Knowable from |
|---|---|---|---|
| Allow | `max(0, session_count − policy_violations_count)` | count of `allow` cells | `GET /api/v1/capability/matrix` — agent × resource × verb |
| Narrow | hardcoded `0` | `not-supported` → `—` | **nothing.** `decide()` cannot emit it |
| Deny | `policy_violations_count` | count of `deny` cells | `GET /api/v1/capability/matrix` |
| Approval | hardcoded `0` | `not-supported` → `—` | **nothing.** `decide()` cannot emit it |

**The `sessions − violations` arithmetic does not survive scrutiny.** `session_count`
counts sessions handled and `policy_violations_count` counts violations recorded — two
different populations over two different windows, with no arithmetic relationship (the
code comment at the old `:170-177` conceded it was "approximate"). The subtraction could
go negative, which the `Math.max(0, …)` clamp then rendered as a confident `Allow 0`. It
is the same defect as the hardcoded zeroes, in a third place, and it is removed rather
than adjusted.

**Narrow and Approval are structurally unreachable**, verified against
`aa-api/src/routes/capability.rs`: `decide()` returns only `Decision::Allow` or
`Decision::Deny`, unmodelled verbs become `Na`, and `POST /capability/override`
**400s** on `Narrow` or `Approval` because such a value "would put a decision in the grid
that no projection can ever produce or restore". They are products of other policy stages
— credential scrubbing, and a tool's `requires_approval_if` CEL condition evaluated
against a *concrete action*. So the honest rendering is a vocabulary state, not a derived
number: re-deriving them from the matrix would yield a measured-looking `0` on every
deployment, which is the same lie with a citation. This is the display-side counterpart
of ADR 0026 Decision 2 — that ADR is `Proposed` and authorises nothing, and nothing here
decides it; a surface may always decline to assert what it cannot measure.

All four figures now flow through the shared truthfulness vocabulary (AAASM-5173):
`certainFromQuery` → `unavailable` on a failed request, `unknown` in flight,
`unconfigured` on an empty policy cascade (AAASM-5106), `not-evaluated` when the agent has
no matrix row, and `not-supported` for the two the backend cannot provide. An absent row
draws **no bar at all** — a zero-width fill is what a measured zero looks like.

### Notes for review

- **Labels changed** from `Allow / Narrow / Deny / Approval` to `… decisions`.
  `design/v2/hi-fi/agent-detail.jsx:270-273` says "resources allow", counting a *resource*
  as allowed when either its read or its write cell allows — a collapse across verbs that
  ignores delete/exec and has no counterpart in the API. Counting cells needs no such
  invented rule, so the label now names the unit that is actually counted. AAASM-5131
  called out the previous labels for drifting between the two units *silently*, not for
  the wording. Checked against ADR 0017 and ADR 0025: neither ratifies this panel, so
  nothing ratified is reverted.
- The `.ad-minibar` class names are deliberately unchanged so the existing agent-detail
  fidelity capture still matches.
- Scope: `pages/AgentDetailPage.*` and `components/agentDetail/**` only. No OpenAPI, no
  generated client, no shared truthfulness module, no Rust.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Display-only. No wire contract, no backend surface, no enforcement path is touched.

## Related Issues

- Related Jira ticket: [AAASM-5131](https://lightning-dust-mite.atlassian.net/browse/AAASM-5131)
- Related GitHub issues: N/A

## Testing

**Unit** — `agentPosture.test.ts` (20 cases) covers every way the matrix can be
untrustworthy and asserts Narrow/Approval stay `not-supported` in *all* of them, including
on a fully healthy matrix. Those assertions fail on the **type** of the figure, so a
reintroduced literal — or the tally's structurally-zero `narrow` — cannot pass them.
`AgentPostureSummary.test.tsx` (8 cases) drives the panel through the real query hook, so
the failed / pending / no-such-agent / empty-cascade branches are real branches.
`AgentDetailPage.test.tsx` pins the wiring: the fixture reports 10 sessions and 4
violations, and neither `6` nor `4` may reach the surface. New-code coverage is
**100% statements / branches / functions / lines**.

**Browser** — `tests/e2e/verify-aaasm-5131.spec.ts` (port 4587) drives a real HTTP 503 on
`/capability/matrix` through `openapi-fetch` and TanStack Query, in light and dark, and
asserts the operator sees `unavailable` rather than a posture — plus zero console errors.
Evidence committed to `dashboard/verify/5131/`.

```
pnpm test        Test Files 248 passed (248) · Tests 2463 passed (2463)
pnpm type-check  clean
pnpm lint        clean
pnpm generate:api → git status --porcelain: no generated-client drift
playwright --config tests/e2e/verify-5131.config.ts   4 passed
```

## Checklist

- [ ] All CI checks passing

[AAASM-5131]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5131, release-blocking)

**Verdict: COMMENT** — the fix is correct and well-evidenced. One user-facing decision needs a design answer before merge; nothing else blocks.

### The fix is right, and the deleted arithmetic is *more* justified than the PR argues
| Figure | Before | After | Knowable from |
|---|---|---|---|
| Allow | `max(0, session_count − policy_violations_count)` | count of `allow` cells | the capability matrix |
| Narrow | hardcoded `0` | `not-supported` → `—` | **nothing on this page** |
| Deny | `policy_violations_count` | count of `deny` cells | the same matrix |
| Approval | hardcoded `0` | `not-supported` → `—` | **nothing on this page** |

**The decisive fact the PR under-argues:** `session_count` and `policy_violations_count` have **no writer anywhere in production code** — every assignment outside tests and fixtures is `0`. I verified this independently. So the old expression evaluated `max(0, 0−0) = 0` on **every real deployment**. It was not a wrong number; it was a hardcoded zero wearing a formula. Deleting it destroyed no information — and it independently confirms **AAASM-5103** (the flagged badge that can never fire), same root cause.

**Unreachability verified**: `capability.rs:480-488` `decide()` returns only `Allow`/`Deny`; `:497-514` unmodelled verbs → `Na`; `:308-317` `POST /capability/override` **400s** on `Narrow`/`Approval` because such a value *"would put a decision in the grid that no projection can ever produce or restore"*. Correctly **not** re-derived from the matrix — doing so would have produced a measured-looking `0` on every deployment, which is the original bug in a better disguise. `tally.narrow`, a valid `known(0)`, is explicitly discarded for the same reason.

An absent matrix row draws **no bar at all**, because a zero-width fill is what a measured zero looks like.

### PRODUCT-DECISION-REQUIRED — the label change reverses the ticket's own target
AAASM-5131 lists the rewording as **part of the defect** (*"silently changing the unit from resources to decisions"*) and states the target as re-deriving the bars **"per the mock"**. This PR keeps `decisions` and makes it explicit.

The PR's rebuttal — that the ticket objected to the *silent* drift, not the wording — is defensible but self-serving, and its supporting argument mis-states the mock: `design/v2/hi-fi/agent-detail.jsx:270-273` uses `||` for allow/narrow/approval but **`&&` for deny**. The stronger and accurate objection is available — all four rows ignore `delete` and `exec` entirely.

**And there is a middle option nobody took**: count **resources where any of the four verbs allows**. That satisfies both the mock's unit and the objection to the verb-lossy collapse. ~10 lines either way. This needs a recorded design answer, not a PR-body rationale.

Note also that **ADR 0025 is `Proposed`**, not accepted — "v2 is authoritative" is an appeal to an authority that does not yet exist. Materially harmless here (v1 and v2 differ by two background-colour lines) but worth not repeating.

### Worth a knowing sign-off before release
On every shipped deployment this panel now renders **four dashes**: the policy cascade is empty (AAASM-5106), so allow/deny fold to `unconfigured`, and narrow/approval are permanently `not-supported`. That is the *honest* outcome and follows directly from the AAASM-5173 doctrine — but somebody should choose to ship it rather than discover it post-release.

### Minor
The bar denominator changed from `max(allow, deny, 1)` to **the sum of known figures**, so allow=2/deny=3 goes from 67%/100% to 40%/60% — a visible change to a fidelity-captured element, undisclosed in the PR body · the "existing fidelity capture still matches" claim is true only of the spec's one element-count assertion; the committed reference PNG is now stale (new caption, new labels, wider value column) · `.ad-minibar` deletion verified clean — exactly two consumers on `main`, both handled.

### Verified clean
ADR-0026 Decision 2 is scoped to `CapabilityFilterBar`/`BulkActionBar`; this PR touches neither, and nothing from that `Proposed` ADR was implemented. Vocabulary choices all correct — `unavailable` for a failed request, `unknown` while in flight, `not-evaluated` for an agent with no matrix row, `unconfigured` for an empty cascade, `not-supported` for the two unreachable verdicts. No local re-implementation of the vocabulary.

**Test integrity: clean.** `AgentDetailPage.test.tsx` is **+26 / −0**; the pre-existing posture assertions survive untouched. The new guards are unusually strong — `agentPosture.test.ts` asserts on the *type* of the narrow/approval figures across five scenarios including a fully healthy matrix, so a reintroduced literal cannot pass, and a 10-session/4-violation fixture asserts neither `6` nor `4` reaches the surface.

### Gates
CI **16 pass / 23 skipped / 0 fail** · Sonar **gate passed, 0 issues**, coverage 100% · `codecov/patch` pass · Playwright 4/4 light+dark with the failure path driven by a real 503.
The 23 skips are Rust jobs excluded by the paths router (dashboard-only), so **no Rust-side verification**; and **no CI job runs Playwright**, so that evidence is author-run.

Reviewed by Claude Code. The absent counter-writers, the `decide()` verdict set and the override 400 were verified against the Rust source directly.

PR #1715, opened 2026-07-26
[AAASM-5157] 🐛 (dashboard): Stop ungoverned agents vanishing from every Teams grouping

Chisanan232:
## Description

The Teams page derived its "unclaimed" list from `/topology/overview`'s
`standalone_root_agents`, which `aa-api/src/routes/topology.rs` filters as
`depth == 0 && team_id.is_none()`. An agent spawned by a parent (`depth > 0`)
that no team claims therefore appeared **under no team and not under unclaimed
either** — invisible on the one page whose job is showing which agents are
governed by whom — while the same response's `total_agent_count` still counted
it. The page's own numbers disagreed with its own list.

Orphans now come from the whole fleet (`GET /api/v1/topology`), filtered on
"no team" alone, so an agent's parentage no longer decides whether the
governance console admits it exists. The count chip and the list are now the
same measurement rather than two.

Three further consequences follow from the same standing rule — never turn a
failed or unavailable request into a legitimate business value:

- the unclaimed count is a `Certain<number>`, so a failed fleet request renders
  **Unavailable**, not a reassuring `0`;
- the unclaimed section renders even while the team list is failing, since it no
  longer depends on that query — an ungoverned agent must not become unreachable
  precisely when something is wrong;
- when the groupings and the registry's `total_agent_count` disagree, the page
  states the discrepancy as **Unknown** ("Agent totals disagree by N — X grouped
  here vs Y reported by the registry") instead of silently picking one number.

All states use the shared `lib/truthfulness` vocabulary (AAASM-5173); no new
status semantics are defined here.

### Departure from the ticket's prescribed approach — please review

The ticket's *Target behaviour* says to derive orphans from
`useTopologyQuery().data.nodes` (`features/topology/api.ts`). **This PR does not
do that**, and that is a deliberate departure a reviewer should weigh.

That hook maps the response through `mapGraph.ts` onto the graph view model,
which drops `depth` and folds any unrecognised `status` to `idle`. The orphan
rows render both fields verbatim (`depth 2 · off`, and the status chip), so reuse
would have regressed real agent state into a lossy projection on a governance
list — and fixing that at source means editing `features/topology/**`, which
another lane owns this wave. Instead `features/teams/api.ts` gets its own typed
query over the same `/api/v1/topology` path, returning raw `AgentNode`s. Same
endpoint, same predicate the ticket asks for, different projection.

Cost of the departure: a second React Query cache entry for one endpoint (keys
are never deduped; prefix matching applies to invalidation only). Mitigated with
the same `staleTime: 5_000` the sibling hook uses — which also matters on its own,
because this is the most expensive topology handler (a policy cascade and
effective permissions per node, plus a budget snapshot and the paged edge set,
nearly all of which Teams discards). If a reviewer prefers the ticket's literal
approach, the right sequencing is a follow-up that adds `depth` and a lossless
`status` to `TopologyNode` in the Topology lane, then collapses the two hooks.

### The census cross-check states less than it knows — on purpose

The two totals arrive in **separate responses**, so a delegation spawn landing
between them produces exactly the same arithmetic as a genuinely hidden agent,
and nothing available to this view distinguishes them. Two guards:

1. **Freshness gate** (`TeamsPage.tsx`) — the registry side is withheld while
   either query `isFetching`. TanStack keeps serving the *previous* payload
   throughout a background refetch, so without this an operator tabbing back
   could see a comparison between snapshots minutes apart.
2. **Weaker wording** (`CensusNotice`) — the notice reports *"Agent totals
   disagree by N … either a snapshot taken mid-change or agents this page cannot
   reach; this view cannot tell which."* It never asserts unreachability, because
   the gate cannot close the window entirely: two responses are never
   simultaneous.

Both are tested, including a skew case (stale fleet payload + fresh overview)
asserting the notice does **not** render; reverting the gate fails those two
tests.

**Matches the authoritative mock.** `design/v2/hi-fi/teams.jsx:223` computes
`nodes.filter(n => n.team === '__orphan__')` — every team-less node at any
depth, which is exactly this predicate. Checked against ADR-0017's ratified
Teams evolutions (items 20–21: members-as-agents, live approval routing); this
reverts neither. ADR 0025 makes `design/v2` authoritative, so the page docstring
now points there.

**Not in scope:** the second, divergent team-detail surface is AAASM-5158 and is
untouched here. Frontend-only — no backend, OpenAPI, or generated-schema change
(`pnpm generate:api` leaves the tree clean).

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Two internal component props changed shape (`TeamListPane.orphanCount`,
`TeamOrphanDetail.orphans`) from raw values to `Certain<…>`. Both components are
private to the Teams feature; no public API or wire contract is affected.

## Related Issues

- Related Jira ticket: [AAASM-5157](https://lightning-dust-mite.atlassian.net/browse/AAASM-5157)
- Related GitHub issues: N/A

## Testing

**Also fixed while here:** `TeamListPane` rendered `0 groups` next to its own
"Failed to load teams" banner, because `rows` is `[]` on both failure and first
load. Same defect as the unclaimed chip forty lines below; now a
`Certain<number>` too.

**Which agents were invisible, and how it was proved.** Exactly the agents with
`depth > 0` and no `team_id`. The gateway builds `team_map` only from records
with a `team_id`, `standalone_root_agents` only from `depth == 0 &&
team_id.is_none()`, and `total_agent_count` from *all* visible records
(`aa-api/src/routes/topology.rs`) — so such an agent is in the total and in
neither list. Empirically: the new Playwright spec was run against a bundle
built from the pre-fix sources and failed with
`expect(locator).toHaveCount(2) … Received: 1` — the spawned orphan was absent
from the only page that should have shown it. Against the fix, all 6 cases pass.

**Do list and count agree now?** Yes, by construction: chip and list derive from
one array. The independent cross-check against `total_agent_count` is asserted
green in the healthy case, and asserted to render an explicit `unknown` state
when the two sources are made to disagree — with the freshness gate and the
weaker wording described above so that state can never be a false accusation.

**Other surfaces** (audited, unchanged here): Fleet, Agent detail, IAM registry
and the violation heatmap already list such an agent; the Topology graph renders
it in an unnamed cluster; the Costs per-team tab and the LiveOps / Analytics
team filters exclude it, as team-scoped rollups legitimately do. Teams was the
only surface that hid it outright.

- `pnpm test` — 247 files, 2464 tests passed
- `pnpm type-check` — clean
- `pnpm lint` — clean, `--max-warnings 0`
- `pnpm generate:api` then `git status --porcelain` — empty
- `pnpm exec playwright test --config tests/e2e/verify-5157.config.ts` — 6 passed
  (3 scenarios × light/dark, console errors asserted `[]`); screenshots committed
  under `dashboard/verify/5157/`

## Checklist

- [ ] All CI checks passing — pending re-run on the review-response commits
- [ ] Commits are signed off (`git commit -s`) — advisory, not enforced

[AAASM-5157]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5157, release-blocking)

**Verdict: APPROVE** ✅ — after one MAJOR was found and fixed.

### The core fix is right, and proven the hard way
The invisible population is exactly `depth > 0 && team_id.is_none()`. Backend predicate confirmed verbatim: `topology.rs:595-597` filters `standalone_root_agents` on `depth == 0 && team_id.is_none()`, `:568-577` builds `team_map` from `is_some()` only, and `:566` counts **all** visible records — so such an agent was in the total and in neither list, on a governance console.

The author proved the new spec catches it by **building a bundle from pre-fix sources and watching it fail** (`toHaveCount(2) … Received: 1`). That is the difference between "my test passes" and "my test catches the bug".

### MAJOR — the census cross-check could state a false conclusion
`reconcileAgentCensus` compared two **independently fetched** snapshots with no freshness guard, and neither hook set a `staleTime`, so TanStack served the previous payload while refetching. A benign delegation spawn during a refetch → *"some agents are not reachable from any team or from this list."* False, on the surface whose entire ticket is about not making false governance claims.

**And my prescribed fix was insufficient.** I said "gate on both queries settled". The author correctly observed that two responses are **never simultaneous**, so a spawn landing *between* them yields the same arithmetic even when both are settled. Fixed with the gate **and** a weakened claim:

> *"These groupings and the registry were read from separate responses and report different totals. That is either a snapshot taken mid-change or agents this page cannot reach; **this view cannot tell which**."*

That reports the observation without asserting the cause — the right answer, and stronger than what I asked for. Reverting the guard reds 2 named tests, one driving exactly the skew scenario.

### Also fixed
`TeamListPane` no longer renders **"0 groups"** on a failed request (it derives a `Certain` from query state — the identical defect the PR was removing 40 lines below) · `staleTime: 5_000` added, matching the sibling hook and narrowing the skew window.

### The beyond-ticket query — justified, now disclosed
The ticket prescribes reusing `useTopologyQuery`. Both reasons for not doing so are **verified real**: `mapGraph.ts:76-92` produces a node with **no `depth` field at all**, and `:35-37` folds any unrecognised status to `'idle'` — and the orphan row renders both verbatim, so reuse would have been a live regression. That departure is now stated plainly in a dedicated "please review" section rather than left implicit.

### Verified clean
List and chip are **literally the same value**, so they cannot disagree by construction · every removed assertion has a strictly stronger replacement (`[…]` → `known([…])`, root-only → root + spawned) · the one fixture value change makes the fixture internally consistent rather than weakening it · pagination, tenant filtering and multi-team membership were each ruled out as census-mismatch causes · the frontend/backend empty-string predicate divergence is **unreachable** (`lifecycle_service.rs:467-471` normalises the proto3 default to `None`).

The sweep claim holds: Teams was the only surface hiding such an agent outright.

### Filed from this review
**AAASM-5181** (High) — the topology-overview cache key **omits `org_id`**, while the comment directly above it claims *"a tenant-scoped response is never served to a different tenant"*. Latent (admin-only, 1s TTL, dashboard never sends the param), but the comment asserts a guarantee the code does not provide.
**AAASM-5182** — Topology treats a team-less agent's `''` as a real team: inflates the header count, renders a blank filter chip, and its cluster click is dead.
**AAASM-5183** — Teams tells a **team-scoped** operator "No unclaimed agents" when `record_visible_to` makes any unclaimed agent invisible to them — a positive claim their scope cannot support. Pre-existing, but squarely inside this ticket's own thesis.

### Gates
CI **15 pass / 0 fail** · 2464 tests · Playwright 6/6 light+dark · `codecov/patch` pass.
**Sonar: gate passed, 0 issues** on the analysis covering the reviewed commit set; the 4 review-response commits were still queued in SonarCloud's compute engine at review time. Stated rather than glossed.
The 23 skips are Rust jobs excluded by the paths router; **no CI job runs Playwright**, so that evidence is author-run.

### Process note
A shared-scratchpad filename collision briefly put another lane's text on this PR's body. Self-detected, restored under a ticket-scoped name, and both PR bodies verified correct afterwards — #1716 was never written to. Second occurrence in this programme; ticket-scoped scratch filenames should be the standing rule.

Reviewed by Claude Code. The backend predicate, the `mapGraph` lossiness and the skew mechanism were verified against the source directly.

PR #1716, opened 2026-07-26
[AAASM-5135] 🐛 (dashboard): Stop Topology rendering unknown budgets as zero

Chisanan232:
## Description

Four Topology defects from the 2026-07-26 design-QA re-audit, all instances of the
same rule: **never convert a failed or unavailable request into a legitimate
business value, and disable actions with no successful production path.**

- **AAASM-5135** — `mapGraph.ts` collapsed `budget.limit_usd` with `?? 0`, so an
  unconfigured budget ceiling reached every consumer as a measured zero: the node
  panel showed `$0.00 / $0.00` at `aria-valuenow=0`, the card `$4.1 / $0`, the team
  bar `$0 / $0 · 0%`. `budgetLimit` is now `number | null` and the absence survives
  to the render sites, which use the shared `lib/truthfulness` vocabulary
  (`unconfigured`). Progress bars omit `aria-valuenow` entirely — ARIA's own encoding
  of an unknown value — and draw no fill, since a zero-width fill is still a rendered
  measurement of zero. A team total goes unknown if any member's ceiling is.
- **AAASM-5136** — added `refetchInterval: 5_000` to `useTopologyQuery`.
- **AAASM-5138** — under a team filter the canvas dropped every cross-team edge while
  the sidebar went on counting them. The sidebar keeps the fleet-wide figure and now
  **states the gap beside it** ("N not shown"), derived as `counted − drawn` from the
  same `isEdgeDrawn` predicate the canvas uses. Each visible card also carries the
  `⇆N` badge from `design/v2/hi-fi/topology.jsx:260,457`, for the narrower question of
  *which* visible agents have undrawn relationships.
- **AAASM-5140** — `⚖ Apply team policy` and `◐ Switch to shadow mode` were enabled
  with `onClick={() => {}}`. Both now render `disabled` with a `title`, matching the
  `View trace` button fifteen lines above. Deliberately **not** wired to anything.

### Three ticket claims were overstated or inverted — with evidence

**1. AAASM-5135 overstates the scope.** The ticket says the null-collapse also makes
the panel "render `0 policies`, asserting no policy governs the agent". That is false
for the endpoint this view model is built from. `project_graph_nodes` sets both fields
unconditionally:

- `aa-api/src/routes/topology.rs:427` — `node.policy_count = Some(cascade.len() as u32);`
- `aa-api/src/routes/topology.rs:434-437` — `node.budget = Some(NodeBudget { spend_usd: …unwrap_or(0.0), limit_usd })`, where `spend_usd` is a non-optional `f64`.

Only `limit_usd` is `Option<f64>` (resolved as per-agent override → server-wide daily
limit → `None`, `topology.rs:429-433`). The wire nullability of `policy_count` /
`budget` exists for the list / tree / team projections, which skip those lookups —
and `mapTopologyGraph` only ever sees the graph endpoint's payload. So `policyCount`
and `budgetSpend` stay plain numbers; their `?? 0` fallbacks are unreachable rather
than a defaulted absence. An agent absent from the budget-tracker snapshot has
genuinely spent nothing — the tracker is the authority on spend, so that `0` is
measured. **Fixed only what is genuinely absent.**

**2. AAASM-5136 overstates it too.** "Nothing refetches" is wrong: React Query v5
defaults `refetchOnWindowFocus` / `OnMount` / `OnReconnect` to `true` and they work.
The accurate claim is **no background polling** — `refetchInterval` appeared zero
times across `dashboard/src`. This is also what ADR-0017's own AAASM-5082 correction
records ("C2 — Item 3: the ratified 5s Topology polling was never implemented"), which
I verified on `main` before implementing. Adding the interval therefore **honours**
ratified item 3 rather than re-litigating it.

**3. AAASM-5138 is inverted.** The ticket says "`TopologyGraph.tsx` has no cross-team
badge" and the filter leaves no indicator. A cross-team badge *does* exist —
`TopologySidebar.tsx:77` renders `⇆ {stats.crossTeam} cross-team`. The real defect is
that it counts `allEdges` while the canvas silently dropped a subset, so the count and
the picture disagreed. The fix makes them agree rather than adding a missing badge.

### No ratified ADR-0017 entry reverted

All three Topology ratifications are intact: the **force-directed layout** (item 1) is
untouched, the **team budget bars on clusters** (item 2) are still rendered — only their
handling of a null ceiling changed — and **5s polling** (item 3) is now implemented for
the first time rather than removed.

### Notes for reviewers

- `TeamBudgetBar.spent` and `.limit` both widen to `number | null`, and `CostsPage`
  now passes them through instead of `?? 0`. Review was right that leaving this as an
  unowned disclosure was wrong: the component had learned to render an unconfigured
  ceiling while its other caller was wired to prevent it, so Costs still showed
  `$X / $0 · 0%` at `aria-valuenow=0`. Both fields are genuinely nullable —
  `joinTeamRows` yields `null` for a team missing from the cost breakdown *and* for
  every team while `/costs` is unresolved — so an absent spend now reports `unknown`
  (we asked, no answer) and an absent ceiling `unconfigured` (nothing was set up).
  `bucketForBudget` is deliberately **not** modified, keeping blast radius on
  `components/costs/`, `features/costs/` and `features/teams/` at zero.
- The optional LIVE indicator mentioned in AAASM-5136 is **not** included. A pill
  claiming "LIVE" needs its own answer for what it says while the poll is failing,
  which is a new product decision rather than part of this fix.
- `features/topology/api.ts` public shape is unchanged (the Teams lane reads it).
- A configured `$0` ceiling is kept distinct from an absent one throughout, and tested —
  `certain()` draws the same line, treating `0` as a value and only `null`/`undefined`
  as missing.

### Correction: an earlier revision of this PR claimed a false invariant

The first version of this branch claimed `drawn + badged == counted` and tested it on
a 2-team fixture where every crossing touched the filtered team — a shape in which the
claim **cannot fail**. Adversarial review found it false in general. Both breaks are
reachable:

1. **Three or more teams.** Teams A–B and B–C. Filter to A: the canvas draws 0
   crossings, badges appear only on A's nodes → 1, so `0 + 1 ≠ 2`. The B–C crossing
   touches no visible node, so no per-node badge can ever represent it.
2. **The `show cross-team` checkbox**, which sits directly beside the counter.
   Unchecking it hides every curve while no team filter is active, so no badge renders
   at all: `0 + 0 ≠ N` — verbatim the defect this ticket set out to fix.

A per-node badge structurally cannot reconcile a fleet-wide count, so it no longer
claims to. The counter is not narrowed to match the picture either — that would throw
away the fleet fact. Instead the discrepancy is rendered, derived as `counted − drawn`
so it also covers a third case neither break named: an unchecked edge kind.

`crossTeam.test.ts` had stated the correct, narrower invariant all along (per-team
badge sum == that team's crossings); it was the PR body and the page test that
generalised it. Both are fixed, and the new page tests use a three-team chain plus the
toggle and edge-kind cases — **all three fail if `crossTeamHidden` is reverted**
(verified).

### Adjacent defects found and deliberately NOT fixed here

While the Teams lane (#1715) audited which surfaces hide team-less agents, it surfaced
two defects in files I own. I reproduced both against this branch — a graph with one
`support` agent and one agent whose `team_id` is null (`mapGraph.ts:80` coerces it to
`''`):

```
HEADER = "· 2 agents · 2 teams"     ← only ONE real team exists
ROWS   = "all"=>"◎All teams" , ""=>"○" , "support"=>"○support"   ← blank-labelled row
CLICK_BLANK_CLUSTER = NOTHING       ← TopologyPage.tsx:131 `else if (selectedTeam)`
                                       is falsy for '', so the panel never opens
```

So: the header count asserts a team that does not exist, the sidebar offers a filter
row that names nothing, and the canvas renders a blank cluster whose click is dead —
the same two classes this PR fixes elsewhere (a count disagreeing with reality; an
action with no successful path).

**They are not fixed here, on purpose:**

1. **The vocabulary belongs to another lane and has not merged.** #1715 establishes the
   "unclaimed" concept in `features/teams/orphans.ts`. Naming the group in Topology
   would mean either duplicating that concept inside `features/topology/` — precisely
   the two-definitions-of-one-thing drift `lib/truthfulness` exists to prevent — or
   importing across feature modules from an unmerged branch. Inventing a third label
   here would be worse than both.
2. **The design-authoritative fix is a feature, not a patch.**
   `design/v2/hi-fi/topology.jsx` specifies a distinct orphan cluster (`:190`, `:784`),
   a dedicated detail panel with its own copy (`:664-686`), a warn-toned
   `⚠ N unclaimed` header stat (`:952`), a `teamCount` that excludes orphans (`:928`),
   and a bulk *Suspend all unclaimed* action (`:765`). Which of those ship is a product
   call; shipping a partial version inside a budget/polling/badge PR would pre-empt it.
3. **Reviewability.** This PR is already four tickets across 20 files.

**Filed as [AAASM-5184](https://lightning-dust-mite.atlassian.net/browse/AAASM-5184)**,
stacked behind #1715 so it can reuse that lane's `orphans.ts` vocabulary rather than
fork it. Root cause recorded there: `mapGraph.ts:80` `team: n.team_id ?? ''` — the
coercion that turns "no team" into a team named empty-string.

**One half was taken here**, because it needed no naming decision: the blank cluster's
click had no successful path (`TopologyPage.tsx` gates its panel on a truthy team), so
it now renders non-interactive rather than looking clickable and doing nothing — the
same rule as AAASM-5140, twelve lines away in the same diff. The count, the blank
filter row and the blank label stay with AAASM-5184.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature _(the ratified 5s poll, previously unimplemented)_

## Breaking Changes

- [x] No

`TopologyNode.budgetLimit` and `TeamBudgetBarProps.limit` widen to `number | null`.
Both are internal dashboard types with no external consumers, and widening a prop is
source-compatible for every existing call site.

## Related Issues

- Related Jira tickets:
  [AAASM-5135](https://lightning-dust-mite.atlassian.net/browse/AAASM-5135),
  [AAASM-5136](https://lightning-dust-mite.atlassian.net/browse/AAASM-5136),
  [AAASM-5138](https://lightning-dust-mite.atlassian.net/browse/AAASM-5138),
  [AAASM-5140](https://lightning-dust-mite.atlassian.net/browse/AAASM-5140)
- Related GitHub issues: N/A

## Testing

The three tests specifically required of this lane:

| Claim | Test |
|---|---|
| A null `limit_usd` never renders `0` or `aria-valuenow=0` | `TeamBudgetBar.test.tsx` "never renders a 0 limit or a 0 percentage" / "omits aria-valuenow so the progressbar reads as indeterminate"; `NodeDetailPanel.test.tsx` "never renders aria-valuenow=0 on the progress bar"; `mapGraph.test.ts` "keeps a null budget limit null rather than collapsing it to 0" |
| The sidebar count and the rendered edge set agree | `TopologyPage.test.tsx` "accounts for every crossing the team filter removes from the canvas" (`drawn + badged == counted`); `crossTeam.test.ts` "sums per-team badges to that team's crossing count" |
| The dead buttons cannot be activated | `NodeDetailPanel.test.tsx` "%s cannot be activated by click or keyboard" |

Each has a control alongside it so the fix cannot pass by blanking the surface: a
configured `$0` ceiling still reports a ratio, an unfiltered graph draws every crossing
it counts, and the suspend / view-trace actions beside the disabled pair stay enabled.

**Gates** (re-run after the review fixes, all green):

```
pnpm test        Test Files  248 passed (248)   Tests  2508 passed (2508)
pnpm type-check  tsc --noEmit — no output
pnpm lint        ESLint: No issues found
pnpm generate:api → git status --porcelain — empty (no generated-schema drift)
```

**Playwright** — repo-local `@playwright/test` on port 4590
(`tests/e2e/verify-5135.config.ts`), `6 passed`. Light **and** dark for both
scenarios, console errors asserted `[]`. Covers the ≥3-team break and the
`show cross-team` toggle, and serves a *drifting* payload on each poll so the
changed-payload re-render path is genuinely exercised. Uses `page.route(...)` rather than a `fetch`
shim (`openapi-fetch` captures `globalThis.fetch` at module load) and seeds `aa_token`
via `addInitScript`. The 5s poll is counted at the network layer rather than read back
off the query options — reading the option back is effectively what ADR-0017 item 3 did
while the poll did not exist. Screenshots committed under `dashboard/verify/5135/`.

No Rust suite run — this change is dashboard-only.

## Checklist

- [ ] All CI checks passing _(pending first run)_
- [ ] Commits are signed off — advisory, not enforced

[AAASM-5184]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5135 · 5136 · 5138 · 5140)

**Verdict: APPROVE** ✅ — after one BLOCKER and two MAJORs were found and fixed.

### BLOCKER — the claimed invariant `drawn + badged == counted` was false
Two reachable breaks: with **≥3 teams** (A–B and B–C crossings, filter to A → `0 + 1 ≠ 2`), and via the **`showCrossTeam` checkbox sitting directly beside the counter** (`0 + 0 ≠ N`) — verbatim the AAASM-5138 defect, reachable from the control next to the number. **The test could not detect either**: a 2-team fixture in which *all* crossings touch the filtered team.

The fix generalised past both: `isEdgeDrawn` is now **one statement of what reaches the screen, used by both the canvas and the counter**, with the gap derived as `counted − drawn`. That derivation covers a **third** case neither the review nor the PR had named — an unchecked **edge kind**. The sidebar shows `⇆ 4 cross-team` / `4 not shown` naming all three causes, and the per-node badge keeps only the narrower claim `crossTeam.test.ts` already stated correctly; its tooltip no longer says "hidden by the team filter", because it cannot know the cause. Reverting reds exactly the 3 new tests.

### MAJOR — the new 5s poll re-heated the force simulation
`teamCenterById` produced a **new `Map` every render**, so `simulation` was rebuilt and `sim.alpha(1).restart()` ran over freshly-constructed, un-positioned nodes. In a live fleet `budgetSpend`/`status` change constantly, so the graph would **re-scatter every 5 seconds, moving click targets under the operator**. The e2e was blind to it: `GRAPH` is a module constant, so the poll never produced a changed payload.

**And the obvious test for it does not work** — d3 is deterministic, so a full rebuild re-initialises to the same phyllotaxis and converges back to *identical* transforms; the first attempt passed against the reverted code. It asserts **simulation construction count** instead. Fixing it then surfaced a second bug the author caught: with the sim no longer rebuilt, `positions[].source` goes stale, so *"stop re-scattering"* would have become *"stop updating"*. Node data now reads live via `nodeById`.

Everything else about the polling was verified sound: no polling on a hidden tab, stops on unmount, concurrent fetches deduped, scoped to the one query.

### MAJOR — the same `?? 0` lie survived on a caller of the component this PR fixed
`CostsPage.tsx:145` still passed `?? 0` into the `TeamBudgetBar` this PR had just taught to render `unconfigured`. The PR had disclosed it as "that lane's own ticket" while naming **no ticket**. Taken here — and correctly widened, because `spent` is nullable too: `joinTeamRows` yields `null` for *every* team while `/costs` is unresolved, so `$0 spent` was a lie whenever that request failed. Absent spend → `unknown`; absent ceiling → `unconfigured`; a measured `$0` and a configured `$0` both stay measurements.

### Verified correct
Every backend citation holds: `topology.rs:427`/`:434` are unconditional, only `limit_usd` is `Option`. Progress bars **omit `aria-valuenow`** and render **no fill element** — not a zero-width one. The three ratified ADR-0017 Topology entries are intact, and 5s polling is **implemented for the first time**, not reverted.

**All three of this PR's ticket corrections were independently confirmed**: 5135 overstated (`policy_count`/`spend_usd` are always populated); 5136 overstated ("nothing refetches" is wrong — focus/mount refetch works; the gap is *background polling*); 5138 **inverted** (a badge already existed; the defect was count-vs-picture disagreement).

### Declined items — right call, and now filed
The orphan-cluster feature was rightly declined: #1715 owns the `orphans.ts` vocabulary and was unmerged, and design v2 specifies a whole orphan feature that is a product call. **AAASM-5184 filed** rather than merely recommended, and the **dead-click half was taken** — making the unnamed cluster non-interactive needed no naming decision.

Also corrected: `docs/src/adr/0017` now records item 3 as resolved by AAASM-5136, appended rather than edited — for a lane whose thesis is that records must not lie, leaving it "ratified but unimplemented" would have been wrong in the opposite direction.

### Gates
CI **19 pass / 0 fail** · 2508 tests · Playwright 6/6 light+dark.
**Sonar: gate passed, 0 issues** on the analysis covering the reviewed commit set — the 6 review-response commits were still queued in SonarCloud's compute engine at review time (nothing processed project-wide since 13:10Z). Stated rather than glossed.
This lane also sharpened the verification method twice: a Sonar `0` can predate the commits, and **the reliable check is the CE task status, not the PR's presence in `project_pull_requests/list`** — my own guard passed while the reading was stale.
The 24 skips are Rust jobs excluded by the paths router; **no CI job runs Playwright**, so that evidence is author-run.

Reviewed by Claude Code. Both invariant breaks and the simulation chain were traced in the source directly.

PR #1717, opened 2026-07-26
[AAASM-5129] 🐛 (dashboard): Stop Live-Ops claiming latency and approvals it never measured

Chisanan232:
## Description

Four Live-Ops defects, all the same shape: the page asserted things it had not
measured, and offered controls that could not succeed.

- **[AAASM-5129](https://lightning-dust-mite.atlassian.net/browse/AAASM-5129)** —
  `useLiveOpsStream` wrote `latencyMs: 0` on every row (unconditionally on
  `ops_change`, via `?? 0` on violations) and `formatLatency` rendered anything
  under 1 as `<1ms`. `ViolationPayload.latency_ms` is documented as *not
  populated today*, so in production **every stream row claimed a
  sub-millisecond latency that was never measured**. `opType` and `resource`
  were coerced to `'unknown'` / `''` the same way. All three are now
  `Certain<T>` and render through the shared `TruthfulValue`: a missing
  violation latency is `unknown` (the field exists on the wire and fills itself
  in), an `ops_change` row is `not-supported` (that payload has no field for
  it), and a latency that really arrived — including a measured `0`, which now
  prints `0ms` — is still asserted.
- **[AAASM-5167](https://lightning-dust-mite.atlassian.net/browse/AAASM-5167)** —
  `ApprovalPool` returned `null` when empty, so a broken queue and a clear one
  were the same blank panel. It now renders `No pending approvals` for a
  loaded-and-empty queue and the `unavailable` state (with a retry) for a
  failed one, and the pane head carries the waiting count as a `Certain` value
  so an outage reads `—`, never `0 waiting`.
- **[AAASM-5148](https://lightning-dust-mite.atlassian.net/browse/AAASM-5148)** —
  no RBAC on approve/reject, the row action menu, or the fleet-wide halt-all. A
  read-only operator saw an enabled kill switch and learned otherwise from a
  403 toast. Gated on `canWrite`, following the AAASM-5147 pattern — including
  its lesson: the gate is applied at the *last* control before the mutation
  (`ApprovalActions` itself, the reject-confirm button, and the confirmation
  dialogs), not only at the entry points.
- **[AAASM-5128](https://lightning-dust-mite.atlassian.net/browse/AAASM-5128)** —
  the approval queue was the `status === 'pending'` slice of the ops ring, and
  POSTed the row's op id (a governance-event id, or a `"{trace_id}:{span_id}"`
  composite) to `/approvals/{id}/approve`. The gateway parses that segment with
  `Uuid::parse_str`, so **every decision issued from this pane 400'd before the
  queue was consulted**. It is now fed by `useApprovalsQuery` +
  `useApprovalsStream`, keyed by `Approval.id` — which also brings `expires_at`
  and so the TTL countdown.

### Notes for the reviewer

- **A real latency source exists in the schema but is not populated today.**
  `ViolationPayload.latency_ms` is `Option<u64>`, documented "populated once
  the audit pipeline tracks per-action duration end-to-end"; `CallStackNode`
  carries its own. `OpsChangePayload` has no latency field at all. That is why
  the two absences use different states — one resolves itself when the backend
  lands, the other never will.
- **The speed / pause header controls are deliberately left ungated.** They
  drive the client-side pipeline animation and mutate nothing; gating them on
  write scope would misrepresent them as fleet actions. Related: the header
  pause is local `useState` and does not stop the event stream, while `PAUSED`
  takes the pill's highest precedence and can therefore mask a dead socket —
  that is AAASM-5153 / ADR-0026 Decision 7, still `Proposed`, and **is not
  decided here**. This lane does not touch `derivePill` and does not make the
  conflation worse.
- **`page on-call` is still a mock action** (`toast('… mock action')`). Gating
  it on write scope would have implied it is a real fleet mutation, so it was
  left alone — flagging it rather than deciding it.
- Checked against ADR-0017 and ADR-0025 (`design/v2/hi-fi/live-ops.jsx` is
  authoritative): the 5-state pill and the client-side pipeline simulation are
  ratified evolutions and are untouched. The waiting-count chip and the
  `no pending approvals` copy come from the v2 mock.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

`LiveOperation`'s `opType` / `resource` / `latencyMs` change type, but every
consumer is inside `features/liveOps/**` and is updated here. No wire, OpenAPI
or generated-client change: `pnpm generate:api` is a no-op on this branch.

## Related Issues

- Related Jira tickets:
  [AAASM-5129](https://lightning-dust-mite.atlassian.net/browse/AAASM-5129),
  [AAASM-5167](https://lightning-dust-mite.atlassian.net/browse/AAASM-5167),
  [AAASM-5148](https://lightning-dust-mite.atlassian.net/browse/AAASM-5148),
  [AAASM-5128](https://lightning-dust-mite.atlassian.net/browse/AAASM-5128)
- Related GitHub issues: N/A

## Testing

Vitest — the branches that matter, driven rather than asserted around:

- a hardcoded latency cannot come back: the mapper test fails if a missing
  `latency_ms` becomes a known value, and the row test fails if an absent
  latency renders any duration string;
- a broken pool and an empty pool are asserted to render *differently*, not
  just to each render something;
- the RBAC tests mount an `AuthContext` so they reach the gated state at all —
  without a provider `usePermissions` falls back to every scope, which is why
  this gap survived — and the confirm-dialog / reject-confirm cases drive a
  scope that lapses mid-flow, the only state that reaches those controls past a
  disabled entry point.

Playwright (`tests/e2e/review-aaasm-5129.spec.ts`, port 4589, light **and**
dark, console errors `[]`): the socket withholds `latency_ms` exactly as the
gateway does, an `ops_change` frame arrives, `GET /api/v1/approvals` is failed
at the network layer, and the approve request's path segment is asserted to be
a UUID. Screenshots committed to `dashboard/verify/5129/`.

Gates, each run once from `dashboard/`:

```
pnpm test        → Test Files 247 passed (247) · Tests 2466 passed (2466)
pnpm type-check  → tsc --noEmit, clean
pnpm generate:api → git status --porcelain empty
pnpm exec playwright test -c tests/e2e/review-5129.config.ts → 14 passed
```

No Rust suite run — this branch touches no crate.

## Checklist

- [ ] All CI checks passing

[AAASM-5129]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAASM-5167]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Claude Code — adversarial review (AAASM-5129 · 5167 · 5148 · 5128)

**Verdict: APPROVE** ✅ — after one BLOCKER and one MAJOR were found and fixed.

### BLOCKER — deciding an approval left two surfaces asserting a stale count
The pool hid decided rows in **component-local state**, while the pane-head chip and the header bell both read the raw `['approvals']` cache — and nothing wrote it. Verified: `features/approvals/api.ts` had **0** `setQueryData`/`invalidateQueries` calls; `ApprovalsPage.tsx` has **4**.

Approve on Live-Ops with one pending item → body reads "No pending approvals" while the chip reads **"1 waiting"** and the bell badge reads **"1"**. Exactly the head/body contradiction AAASM-5167 exists to remove — and the assertion that guarded it (`// Count reflects the shrunk queue`) had been **deleted, not relocated**.

**Fixed on the mutations, not the callers** — `useApproveAction`/`useRejectAction` own the decide request, so they are the only place that sees every decision (Live-Ops, the Trace drawer, and whatever mounts them next). `undefined` is preserved rather than replaced with `[]`, because an unloaded cache has no list to filter and inventing an empty one would assert a clear queue off a single decision. Success-only, so a refused decision leaves the approval pending.

**The new test deliberately does *not* mock `useApprovalsQuery`** — every other spec does, which makes body and chip independent fixtures that can never contradict each other. It runs the real hook against a real `QueryClient` and mounts the bell alongside the page under one client, because "the two disagree" is only observable when they share a cache, which in the real app they do. Reverting reds 3 tests, the first verbatim: `Expected: 1 waiting / Received: 2 waiting`.

### MAJOR — AAASM-5167 was fixed in the pane but not on the always-visible header
`ApprovalsBellButton` was `data?.length ?? 0` with the badge hidden at `0`, so an approvals **outage** rendered exactly like a clear queue — visible in the PR's own `approvals-outage-light.png`, where the pane correctly announced UNAVAILABLE beside a bare "⚑ approvals". Now three-way via `certainFromQuery`, with the distinction carried in the **`aria-label`** (the badge is `aria-hidden`, so the link's accessible name is all assistive tech receives) and a test asserting it matches `/unavailable/` and **not** `/no approvals are waiting/`.

### Verified
All three AAASM-5128 justifications hold in source: `approvals.rs:360` `Uuid::parse_str` 400s **before** the queue is consulted; the pool was sending a governance-event id or a `{trace}:{span}` composite; and the ops socket subscribes to `types=violation,ops_change`, so it **never receives an approval event at all**. Re-sourcing (rather than disabling) was right, and it brought `expires_at` and the TTL countdown with it.

**5129's latency states are correct and deliberate**: `ws_payloads.rs:67-70` declares `latency_ms: Option<u64>` "populated once the audit pipeline tracks per-action duration", while `OpsChangePayload` has **no such field**. So a missing violation latency is `unknown` (fills itself in when the backend lands, no dashboard change) and an ops_change row is `not-supported` (waiting will never help). A wire `0` stays a known `0ms`.

Side-effect analysis of the data-source swap: no consumer depends on the old shape; `useApprovalsQuery` was **already** app-wide via the bell, so no extra HTTP request; the pool and stream dedupe by `id` into one cache; reconnect is safe. `ApprovalActions` is shared with the Trace drawer, so the new RBAC gate correctly covers that surface too.

**Slow/fast/pause stay ungated — correct.** All three are local state with no network call; gating them would assert they are fleet actions, which is precisely the AAASM-5153 conflation. The test pinning that choice is kept.

### Also fixed
`page on-call` disabled with a reason (matching sibling PR #1716's treatment of dead governance buttons) · call-stack latency now carries state, tone and an announcement instead of a bare dash · symmetric `if (!canWrite) return` guards · the approvals socket's `connected` flag consumed instead of discarded.

### Filed from this review
**AAASM-5180** (High) — `usePermissions.ts:36` returns **ALL_SCOPES** without an `AuthContext`, so five further gated surfaces' RBAC specs pass against a permissive default, including **API-key revoke**. Deleting a gate would keep CI green. This lane is what exposed it.

### Gates
CI **15 pass / 0 fail** · 2483 tests · Playwright 14/14 light+dark · changed areas 93–100% lines locally · `codecov/patch` pass.
**Sonar: gate passed, 0 issues** on the analysis covering commits 1–9 — the 7 review-response commits were still queued in SonarCloud's compute engine at review time (task PENDING ~65 min; nothing processed project-wide since 13:10Z). Stated rather than glossed. Re-check once it drains.
The 23 skips are Rust jobs excluded by the paths router (dashboard-only PR), so **no Rust-side verification**; and **no CI job runs Playwright**, so that evidence is author-run.

Reviewed by Claude Code. The cache-write absence, the UUID parse order and the ops-socket subscription were verified against the source directly.

PR #1718, opened 2026-07-26
[AAASM-5124] 🐛 (dashboard): Stop the bulk override defaulting to a decision the API rejects

Chisanan232:
## Description

**Reworked after review — the earlier `deny`-default rationale is superseded by an explicit engineer decision.** The first revision of this PR argued that, of the three decisions the endpoint accepts, `deny` was the safe pre-selection. Review showed that trade was backwards: it converted an unconsidered click from a harmless guaranteed-400 into a **successful bulk write across every selected agent, with no undo in the UI** — the FE client exposes only `getMatrix` and `applyOverride`, though the endpoint has a `DELETE`. That rationale no longer applies. This PR now implements three product decisions taken by the engineer.

### Decision 1 — no default; disabled until chosen; confirmed before writing

> *"Do not default the bulk action to deny. The default state must be no selection. Disable Apply until the operator explicitly selects an action. Require confirmation before a bulk write. Do not introduce a destructive or restrictive default with no undo."*

- Nothing is pre-selected — the decision `<select>` opens on a `select a decision` placeholder that is not a member of `Decision`.
- The primary control is **disabled** until a decision is chosen, with the reason on the control (`NO_DECISION_TITLE`), matching the disabled-with-a-reason pattern already used in `components/topology/NodeDetailPanel.tsx`.
- The write sits behind a **confirmation** that names the decision, the verb, the resource, and **how many agents** it affects. Changing the decision or the resource retracts a pending confirmation, so the text an operator accepts always describes the write that follows.
- An unparseable `<select>` value now falls back to *no selection* (re-disabling the control) rather than leaving the previous decision armed behind it.

**No undo affordance was added, and it is not cheap.** The gateway has `DELETE /capability/override`, but reaching it needs a new client method in `src/api/capability.ts`, a new page-level revert path, and a product answer about what "undo" means for an overlay that is replayed on read and has no per-write handle in the FE (the response's `overrideId` is currently discarded). That is a ticket, not a line. The confirmation step is the compensating control; the residual risk is recorded in the ADR rather than half-built here.

### Decision 2 — legend narrowed to what the projection can emit (resolves ADR-0026 D2)

> *"show only states the current projection can actually emit · include unconfigured / unknown where applicable · remove narrow and approval from the legend, filters, and mutation controls until a real backend computation can produce them · do not preserve aspirational states merely to match an old mock"*

This answers **ADR-0026 Decision 2 in favour of its recommended option (A)** — legend **and** controls. The previous revision of this PR deferred the legend on the grounds that D2 was `Proposed`; it omitted that D2's option (C), *"keep the legend aspirational"*, is **explicitly rejected in the ADR as the exact pattern this programme exists to remove**. That omission is corrected.

- **Legend** → `allow` / `deny` / `n/a`. The `--narrow` / `--approval` swatch rules are deleted with it.
- **Filters — verified no-op, and pinned as one.** `CapabilityFilters` is `{ search, framework, owner, mode, trustMax }`; the legend is a static non-interactive `<ul>`. Both were re-verified on the current tree — there is no decision filter to remove. A test now pins the control list closed so one cannot be added later without re-deciding D2.
- **Mutation controls** were already narrowed earlier in this PR; unchanged.
- **`unconfigured` is deliberately not added.** The decision says "where applicable" and it is not yet applicable: ADR-0024 *proposes* the state, nothing emits it, and `Decision` has no such member. Listing it now would be the same defect pointing the other way.
- **`Decision` is untouched** — five members, no wire-contract change, as D2's recommendation requires.

### Decision 3 — AAASM-5178 immediate remediation (language and affordance only)

> *"disable, remove, or explicitly label display-only overrides immediately · never report that enforcement changed when only the dashboard annotation changed · keep real enforcement wiring as a separate implementation task"*

`aa-api/src/routes/capability.rs`'s module doc records that overrides *"have never fed enforcement — the store is read by these four handlers and nothing else"*. The button read `Apply override` and the success toast read `override applied to N agents`; both read as though a gateway decision had changed.

- Button → **`Record display-only override`**.
- Confirmation carries the disclosure at the point of action: *"Recorded as a dashboard annotation — this does not change what the gateway enforces."*
- The confirmation is a `<fieldset>`, not a labelled `<div>` — an `aria-label` on a role-less element is ignored by assistive tech, which would have left the disclosure invisible to exactly the operator it protects. Deliberately not `alertdialog`, which would promise modal focus management this does not implement.
- Toast → **`display-only override recorded for N agents — the dashboard annotation changed; gateway enforcement did not`**. It states that the annotation landed *and* that enforcement did not follow; silence on the second half is what made the first deceptive. It is deliberately not alarmist — nothing failed, and it does not say so.

**No enforcement was wired.** That is AAASM-5178 option (c), it overlaps the mutation-safety design in ADR-0021, and it stays a separate task with its own ADR.

## ADR action

**`amend` ADR-0026** — an `Update — AAASM-5124 / AAASM-5178` section records D2's sign-off as option (A), what it resolved to (including the verified filter no-op and why `unconfigured` is excluded), the AAASM-5178 disclosure decision, the explicit deferral of enforcement wiring, and the accepted residual risk (no undo affordance). The header status and the ADR index now read `Proposed (Decision 2 Accepted)`; the other six decisions are untouched and still await sign-off.

Searched before editing: ADRs 0017 (Accepted — *"where the mock implies a capability the backend does not model, ship what the data supports"*, which **supports** this change), 0021, 0024, 0025, 0026, plus `SECURITY.md` / `CONTRIBUTING.md` / `CLAUDE.md` and the linked Jira. **No conflict with any Accepted decision.** No new ADR number, so no `SUMMARY.md` change.

Chain: Jira [AAASM-5124] / [AAASM-5178] → ADR-0026 Decision 2 + Update → this PR.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

No wire contract changes. `Decision` still has five members and the OpenAPI schema is byte-identical after `pnpm generate:api`. The changes are the FE's advertised subset, the bulk bar's interaction model, and operator-facing copy.

## Related Issues

- Related Jira tickets: [AAASM-5124](https://lightning-dust-mite.atlassian.net/browse/AAASM-5124), [AAASM-5178](https://lightning-dust-mite.atlassian.net/browse/AAASM-5178)
- Related ADR: `docs/src/adr/0026-open-dashboard-product-semantics.md` (Decision 2)
- Related GitHub issues: N/A

## Testing

- **`BulkActionBar.test.tsx`** — nothing pre-selected; the control disabled with its reason attached; the resource fallback and both render-nothing guards; no write when applied without a decision; no write until the confirmation is accepted; the confirmation names the decision and the affected-agent count (and singularises it); the disclosure text present; cancel writes nothing and re-arms; a pending confirmation retracted when the decision or the resource changes under it; the offered set is exactly the accepted subset plus the placeholder; a rejected decision pushed at the select disarms the control.
- **`CapabilityFilterBar.test.tsx`** — the legend is exactly `allow`/`deny`/`n/a` with neither removed state surviving as an entry or a swatch class; and the filter bar carries exactly five controls, none keyed on a decision.
- **`CapabilityPage.test.tsx`** — every run that expects a POST now goes through one choose-and-confirm helper, so removing either gate fails all of them. Adds: no write before a decision is chosen; no write when the confirmation is dismissed; and the success toast reports the annotation without claiming enforcement moved.
- **`tests/e2e/review-aaasm-5124.spec.ts`** (port **4601**) — four runs per theme in a real browser: nothing pre-selected and the apply control disabled (force-clicked past its pointer-events guard, to prove the handler itself writes nothing); the confirmation naming count, decision and disclosure, then cancelled with no request sent; the narrowed legend; and a confirmed override asserted through the held-open POST via a `MutationObserver` installed before any app module. Light **and** dark, console errors asserted `[]`. Screenshots in `dashboard/verify/5124/`.

**Revert-checks.** Each new behaviour was reverted in source and watched go red, then restored with `git checkout HEAD -- <path>`:

| Reverted behaviour | Red |
|---|---|
| Restore a `deny` default | 4 tests |
| Apply writes directly, bypassing the confirmation | 13 tests |
| Confirmation text drops the count and the decision | 2 tests |
| Re-add `narrow`/`approval` to the legend | 1 test |
| Restore a `deny` default, **rebuilt**, e2e | 2 e2e runs |

The e2e revert-check was run against a **rebuilt** `dist` with the build's exit code confirmed (`0`) and `dist` removed first — a failed `pnpm build` leaves `vite preview` serving a stale bundle, which would make a mutation check pass and falsely read as "the tests aren't load-bearing".

Gates, each run once on the final tree:

```
pnpm test         →  Test Files 254 passed (254) · Tests 2638 passed (2638)
pnpm type-check   →  tsc --noEmit, no output
pnpm lint         →  ESLint, no issues (--max-warnings 0)
pnpm generate:api →  git status --porcelain empty (no schema drift)
playwright --config tests/e2e/review-5124.config.ts → 8 passed
```

CI on the head commit: **20 passed, 24 skipped, 0 failed**. SonarCloud: **Quality Gate passed, 0 new issues, 0 accepted, 0 security hotspots, 97.4% coverage on new code, 0.0% duplication** — an S6819 finding on a first-pass `role="group"` was fixed by switching to `<fieldset>`, not suppressed; no `NOSONAR`, no rule disables. Codecov patch green. Locally every changed component is fully covered: `BulkActionBar.tsx` 24/24 lines and 17/17 branches, `CapabilityFilterBar.tsx` 16/16 and 4/4.

No Rust code changed, so no Rust suite was run.

## Noticed, not changed

The Capability summary row renders a **`narrowed` counter** (`CapabilitySummary.tsx:61`) which, for the same reason the legend entry was removed, can only ever read `0` from this projection. It is not one of the three surfaces the decision enumerated (legend, filters, mutation controls), so it is reported rather than silently folded in.

`dashboard/verify/5124/bulk-options-{light,dark}.png` are now orphaned — the run that produced them was renamed, and they show the superseded pre-selected bar. Left in place rather than deleted without sign-off.

## Checklist

- [x] Documentation updated if behaviour changed (ADR-0026 amended)
- [x] All CI checks passing

[AAASM-5124]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Adversarial review — AAASM-5124 / PR #1718

Reviewed independently: every claim in the PR body was re-derived from the code and re-run
locally, including a mutation check that reintroduces the pre-fix behaviour. I did not
write this PR and did not take the author's reasoning on trust.

**Verdict: the fix is correct and genuinely evidenced. No BLOCKER.** One MAJOR safety
consequence of the chosen default should be resolved or explicitly accepted before merge,
and one MAJOR consistency gap should be filed as a follow-up. Not approving or merging —
this is a review record only.

---

### What I verified (all claims re-derived, not relayed)

| Claim | Result |
|---|---|
| `POST /capability/override` 400s on `narrow`/`approval` | **TRUE** — `aa-api/src/routes/capability.rs:308-317`, guarded by `matches!(body.decision, Decision::Narrow \| Decision::Approval)` |
| Bar previously defaulted to `narrow` | **TRUE** — reproduced by mutation |
| Optimistic paint happens *before* the POST | **TRUE** — `CapabilityPage.tsx:94` `setOptimistic(...)` precedes `await capabilityClient.applyOverride(...)` at `:97` |
| "8 impossible cells" under the pre-fix default | **TRUE, exactly** — my mutation run reproduces `expected [...] to have a length of +0 but got 8` |
| Filter bar has no decision filter | **TRUE** — `filters.ts` is exactly `{search, framework, owner, mode, trustMax}`; the prior ticket claim is correctly refuted |
| ADR-0026 is still `Proposed` | **TRUE** — `docs/src/adr/0026-open-dashboard-product-semantics.md:3` |
| `Decision` unchanged / no wire change | **TRUE** — still 5 members (`types.ts:3`); `pnpm generate:api` leaves the tree clean, zero drift |
| The `as Decision` cast is replaced by a guard that *rejects* | **TRUE** — `isOverridableDecision` parses via `.includes` over the 3-member list; an unrecognised value leaves state untouched rather than being coerced. Confirmed by test and by mutation |
| `MutationObserver` installs before first paint | **TRUE** — `addInitScript` runs at document-creation, before any app module; observing `document` (not `documentElement`) is the correct choice for that timing |

**Side effects — clean.** `OverrideRequest.decision` narrowing breaks no caller: the only
consumers are `override.ts`, `BulkActionBar.tsx`, `CapabilityPage.tsx` and
`api/capability.ts`, and narrowing a value flowing *into* the still-wide generated
`CapabilityOverrideRequest` is assignment-safe. `tsc --noEmit` clean.

**Gates re-run on the head commit — every number in the PR body reproduces exactly:**
`pnpm test` 254 files / 2625 tests passed · `pnpm type-check` clean · `pnpm lint` no issues ·
`pnpm generate:api` no drift · CI 16/16 green · SonarCloud **Quality Gate passed, 0 new
issues, 0 accepted, 0 hotspots, 100.0% coverage on new code, 0.0% duplication** · Codecov
all modified lines covered.

**Playwright**, repo-local `@playwright/test` re-pinned to **port 4601**: **4/4 passed**.
Mutation check on a correctly-rebuilt bundle: **4/4 fail** (`offered` gains `narrow`;
submitted decision is `narrow` not `deny`). Unit mutation check: **4 of the 5** new tests
fail. These tests are load-bearing, not assertions-around-the-claim.

> Method note for anyone repeating this: my first mutated e2e run *passed*, which is a
> false negative. `pnpm build` had failed on `TS6133` (unused import) and `vite preview`
> silently served the **stale** `dist`. Always confirm the bundle actually rebuilt before
> trusting a mutation result.

---

### Findings

**MAJOR — the new default turns an unconsidered click from a harmless no-op into a
successful bulk write, with no confirmation and no in-UI undo.**
Pre-fix, "select agents → Apply override" was a guaranteed 400: nothing was recorded.
Post-fix the identical click *succeeds* and records `deny` for every selected agent against
whichever resource happens to be first in the dropdown. A `select all agents` checkbox
exists, so the one-click blast radius is every visible agent. `BulkActionBar.tsx:69-75`
calls `onApply` directly — no confirmation step — and `src/api/capability.ts` exposes no
`removeOverride`/DELETE, so there is **no way to undo it from the UI** even though the
backend has `DELETE /capability/override`. The author's rationale weighs permissive vs
restrictive (`allow` vs `deny`) but never weighs *mutation vs no-op*, which is the axis
that actually changed. A placeholder default (`— select —`) would satisfy the ticket's
"primary path succeeds" intent while leaving Apply inert until an explicit choice — and the
new guard already rejects `''`, so the machinery for that is in this PR. Noting fairly that
`deny` is exactly what the ticket prescribed, so this may be accepted as-is — but it should
be an explicit decision, not an inherited one.

**MAJOR / PRODUCT-DECISION-REQUIRED — legend still advertises five states beside a control
that offers three; the PR body's justification is materially incomplete.**
The body says the legend is untouched "because ADR-0026 Decision 2 is `Proposed`". D2 *is*
`Proposed` — verified. But the body omits that D2's **recommended** option (A) explicitly
bundles both halves: "Remove `narrow`/`approval` from the legend **and** from the
bulk-action options, and re-default the bulk selector away from `narrow`" — and that option
(C), "keep the legend as aspirational", is **explicitly rejected** by the ADR as "the exact
pattern this programme exists to remove". This PR implements half of (A) unilaterally and
leaves the half the ADR names as the anti-pattern. The split is defensible on the merits
(the control fix stands on its own — a guaranteed 400 — while the legend is pure display),
but the framing presents ADR status as neutral cover when the ADR in fact recommends going
further. `bulk-options-light.png` in this PR's own evidence shows the legend swatches
(`allow`, `narrow`, …) rendering in the filter bar directly above the narrowed bar.
Recommend filing the legend follow-up now and linking it to D2.

**OUT-OF-SCOPE-FOLLOW-UP — the `NARROWED` summary tile reports a structurally impossible
metric as a real `0`.**
Visible in this PR's own screenshot and confirmed at `CapabilitySummary.tsx:57-61`: the tile
renders `narrow` as a plain count. Since the projection can never emit `narrow`, that tile
is always `0` — presenting an unreachable measurement as a legitimate business value, which
is precisely what the truthfulness rules forbid. The **adjacent** `FLAGGED AGENTS` tile
already does this correctly, rendering `NOT EVALUATED`. `narrowed` should use the same
`not-supported` / `not-evaluated` vocabulary. Same defect class as this ticket, in the same
component tree, and the "Deliberately not in scope" survey in the PR body missed it.

**OUT-OF-SCOPE-FOLLOW-UP — AAASM-5178 (authority overstatement) is correctly deferred, but
this PR increases exposure to it.** `capability.rs`'s module doc is explicit that overrides
"have never fed enforcement" — a display overlay. `Apply override` and the
`override applied to N agents` toast read as enforcement. By making the default path
*succeed* rather than 400, this PR raises the rate at which operators receive that success
message for an action that changes nothing the gateway enforces. Worth linking as
motivation to prioritise 5178.

**MINOR — the PR body overstates the mutation check.** It says "all five fail" under "the
pre-fix default and option list restored". Under exactly that mutation I get **4 of 5**:
`BulkActionBar` ×3 and `CapabilityPage` ×1. The fifth, `isOverridableDecision` in
`__tests__/override.test.ts`, is by construction independent of the bar's default and option
list — it can only fail if the guard itself is reverted, which is a different mutation. Not
a test defect; an imprecision in the claim.

**NIT — the `MutationObserver` assertion is sound but not load-bearing.** I confirmed it
installs before first paint and that the legend uses `className`, not `data-decision`, so it
is not confounded. But under the pre-fix mutation the test dies earlier — on the `offered`
equality and on `overrideBody().decision` — so `expect(seen).not.toContain('narrow')` is
never reached. Its real value is guarding a *future* regression that paints `narrow` without
changing the submitted decision. The body's "so a value that survived a single frame would
still be caught" is true, but the caught-ness is currently carried by the other assertions.

**NIT — rollback loses the selection.** `handleBulkApply` clears `selected` before the
await and restores only `optimistic` on failure, so a failed override forces the operator to
re-select. Pre-existing, not introduced here.

---

### Scope

Matches the ticket precisely. AAASM-5124 asks for `DECISION_OPTIONS = ['allow','deny','na']`,
default `'deny'`, and "a test that asserts the option set matches the values the API
accepts" — all three delivered, plus the type-level narrowing that makes the optimistic
paint structurally incapable of showing an impossible decision. Nothing beyond scope; the
`Decision` enum and the wire contract are correctly left alone. Commits are granular and
gitmoji-conventioned. Merge-ready on correctness once the MAJOR default question is settled.

— Claude Code

Chisanan232:
## Reworked since review

Three engineer decisions implemented on this same surface. The previous `deny`-default rationale is **superseded** and recorded as such in the PR body.

**1 — bulk action default.** No pre-selection; the primary control is disabled with its reason on it until a decision is chosen; the write is gated behind a confirmation naming the decision, the verb, the resource and the affected-agent count. Changing the decision or resource retracts a pending confirmation. This directly closes the review finding — an unconsidered click can no longer produce a successful bulk write.

**2 — legend (ADR-0026 D2).** Resolved in favour of the ADR's recommended option **(A)**: legend narrowed to `allow`/`deny`/`n/a`. The earlier body cited "D2 is Proposed" while omitting that D2's option (C) — keep the legend aspirational — is explicitly rejected in the ADR; that omission is corrected. **`unconfigured` was deliberately not added**: ADR-0024 proposes it, nothing emits it, and listing it would be the same defect inverted.

**Filter removal was a no-op, re-verified.** `CapabilityFilters` is `{search, framework, owner, mode, trustMax}` and the legend is a static non-interactive `<ul>`. There was nothing to remove — so rather than invent work, a test now pins the control list closed so a decision filter cannot be added later without re-deciding D2.

**3 — AAASM-5178.** Language and affordance only. Button → `Record display-only override`; the confirmation carries the disclosure at the point of action; the toast now reads `display-only override recorded for N agents — the dashboard annotation changed; gateway enforcement did not`. **No enforcement was wired** — that stays a separate task under its own ADR alongside ADR-0021.

**ADR action: `amend` ADR-0026.** An `Update — AAASM-5124 / AAASM-5178` section records the D2 sign-off, the 5178 disclosure decision, the deferral of enforcement wiring, and the accepted residual risk. Searched 0017 (Accepted), 0021, 0024, 0025, 0026 + `SECURITY.md`/`CONTRIBUTING.md`/`CLAUDE.md`/Jira first — **no conflict with any Accepted decision**; ADR-0017 in fact supports the narrowing.

### Two things called out rather than folded in

- **Undo is not cheap.** The gateway has `DELETE /capability/override`, but the FE client exposes no delete, the page has no revert path, and the response's `overrideId` is discarded — so there is no per-write handle to revert. That is a ticket, not a line. The confirmation step is the compensating control and the residual risk is recorded in the ADR rather than half-built here.
- **The summary row's `narrowed` counter** (`CapabilitySummary.tsx:61`) can only ever read `0` from this projection, for exactly the reason the legend entry was removed. It is not one of the three surfaces the decision enumerated, so it is reported, not silently changed.

### Verification

Every new behaviour was revert-checked — reverted in source, watched go red, restored with `git checkout HEAD -- <path>` (no `stash`): `deny` default → 4 red · bypass confirmation → 13 red · confirmation drops count/decision → 2 red · re-add `narrow`/`approval` to legend → 1 red · `deny` default rebuilt + e2e → 2 e2e red.

The e2e revert-check ran against a **rebuilt** `dist` with `rm -rf dist` first and the build exit code confirmed `0` — a failed build leaves `vite preview` serving a stale bundle, which makes a mutation check pass and falsely read as "the tests aren't load-bearing".

Gates, each once: `pnpm test` 2636 passed · `pnpm type-check` clean · `pnpm lint` clean · `pnpm generate:api` → `git status --porcelain` empty · Playwright 8 passed (4 runs × light/dark, console errors `[]`, port 4601).

Chisanan232:
### Final state

All checks green: **20 passed, 24 skipped, 0 failed**.

**SonarCloud: Quality Gate passed, 0 new issues**, 0 accepted, 0 security hotspots, 97.4% coverage on new code, 0.0% duplication. Sonar's first pass on this rework raised **one** MAJOR — `typescript:S6819` on a `role="group"` I had added to expose the confirmation's label to assistive tech. It was **fixed, not suppressed**: the panel is now a `<fieldset>`, which carries the implicit `group` role semantically (and matches how `features/alerts` already groups controls). No `NOSONAR`, no rule disables. UA border/margin/`min-inline-size` are reset in CSS so rendering is byte-identical — re-captured screenshots confirm it.

**Codecov patch green.** Locally, every changed component is fully covered — `BulkActionBar.tsx` **24/24 lines, 17/17 branches**; `CapabilityFilterBar.tsx` **16/16, 4/4**. That is real coverage, not padding: three of those branches exist only because I went looking for them (the resource-name fallback when a refetch drops the selected resource, and both render-nothing guards, which one test previously claimed to cover in its name but did not).

### Against the review's findings

| Finding | Status |
|---|---|
| MAJOR — unconsidered click became a successful bulk write, no confirmation, no undo | **Resolved.** No default, control disabled with its reason, confirmation naming decision + count. The reviewer's own suggested shape (placeholder default leaving Apply inert) is what shipped. |
| MAJOR — legend advertises five states; body's ADR framing incomplete | **Resolved.** Legend narrowed; ADR-0026 D2 signed off as option (A) and amended; the omission about option (C) being explicitly rejected is corrected in the body. |
| FOLLOW-UP — AAASM-5178 exposure increased by making the default path succeed | **Resolved on this surface.** Label, confirmation disclosure and toast all reworded; enforcement wiring explicitly deferred. |
| MINOR — body overstated the mutation check ("all five fail") | **Corrected.** The body now gives exact per-mutation counts, each re-derived. |
| NIT — `MutationObserver` assertion not load-bearing under the old mutation | **Acknowledged, unchanged.** Still true, still worth keeping as a guard against a future regression that paints an impossible state without changing the submitted decision. The body no longer leans on it. |
| NIT — rollback loses the selection | **Unchanged.** Pre-existing, outside the three decisions, not introduced here. |

### Still open, deliberately

- **No undo affordance.** Not cheap — see the PR body. Confirmation is the compensating control; residual risk recorded in ADR-0026.
- **The summary row's `narrowed` counter** can only ever read `0` from this projection. Same defect as the legend entry, but not one of the three surfaces the decision enumerated — reported, not silently folded in.
- **`dashboard/verify/5124/bulk-options-{light,dark}.png`** are orphaned by a renamed run and show the superseded pre-selected bar. Left in place rather than deleted without sign-off — worth removing when you next touch this.

Not approving, not merging.

PR #1720, opened 2026-07-26
[AAASM-5180] 🐛 (dashboard): Fail closed when no AuthContext is mounted, and test the disabled branch

Chisanan232:
## Description

`dashboard/src/auth/usePermissions.ts` resolved the caller's scopes with
`return ctx ? ctx.scopes : ALL_SCOPES`, where `ALL_SCOPES = ['read', 'write', 'admin']`.
**With no `AuthContext` mounted, every scope was granted.** That path is unreachable
in the app — `main.tsx` is the sole entrypoint and wraps the whole tree in
`<AuthProvider>` — but it was reachable in every spec that rendered a gated control
bare. Those specs silently exercised the fully-permissive branch, so their RBAC
assertions passed vacuously: **deleting a gate entirely kept CI green.** Found during
the AAASM-5148 review (#1717), where that lane's RBAC gap had survived for exactly
this reason.

**Part 1 — fail closed.** `useScopes()` now returns no scopes when no provider is
mounted, making the unprovided case the safe one and forcing a spec to name the
caller it acts as. A new `GrantScopes` test helper (plus `WRITE_SCOPES`) is that
declaration, in one place instead of re-hand-rolled per spec.

Fail-closed was chosen over the "keep the default, add a test-only guard" option
deliberately:

- **No non-test consumer relies on the permissive default.** `src/main.tsx` is the
  only `createRoot` call and it mounts `<AuthProvider>` above everything, so the
  fallback protected nothing real. All 30 `usePermissions`/`useCan` call sites render
  beneath it.
- **The guard option cannot meet the requirement.** A guard that throws whenever a
  gated component renders without a provider has *exactly* the same blast radius as
  failing closed (the same 15 specs break), while adding test-only machinery to
  production code. A guard that is opt-in is worthless — the vacuous RBAC spec is
  precisely the one that would not opt in.
- Failing closed matches the posture the codebase already documents elsewhere: a
  surface whose provider wiring regresses now shows disabled controls rather than
  silently granting `admin`.

15 specs were relying on the old default to reach their enabled branch; each had its
shared render helper wrapped in `GrantScopes`. No behaviour in those specs changed —
they just stop depending on a default that no longer exists.

**Part 2 — RBAC coverage for the five uncovered surfaces**, each asserting both the
enabled and the disabled branch, and — as #1717 did — **the absence of the network
write**, not merely a `disabled` attribute. A disabled attribute proves the button
looks right; the absent request proves the gate works.

| Surface | New spec | Write asserted absent |
|---|---|---|
| Fleet bulk suspend / resume | `pages/FleetPage.rbac.test.tsx` | `api.POST` never called; suspend reason dialog never opens |
| Policy write | `pages/PoliciesPage.rbac.test.tsx` | editor overlay never opens; `api.POST` never called |
| API key rotate / revoke | `features/iam/ApiKeyList.rbac.test.tsx` | confirm dialog never opens; `api.POST` never called |
| Destination test / write | `features/alerts/DestinationManager.rbac.test.tsx` | no create/update/delete/test mutation invoked |
| Alert rule write | `features/alerts/AlertRulesTable.rbac.test.tsx` | no non-`GET` `fetch` leaves the component; form callbacks never fire |

Each also covers `admin` satisfying the `write` requirement.

> **Note for the parallel dashboard lanes:** `pages/PoliciesPage.test.tsx` is touched,
> but only to add `GrantScopes` to its single shared `Wrapper` (a 3-line addition at
> the top of the file). No test bodies or production files in that area were changed.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

No runtime behaviour change in the app: the fallback being removed was unreachable
in production, because `main.tsx` always mounts `AuthProvider`.

## Related Issues

- Related Jira ticket: [AAASM-5180](https://lightning-dust-mite.atlassian.net/browse/AAASM-5180)
- Follows up the RBAC gap found during AAASM-5148 (#1717)

## Testing

Every new spec was **verified by reverting the gate it covers** — deleting the
`disabled={!canWrite}` / `title={...}` pair from the production component — and
confirming the spec fails. All five failed with an assertion about the write, not
about styling:

| Gate deleted | Spec | Result |
|---|---|---|
| Fleet bulk suspend + resume | `FleetPage.rbac.test.tsx` | ✗ `expected "POST" to not be called at all, but actually been called 2 times`; `expected <textarea/> to be null` |
| Policies new-policy (header + empty state) | `PoliciesPage.rbac.test.tsx` | ✗ `expected <div class="editor">…</div> to be null` |
| API key rotate + revoke | `ApiKeyList.rbac.test.tsx` | ✗ confirm dialog `expected <div>…</div> to be null` (both rotate and revoke) |
| Destination test / edit / delete / submit | `DestinationManager.rbac.test.tsx` | ✗ `expected 3 to be +0` (three mutations issued); `expected <button/> to be null` |
| Alert rules create / edit / delete | `AlertRulesTable.rbac.test.tsx` | ✗ `expected [ {…} ] to have a length of +0 but got 1` (a DELETE reached the wire); `expected "vi.fn()" to not be called at all` |

Gate output (each run once, from `dashboard/`):

```
$ pnpm test
 Test Files  258 passed (258)
      Tests  2647 passed (2647)

$ pnpm type-check
> tsc --noEmit
(exit 0, no output)

$ pnpm lint
ESLint: No issues found

$ pnpm generate:api && git status --porcelain
(empty — no codegen drift)
```

## Checklist

- [ ] All CI checks passing

No Rust crate is touched by this PR, so `cargo fmt` / `clippy` are unaffected.

[AAASM-5180]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Adversarial review — AAASM-5180 (RBAC fail-closed)

Reviewed independently; I did not write this PR. **Verdict: ready to merge.** No BLOCKER.
Two MINOR findings and two NITs below — none block, one is a correction to a claim in the
PR description rather than to the code.

### Gates

| Gate | Result |
|---|---|
| CI | 16/16 green, 0 failed |
| SonarCloud | Quality Gate **passed** — 0 new issues, 0 accepted, 0 security hotspots, **100.0% coverage on new code**, 0.0% duplication |

### The load-bearing claim, re-derived independently

The PR rests on *"the permissive fallback was unreachable in production."* I did not take
this on trust — I swept every mount path rather than only the ones the description names:

- **Exactly one React root**: `createPortal`-free `createRoot` appears once, `src/main.tsx:11`.
  No `ReactDOM.render`, no `hydrateRoot`, no `renderToString`/`renderToStaticMarkup`, no
  `customElements.define`, no `attachShadow`, no `.mount(`, no MSW anywhere in the repo.
- **`<AuthProvider>` is unconditional** (`main.tsx:14`) — not inside a ternary, `&&` guard,
  `<Suspense>`, error-boundary fallback, or `<Route>`. It sits *above* `<BrowserRouter>`
  (`App.tsx:34`) and above the `ErrorBoundary` in `AppShell.tsx:64`, so a caught render error
  cannot detach the auth context.
- **No second entry point**: `vite.config.ts` has no `build.rollupOptions.input` override;
  `index.html` has exactly one React script tag (`/src/main.tsx`) plus a plain
  `theme-init.js`. No other `*.html`, no iframe, no nested `package.json`.
- **Not mountable from outside**: `dashboard/package.json` is `"private": true` with no
  `main`, `module`, `exports`, or `bin`.
- **All four `createPortal` sites** (`OverlayHost`, `Drawer`, `ConfirmDialog`,
  `PolicySimulatePanel`) render from within the single root, and `createPortal` preserves
  React context — none escapes the provider.

One reinforcing detail the description does not mention, which further bounds the blast
radius: `parseScopesFromJwt` already returns `[]` for a missing or malformed token, so the
*provider-present-but-unauthenticated* case was **already** fail-closed before this PR. The
only thing this change touches is the no-provider case.

**Claim 1 confirmed.** The permissive fallback was genuinely unreachable in the production
browser path.

### Scope vs ticket

Covered. I read all five new `*.rbac.test.tsx` specs in full, specifically checking the
failure mode where a spec asserts only `toBeDisabled()` on a button whose handler still
fires. **None of them do that.** Every one asserts the absence of the write itself —
`api.POST` never called (Fleet, Policies, ApiKeyList), zero non-`GET` `fetch` calls
(AlertRulesTable), zero mutation-hook invocations (DestinationManager) — *and* each pairs
that with a positive control that drives the same action as a `write` caller and asserts the
request does reach the wire. That pairing is what stops the negative assertions being
vacuous, and it is the right shape.

### Side effects

**(a) No production surface loses functionality** — verified by the sweep above and by the
Playwright run below, not by inference.

**(b) The 15 modified specs were not weakened.** I diffed every one. Each change is an import
plus a wrapper insertion around an existing shared render helper. **Zero test bodies changed,
zero assertions deleted, zero assertions relaxed.** One change is genuinely semantic and worth
naming, since the description glosses it as "no behaviour changed":

`features/liveOps/RowActionMenu.test.tsx` — `setup()` went from
`overrides?.scopes ? withScopes(...) : menu` (bare render → previously *all* scopes,
including `admin`) to `withScopes(overrides?.scopes ?? ['read', 'write'], menu)`. The default
caller therefore silently loses `admin`. That is safe here only because **`canAdmin` gates no
production control anywhere** — it is defined at `usePermissions.ts:58,75` and has zero UI
consumers (`useCan` likewise has zero production call sites). Same reasoning clears
`WRITE_SCOPES = ['read','write']` for the other 14 specs. Worth stating explicitly, because
if an `admin`-only gate is ever added, these specs will quietly stop covering its enabled branch.

**(c) No test-only primitive can reach a production bundle.** `GrantScopes` and `WRITE_SCOPES`
have **zero importers outside `*.test.tsx`**. Vite bundles only from the entry graph
(`index.html` → `main.tsx`), so an unreferenced module in `src/` is never emitted —
tree-shaking is not even load-bearing here, reachability is.

### Frontend validation

Wrote a review-only spec driving the **built** app through the **real** `AuthProvider` (no
test helper in the tree), repo-local `@playwright/test` on port 4603, API intercepted with
`page.route` and the token seeded via `addInitScript` into `sessionStorage.aa_token`.
**2/2 passed.**

- An operator holding `read`+`write` still sees every gated control **enabled** — Policies
  `+ new policy`, Fleet bulk suspend/resume, Alerts `open rule form`. This is the check that
  matters: fail-closed did **not** break the happy path.
- The control case: a `read`-only operator sees those same controls disabled, the
  read-entitled `Destinations` opener stays enabled (the gate does not over-reach), and
  **zero mutating requests reached the wire**.
- No console errors or uncaught exceptions in either pass.

Screenshots: `dashboard/verify/5180/` — `write-{policies,fleet-bulkbar,alerts}-enabled.png`,
`read-{policies,fleet-bulkbar,alerts}-disabled.png`.

---

### Findings

**MINOR — Storybook now renders gated controls permanently disabled, and the PR's stated
justification is slightly wrong about this.**
The description argues *"No non-test consumer relies on the permissive default."* Production
does not — but **Storybook is a non-test consumer that did.** `.storybook/preview.ts` defines
no `decorators` at all, and the four stories files rendering `usePermissions` consumers
(`ApprovalActions.stories.tsx`, `RowActionMenu.stories.tsx`, `ApprovalDetailDrawer.stories.tsx`,
`ApprovalPool.stories.tsx`) wrap only `QueryClientProvider`/`MemoryRouter`. Before this PR they
resolved to full `admin`; after it they resolve to no scopes, so approve/reject, the row action
menu, and the drawer footer will all render disabled in Storybook. No production impact, but
this repo actively uses Storybook for design-fidelity review, so it degrades a real surface.
Fix is one global decorator in `.storybook/preview.tsx`. Reasonable as a follow-up rather than
a change to this PR.

**MINOR — `GrantScopes` is a scope-granting primitive sitting in the production source tree
with no guard.** It lives at `src/auth/GrantScopes.tsx`, not under a test-only path, and
nothing prevents a future import from production code — where it would override `AuthProvider`
for its subtree with a hardcoded `'test-token'`. Currently harmless (zero production
importers, unreachable from the bundle graph), but a `no-restricted-imports` ESLint rule
scoped to non-test files, or relocating it beside the specs, would make that structural rather
than incidental.

**NIT — `WRITE_SCOPES: Scope[]` is a mutable exported array.** It is module-level and shared
across all 15 specs, so a spec that mutated it would corrupt the others. The same PR gets this
right one file over: `NO_SCOPES` is `readonly Scope[]` + `Object.freeze([])`. Worth matching.

**NIT — `GrantScopes` stubs `login` as `async () => {}`.** Fine as documented, but a spec that
ever exercises a login path would silently observe a no-op success rather than a failure. Not
reachable today.

---

Confirmed clean against the rest of the wave: this PR's touch of `pages/PoliciesPage.test.tsx`
is confined to the shared `Wrapper` at the top of the file, and
`components/trace/ApprovalDetailDrawer.test.tsx` is touched by no other open PR.

Not approving and not merging — this is a review record only.

— Claude Code

PR #1721, opened 2026-07-26
[AAASM-5109] 🐛 (dashboard): Stop Trace calling a route that does not exist

Chisanan232:
## Description

The Trace view called `/api/v1/agents/{agentId}/sessions/{sessionId}/trace`. **That route does not exist.**

Evidence from `openapi/v1.yaml`:

| Fact | Evidence |
|---|---|
| The called path is absent from the schema | No `/agents/{id}/sessions/…` path exists anywhere in `openapi/v1.yaml`; the only `sessions` path is `/api/v1/fleet/active-sessions` (line 2064) |
| A real trace route **does** exist | `/api/v1/traces/{session_id}` at line 3170, `operationId: get_trace`, tag `traces`, returning `TraceResponse` |
| It is already in the generated client | `src/api/generated/schema.d.ts:1842` (path) and `:8252` (`get_trace` operation) — no `pnpm generate:api` drift |

So every request the surface made 404'd, and the only state it could reach in production was its error branch. Because a real endpoint exists, the fix is to call it — no backend was fabricated, and no affordance was invented.

The wire shape is also incompatible, and that gap is most of this change. `TraceSpan` (`openapi/v1.yaml:7951`) carries **six** fields — `span_id`, `parent_span_id`, `operation`, `decision`, `start_time`, `end_time` — while the view expected `durationMs` / `payloadPreview` / `severity` / `redactedFields` / `violationReason`.

**Mapped from real data:** `operation` → type, `start_time` → timestamp, `end_time - start_time` → duration, `decision` → verdict, `parent_span_id` → nesting (indentation — see below), envelope `agent_id` → agent.

**Nesting.** `parent_span_id` is rendered, not merely carried: `features/trace/nesting.ts` walks each span's parent chain and `TraceTimeline` offsets the row by the resulting depth. Row *order* is untouched — the gateway's `start_time` sort still decides sequence; indentation only shows lineage. Two malformed-ancestry cases are handled deliberately, because both are reachable from a well-behaved gateway: a parent outside the response (`build_trace_from_audit` scans a bounded 10 000-entry window, so chains can be truncated) keeps the hops that did resolve, and a cycle roots the span rather than inventing a depth to break the loop. `data-depth` carries the true depth however deep the chain runs; only the drawn offset is clamped (`MAX_INDENT_DEPTH`), so the attribute stays a fact about the data rather than a description of the layout.

**No wire source → `not-supported`** (AAASM-5100): `payload`, `payloadPreview`, `severity`, `redactedFields`, `violationReason`. These render the shared em-dash, never a fabricated value.

Three further honesty defects surfaced while mapping against the real contract:

1. **`deriveVerdict` fabricated `ALLOWED`.** It ended in `return 'allowed'` while testing snake_case names (`policy_violation`, `credential_leak`) that the API never emits — `operation` is `AuditEventType::as_str()`, i.e. PascalCase `PolicyViolation`. Against the real wire *every* span, including a recorded policy violation, would have worn a green ✓ ALLOWED chip. It now returns `Certain<Verdict>` with **no default**, reading the explicit `decision` first and then only those audit event types whose own documentation states an outcome. A stringified-integer `decision` (what the audit-reconstruction path emits) is deliberately **not** interpreted — the schema documents no encoding for it, and guessing backwards would report a denial as a pass.
2. **The error branch was a bare red `<p>`.** Now the shared `StatusState` as `unavailable`, so an unreadable trace cannot be skimmed as an empty one.
3. **The severity filter had no production path.** With no severity on the wire, three of its four checkboxes can never match a row — unchecking "Critical" to no effect reads as "there are no critical events". It is hidden until a severity is actually present, so it returns on its own when AAASM-5100 lands.

### AAASM-5165

`RedactionPreview` did `JSON.stringify(payload, null, 2) ?? 'null'`, reaching the literal string `"null"` by two routes: `stringify(undefined)` returns the *value* `undefined` so the `??` fired, and `stringify(null)` returns the four-character *string* directly. Extracted to `previewJson()`, which returns `Certain<string>`; the body now reads **"no payload recorded"** with a labelled absence marker. After AAASM-5109 this matters more, not less: `payload` has no wire source at all, so `"null"` would have been the *only* thing the preview ever rendered.

**On the ticket's `NaN` half — it was genuinely unreachable, and this PR is what makes it real.** `durationMs` was a required `number`, and with the route 404-ing the timeline never rendered at all, so neither `null ms` nor `NaN ms` could reach a production screen. Mapping `end_time - start_time` creates the branch for the first time: `end_time` is nullable, and `aa-api/src/routes/traces.rs` leaves it `None` on *every* reconstructed span. The guard therefore lives in the mapper (`spanDurationMs`), where it returns an absence for a null, unparseable, or inverted end time — a measured `0` stays a fact.

Fields the wire cannot supply are typed `Certain<T>`, not `T | undefined`, so no component can interpolate an absence into a template. Vocabulary is imported from `lib/truthfulness`; nothing was redefined.

**ADR-0017 items 11–13 respected** — the generic `RedactionPreview` (item 12: per-type templates would fabricate data), the single `TraceDrawer` container (item 11) and the seven-state `LayerSteps` renderer (item 13) are all preserved. `LayerStep.status` became `Certain<LayerStatus>` rather than gaining an eighth state, so the ratified seven-state vocabulary is intact.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [ ] No
- [x] Yes (describe below)

Frontend-internal only; no public API or backend change.

- `useTraceQuery(agentId, sessionId)` → `useTraceQuery(sessionId)`. The trace is keyed by session alone; the agent is an *output* of the call.
- `TraceEvent`'s possibly-absent fields are now `Certain<T>`.
- The trace JSON export is bumped to **version `2`**: possibly-absent fields serialize as the self-describing `Certain` envelope. A v1 export could not tell "took 0 ms" from "never measured" — the same ambiguity the UI was fixed for, written into the file operators download as evidence.

## Related Issues

- Related Jira ticket: [AAASM-5109](https://lightning-dust-mite.atlassian.net/browse/AAASM-5109), [AAASM-5165](https://lightning-dust-mite.atlassian.net/browse/AAASM-5165)
- Related GitHub issues: N/A

## Testing

`pnpm test` → **253 files, 2675 tests passed**. `pnpm type-check` and `pnpm lint` clean. `pnpm generate:api` produces no diff.

New coverage targets real branches:

- The route literal is pinned and asserted **not** to contain `/sessions/`, so a drift back to the non-existent path fails a test instead of 404-ing in production.
- `spanDurationMs`: measured, measured-zero, null / omitted / empty `end_time`, unparseable timestamp (the NaN source), inverted range.
- `deriveVerdict`: explicit decision words, case/whitespace, **numeric decision refused**, each operation-derived verdict, and that `ToolCallIntercepted` et al. report `not-evaluated` rather than defaulting to allowed.
- `previewJson`: absent payload, JSON `null`, `undefined`, a value with no JSON representation, and absence propagation — the literal `"null"` is asserted absent from the rendered body.
- Export v2 distinguishes `{known:true,value:0}` from `{known:false,state:…}`.

Playwright review pass — `tests/e2e/review-aaasm-5109.spec.ts` (config `review-5109.config.ts`, port 4591): **14/14 passed**, 7 scenarios × light and dark, `expect(harness.errors).toEqual([])` on every test. Failure paths are driven explicitly via `page.route` (auth seeded through `addInitScript`, since `openapi-fetch` captures `globalThis.fetch` at module load). Screenshots committed to `dashboard/verify/5109/`.

> Note: port 4591 is shared with `review-5090.config.ts`. The spec asserts on test ids introduced by this branch, so a stale preview server fails the run loudly rather than yielding false-green evidence.

## Checklist

- [x] Code follows project style guidelines (`cargo fmt`, `cargo clippy`) — no Rust touched; `pnpm lint` clean
- [ ] Commits are signed off (`git commit -s`) — optional/advisory

https://claude.ai/code/session_01MfZAdbqTbJCCYhyzjDTGTB

[AAASM-5109]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Adversarial review — PR #1721 (AAASM-5109 + AAASM-5165)

Independent review. I did not write this PR and verified every load-bearing claim against the
source rather than relaying the PR body. **Not approving, not merging.**

**Verdict: not ready — one MAJOR to resolve. No BLOCKER.**
The route repoint and the verdict rewrite are correct, in scope, and well evidenced. The single
thing standing in the way is a PR-body claim the code does not support.

---

### Route claims — verified independently, both true

| Claim | Result |
|---|---|
| `/api/v1/agents/{agentId}/sessions/{sessionId}/trace` absent from `openapi/v1.yaml` | **Confirmed.** Zero matches on `main`; the only trace path in the whole spec is line 3170. |
| `/api/v1/traces/{session_id}` exists, `operationId: get_trace` | **Confirmed** at `openapi/v1.yaml:3170`, tag `traces`, backed by `aa-api/src/routes/traces.rs::get_trace`. |
| Response schema compatible with the repoint | **Confirmed.** `TraceResponse{session_id, agent_id, spans[]}`, `TraceSpan{span_id, parent_span_id, operation, decision, start_time, end_time}`. `api.ts` maps all six. No schema mismatch — this is not a repoint to a differently-shaped route. |

Supporting backend facts I re-derived rather than took on trust:
- `decision` really is a **stringified integer** — `payload.get("decision").and_then(|v| v.as_i64()).map(|d| d.to_string())` (`traces.rs`). Refusing to interpret it is correct: `openapi/v1.yaml` types it as a bare nullable string with no enum, so `"0"` has no published meaning.
- `end_time: None` on **every** audit-reconstructed span (`traces.rs`, `build_trace_from_audit`). The absent duration is the ordinary case, not an edge case.
- `operation` is `AuditEventType::as_str()` (`aa-core/src/audit.rs:132-157`) — all 22 variants PascalCase.

---

### The `deriveVerdict` green-chip claim — VERIFIED TRUE, and understated

This was the most serious item in the wave, so I tested it rather than reasoned about it. I ran the
old deriver **verbatim** from `main` against real wire operation names:

```
PolicyViolation              -> allowed
CredentialLeakBlocked        -> allowed
MessageBlocked               -> allowed
ApprovalDenied               -> allowed
ToolCallIntercepted          -> allowed
A2AImpersonationAttempted    -> allowed
BudgetLimitExceeded          -> allowed
```

The claim holds, and the real cause is worse than "snake_case vs PascalCase". The old deriver read
`event.type`, `event.redactedFields` and `event.violationReason` — **none of those three fields
exists on `TraceSpan` at all**. So the fall-through to `return 'allowed'` was not merely likely, it
was unconditional. Every span in every trace, including recorded policy violations, credential-leak
blocks and denied approvals, would have carried a green ✓ ALLOWED chip.

Returning `Certain<Verdict>` with no default is the right fix, and declining to guess the
stringified-integer encoding is the right call — getting it backwards would report a denial as a pass.

---

### Scope — the "went well beyond its tickets" premise does not survive reading the tickets

I pulled both tickets. AAASM-5109's own *Target behaviour* section explicitly requires:

> "`decision` -> verdict", "`parent_span_id` -> nesting", "Swap the error branch to the shared
> ErrorState component", "Fields with no wire source … must render an em-dash, not a fabricated value."

So of the three "extra" fixes:

1. **`deriveVerdict` rewrite — IN SCOPE.** The ticket names `decision -> verdict` as a deliverable. Not creep.
2. **Severity filter hidden — the one debatable item.** The ticket says severity must "render an em-dash"; hiding the control is a stronger reading. Defensible (a filter whose 3-of-4 checkboxes can never match is a false affordance — an operator unchecking "Critical" and seeing nothing change concludes there are no critical events), and it self-restores when the backend supplies severity. But this is the item a reviewer could fairly ask to split. Calling it **PRODUCT-DECISION-REQUIRED**, not creep.
3. **Prototype-pollution Map conversion — IN SCOPE by definition.** The lookup tables are new/rewritten in this PR; the author fixed a bug their own change introduced.

26 source files (13 source + 11 test + 2 e2e) is proportionate — the `T | undefined` → `Certain<T>`
retype ripples mechanically into every consumer and its test. No backend files touched.

---

## Findings

### MAJOR

**M1 — `parent_span_id → nesting` is claimed but not delivered.**
The PR body states under *Mapped from real data*: "`parent_span_id` → nesting". AAASM-5109 lists it as
an explicit target-behaviour item. What the code actually does:

- `api.ts:83` maps `parentSpanId: span.parent_span_id ?? null` ✅
- `types.ts:32` and `exportSchema.ts:48` carry it ✅
- `TraceTimeline.tsx:170` renders a **flat `events.map`** — no depth computed, no indent, no parent
  linkage, and `parentSpanId` is never passed to `TraceEventRow`. No `indent`/`depth`/`nest` anywhere
  in the component or its CSS.

The field is *carried*, not *rendered as nesting*. This is a true observation ("parent_span_id is
mapped") with a conclusion attached that the diff does not support — exactly the failure mode this
review programme exists to catch. Either render the nesting, or amend the PR body and file the
follow-up. Not a correctness or safety bug — nothing false is displayed — but a stated deliverable
that isn't there.

### MINOR

**M2 — `VERDICT_BY_OPERATION` applies its own stated rule inconsistently.**
The doc comment says only variants "whose own documentation states the outcome" are mapped. But
`A2AImpersonationAttempted`, whose doc reads *"The action is denied before policy evaluation runs"*,
is **not** mapped — nor are the sandbox `*Blocked` variants. A recorded impersonation denial
therefore renders as `not-evaluated` rather than DENIED. This is the **fail-safe** direction (an
absence marker, never a false ALLOWED), so it is not a blocker — but either map them or narrow the
doc claim so the rule matches the table.

**M3 — the deliberate e2e port collision has inverted reasoning.**
`review-5109.config.ts` pins 4591, colliding with `review-5090.config.ts`, justified as making a
stale server "fail loudly rather than producing green evidence". With `reuseExistingServer: true` a
colliding server is **reused, not rejected** — 5109 fails only as a side effect of asserting on
test-ids this branch introduces. The reverse case is unprotected: with a 5109 server up, a 5090 run
tests the 5109 build, which is a superset, and can go **false green**. A unique port plus
`reuseExistingServer: false` produces the loud failure deterministically and in both directions.
The instinct is right; the mechanism doesn't implement it.

### NIT

**N1** — AAASM-5165 asked for "a single `formatDurationMs(ms)` helper" at three call sites; delivered
instead as `Certain<number>` + `TruthfulValue`. Stronger (absence becomes unrepresentable rather than
formatted late) but a deviation from the ticket's stated implementation — worth a ticket note.

**N2** — `LayerStatus` `'skip'` is now unproducible (`l3Status` returns only `unreached` or an
absence). ADR-0017 item 13 is about the *renderer* handling 7 states, which still holds, so the ADR
is not violated — but the state is dead until the span schema widens.

**N3** — Sonar's 3 new issues are all `EmptyState` deprecation warnings on lines this PR **did not
change** (`TraceViewPage.tsx` import + 117 + 137), surfaced only because the file is in the diff.
`EmptyState` is a thin wrapper over `StatusState` with identical props and test-id, and "genuinely
empty" is the correct semantic at both sites. Leaving them is defensible.

**N4** — `exportSchema` bumps `'1'` → `'2'` with no migration branch, though the file's own comment
prescribes "bump the literal **and add a migration branch**". Harmless today — nothing in the repo
parses a trace export — but worth honouring when an importer lands.

### OUT-OF-SCOPE-FOLLOW-UP

**F1 — the identical prototype-pollution bug survives one directory over.**
The sweep is complete *within the diff* — I checked every lookup in the changed files.
`ICON_BY_TYPE`, `VERDICT_BY_DECISION` and `VERDICT_BY_OPERATION` are all `Map`s; `STATUS_META`,
`VERDICT_META` and `L2_STATUS_BY_VERDICT` remain object literals but are keyed by **closed literal
unions** (`LayerStatus` / `Verdict`), never by wire strings — safe.

But untouched by this PR, in the same `components/trace/` directory:

```ts
// ApprovalDetailDrawer.tsx:34
const STATUS_VERDICT: Record<string, Verdict> = { pending, approved, rejected }
// :41
return STATUS_VERDICT[status] ?? 'pending'   // status = wire `approval.status`
```

`status === 'constructor'` resolves `Object` off the prototype, which is not `undefined`, so `??`
never fires and a **function** is returned as a `Verdict` — `VERDICT_META[Object]` is then
`undefined` and `.label` throws. It is also a silent default for unrecognised input. Note ADR-0017
item 11 records this drawer as **superseded** by the generic `TraceDrawer`, and it has no production
mount (only a `.stories.tsx`), so deleting it may be the right fix. Worth a ticket either way.

---

### Side-effect checks

**(a) `LayerStep.status` → `Certain<LayerStatus>` — ADR preserved, no eighth state.**
`docs/` is untouched by this PR (empty diff), so ADR-0017 items 11-13 cannot have been reverted; I
re-read them to confirm they still describe shipped behaviour. Item 11 (single generic `TraceDrawer`)
— `TraceDrawer` not in the diff. Item 12 (generic `RedactionPreview`) — still type-agnostic, no
per-type semantic templates added. Item 13 (7-state renderer) — `LayerStatus` is still **exactly**
`pass | fail | pending | narrow | scrub | skip | unreached`, and `STATUS_META` still has 7 entries.
The absence lives in the `Certain<>` wrapper, not in a new union member. **No eighth state.**

**(b) Export artifact shape — it does change, and that is handled.**
Every possibly-absent field becomes a `{known:true,value}` / `{known:false,state,detail}` envelope,
version `'1'` → `'2'`. I searched for consumers: **nothing in the repo parses a trace export** except
`exportSchema.test.ts`, and no doc under `docs/src/` documents the format. So no consumer and no
documented contract breaks. The version literal is the file's own documented evolution mechanism and
was correctly bumped (see N4 on the missing migration branch).

**(c) Narrowing — verified at every consumer.**
`TraceTimeline`, `LayerSteps`, `DecisionExplainer`, `PayloadModal`, `RedactionPreview` and
`TraceViewPage` all guard with `isKnown(...)` before reading `.value`. `LayerSteps` splits the glyph
into its own component so each branch narrows without a non-null assertion; `DecisionExplainer`
guards both `meta` and the inline `VERDICT_META[verdict.value].label`. No unguarded `.value` found.

---

### Validation run locally

| Check | Result |
|---|---|
| CI (`gh pr checks 1721`) | **16 passed / 0 failed** — settled green, no fix needed |
| SonarCloud | **Quality Gate PASSED** · 3 new issues (all MINOR, all `EmptyState` deprecation, see N3) · 0 Security Hotspots · **98.3%** coverage on new code · 0.0% duplication |
| Codecov | Patch **96.34%**, 3 partial lines (`PayloadModal` 66.7%, `TraceTimeline` 85.7%, `TraceViewPage` 87.5%) |
| `pnpm type-check` | clean |
| `pnpm lint` | clean (`--max-warnings 0`) |
| `pnpm test` | **2679 passed / 2679**, 253 files |
| `@playwright/test` review specs | **14 passed / 14** |

**Playwright:** run with repo-local `@playwright/test` (not the MCP server). As instructed I moved
off the colliding 4591 to **port 4604** and set `reuseExistingServer: false`, via a temporary copy of
`review-5109.config.ts` which I removed afterwards — the worktree is back to clean and the author's
committed screenshots are untouched. Evidence at `dashboard/verify/5109/` (light + dark for
`route-called`, `unmeasured-duration`, `payload-absent`, `verdict-absent`, `trace-unavailable`,
`trace-empty`, `no-severity-filter`).

**Do the specs prove the claims or assert around them?** They prove them. The specs drive the wire
with `page.route` at production shape (`end_time: null`, `decision: null`, PascalCase `operation`),
and critically they assert the **negative**: `harness.requested` must contain
`/api/v1/traces/{session_id}` *and* must contain no path with `/sessions/` or ending `/trace`. That
is a real test of AAASM-5109 rather than a re-assertion of the new code's own behaviour. The duration
test includes one genuinely measured span (`834 ms`) alongside three unmeasured, so it proves the
guards did not simply blank the column — the most common way this class of fix fakes a pass. The
outage test waits out React Query's real 3-retry backoff rather than reconfiguring the app to fail
faster. I did find one soft spot: the empty-session and failed-request cases are distinguished by
test-id, which is a weaker guarantee than distinguishing the rendered text, but `StatusState` carries
`role="alert"` so the distinction is real.

---

### Summary

No BLOCKER. The route repoint is correct and schema-compatible, the green-chip claim is true and
materially serious, and the truthfulness treatment is applied consistently with no permissive
defaults introduced. Resolve **M1** (render the nesting, or correct the PR body and file the
follow-up), decide on **M2/M3**, and file **F1**. Then this is good to merge.

— Claude Code

Chisanan232:
## Follow-up — the MAJOR from my review is fixed, plus Sonar/Codecov to zero

Continuation of [my review above](https://github.com/ai-agent-assembly/agent-assembly/pull/1721#issuecomment-5086094044). Eleven commits pushed (`23620fe6..a7d0369e`). Still **not approving and not merging** — this is a record of what changed.

**Verdict: ready.** No BLOCKER, no outstanding MAJOR.

---

### 1. M1 (MAJOR) — `parent_span_id` → nesting is now actually rendered

My review found the PR body claimed this and the code only carried the field. Rather than correct the claim, the ticket asks for the behaviour, so it is now implemented.

`features/trace/nesting.ts` walks each span's parent chain; `TraceTimeline` offsets the row by the resulting depth. Order is untouched — the gateway's `start_time` sort still decides sequence, indentation only shows lineage.

Both malformed-ancestry hazards are handled, and neither is hypothetical:

| Case | Behaviour | Why it's reachable |
|---|---|---|
| Parent not in the response | Keeps the hops that resolved; the walk stops where the evidence stops | `build_trace_from_audit` scans a bounded 10 000-entry window, so a chain can be truncated mid-way |
| Cycle (`A→B→A`, or self-parent) | Span renders at the root | Nothing on the wire validates against one |

Every walk is bounded by a `seen` set that cannot grow past the span count, so neither case can recurse or hang. `data-depth` reports the true depth however deep the chain runs; only the *drawn* offset is clamped (`MAX_INDENT_DEPTH`), so the attribute stays a fact about the data rather than a description of the layout.

**Tests — 18 new.** 12 unit cases on the deriver (two-span cycle, three-span cycle, self-parent, a span hanging off a cycle, truncated chain, out-of-order arrival), 6 component cases, and 2 e2e per theme.

The e2e deliberately does **not** just assert the prop was threaded through — a `data-depth` attribute with no visual consequence would satisfy the ticket on paper and change nothing an operator sees. It reads back `getBoundingClientRect().left` and requires strictly increasing offsets, and requires all-equal offsets for the orphan/cycle fixture.

**Mutation-checked.** With `depth += 1` changed to `depth += 0`, build exit 0 and `dist` present, both nesting e2e tests fail. (A first attempt at the mutation failed to compile, which left no `dist` at all — Playwright refused to start rather than serving a stale build, so that run proved nothing and was redone.)

### 2. Sonar: 3 → **0**

The 3 MINORs were `EmptyState` deprecation warnings. I said in my review that leaving them was defensible; the standing bar here is zero, and they are counted as new on this PR, so they are gone.

Both call sites moved to `StatusState` with `state={null}` and the `empty-state` test id preserved — a no-op for callers and tests. `state={null}` is the correct value for both: a trace that was read successfully and held no spans, and a list every row of which the operator's own filter is hiding, are **real empties, not absences**. Dressing either as an absence badge would report a healthy read as a fault, which is the same class of error this lane exists to remove. Blast radius was two call sites in one file — the deprecated wrapper already delegates to `StatusState` internally, so nothing cascaded.

**Quality Gate passed · 0 New issues · 0 Accepted · 0 Security Hotspots · 100.0% Coverage on New Code · 0.0% Duplication.**

### 3. Codecov: 96.34% → **"All modified and coverable lines are covered by tests"**

Three partials. Two were genuinely dead branches, so they were removed rather than tested around; one was a real uncovered state, so it got a real test.

| Partial | Resolution |
|---|---|
| `PayloadModal` `verdict === null` | **Dead — removed.** Memoising the verdict above the early return forced its type to `Certain<Verdict> \| null` so the hook could run on a closed modal, which forced a null check unreachable once `event` was non-null. `deriveVerdict` is two map lookups and does not need memoising; deriving it after the guard deletes the branch. |
| `TraceTimeline` `depths.get(id) ?? 0` | **Dead — removed.** `computeSpanDepths` now returns depths positionally aligned with `events`, so there is no lookup that can miss. Also correct when two spans share an id, where a by-id map would silently collapse them. |
| `TraceViewPage` duplicated `data ?? []` | **Dead — removed.** Coalesced once; repeating it per filter arm left one arm's fallback unreachable, since `data` is always defined by the time a severity is known. |

Two further partials were real gaps and got real tests: the unmapped-operation glyph fallback (a `SandboxCpuTimeout` span must render a neutral `·`, not guess a meaning — and the test asserts a *mapped* operation still gets its own glyph, so the fallback isn't swallowing everything), and the modal title when no verdict was recorded (the ordinary production case — `ToolCallIntercepted` with a null `decision` must show the absence marker, not reach for a chip).

Every remaining partial in these files sits on **pre-existing lines this PR does not touch** (the scrim Enter/Space handler, `agentId ?? params.id ?? ''`).

### 4. Correction to my own review — `ApprovalDetailDrawer`

I wrote that ADR-0017 item 11 marks it superseded and it "may simply be dead code to delete". **Having verified the mount claim properly, that recommendation was wrong** and I withdraw it.

It has no production mount — the only importers are its own `.test.tsx` and `.stories.tsx`, no route, no lazy import. But `tests/e2e/verify-aaasm-5075.spec.ts` renders it from the Storybook harness and states plainly: *"the drawer is not yet mounted on a product route (its full approval-review flow is backend-blocked, 5095)"*. It is a **deliberate pre-build awaiting AAASM-5095**, with parity evidence already captured under AAASM-5075 — not abandoned code. ADR-0017 item 11 supersedes the *mock's per-variant* drawer, which is a statement about the design mock, not a decision to delete this component.

So the follow-up ticket should **fix the bug in place, not delete the file**: `STATUS_VERDICT[status] ?? 'pending'` (`ApprovalDetailDrawer.tsx:34-41`) is an object literal keyed by wire `approval.status`, so `'constructor'` resolves `Object` off the prototype, `??` never fires, and a function is returned as a `Verdict` — `VERDICT_META[Object]` is then undefined and `.label` throws. Same shape as the bug fixed inside this PR's diff. Convert to a `Map` and replace the silent `'pending'` default.

### 5. Standing by my other two MINORs

**M2** (`A2AImpersonationAttempted` unmapped despite its doc saying "the action is denied") and **M3** (the 4591 port collision, whose `reuseExistingServer: true` *reuses* a colliding server rather than rejecting it, leaving the reverse direction able to go false-green) are unchanged. Both are fail-safe and neither blocks. I re-ran everything on **4604** with `reuseExistingServer: false` via a temporary config, removed afterwards.

---

### Verification

| Gate | Result |
|---|---|
| CI | **16 passed / 0 failed** |
| SonarCloud | Quality Gate passed · **0** new issues · 0 hotspots · **100.0%** new-code coverage |
| Codecov | **All modified and coverable lines covered** |
| `pnpm test` | **2699 passed** (was 2679 — +20) |
| `pnpm type-check` / `pnpm lint` | clean |
| `pnpm generate:api` | no drift |
| Playwright (port 4604) | **18 passed / 18** |

Screenshots in `dashboard/verify/5109/`, including new `nesting-light.png` / `nesting-dark.png`.

— Claude Code

PR #1722, opened 2026-07-26
[AAASM-5149] 🐛 (dashboard): Stop the shell badge reporting a failed alerts query as zero critical

Chisanan232:
## Description

Three tickets on the app-shell surface. One is a deceptive-UI defect; two are cosmetic. **One of the three is not implemented here** — see *Not in this PR* below.

**AAASM-5149 — the rail badge turned an alerts outage into "0 critical".**
`AppShell.tsx:133` counted `(alerts.data ?? []).filter(a => a.severity === 'CRITICAL')`. Two defects in one expression:

- the `?? []` folds a *failed* request to `0`, and `0` renders no badge — so an alerts outage was pixel-for-pixel identical to a healthy fleet;
- there was no status predicate, so a CRITICAL resolved weeks ago kept a red badge on the nav item permanently.

Both go away by consuming the selector `features/alerts/alertBadge.ts` already ships (delivered by PR #1712, which deliberately left this one line to the lane that owns the shell): `criticalFiringBadge(certainFromQuery(alertsQuery))`. The query *outcome* now travels to the DOM, so an absence renders as the shared `—` marker — tagged `unavailable`, tooltip and screen-reader sentence carrying the reason — and never as a number. A **known** zero still renders no badge: an unadorned rail item is the honest rendering of "nothing is firing", and that was already correct.

No `role="alert"` on the marker even for `unavailable`: the rail is persistent chrome that mounts with the session, so the role would announce on every cold boot. The absence is announced by the marker's own sentence instead.

**AAASM-5134 — the rail was still the pre-tokenisation slate/blue.** `#1e293b` with a 3px blue left-border accent, sitting beside fully hi-fi-tokenised page bodies. Repointed at `design/v{1,2}/hi-fi`: near-black ground, warm-grey items, solid white active pill, no border accent.

Two notes on the fidelity:

- The ticket says "bg `var(--ink)`". Taken literally that breaks the design: `--ink` inverts to a near-white in the dark theme, and the rail is the one surface kept dark in **both** themes for continuity (asserted today by `tests/e2e/theme-visual.spec.ts`). The literal hex `#0e0e0e` is used, which is what `--ink` evaluates to in light mode and what `--rail-bg` is in the mock.
- `:hover` outranks the single-class active rule, so without an explicit `--active:hover` the new white pill inverted to the dark hover fill under the cursor. The mock's two rules are equal-specificity with active declared last, i.e. the pill wins there too.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Behaviour changes visibly but no API does:

- The Alerts rail slot now shows `—` while the alerts query is in flight, has failed, **or returned a payload the client cannot parse**, where it previously showed nothing.
- `certainFromQuery` tests `isError` before `data`, and the alerts query retains its last successful rows on a failed refetch (`refetchOnWindowFocus` is on by default). So a *transient* refetch failure after a good load now replaces a real count with `—` rather than leaving the stale number on screen. That is the intended reading of the truthfulness rules — a stale count presented as current is its own false claim — but it is a wider change than "in flight or failed on first load" and is called out rather than buried. Narrowing it would mean changing `lib/truthfulness/query.ts`, which is shared by 15+ call sites across eight lanes, so it is deliberately not done here.
- The nav group headings are a slightly lighter grey than the mock specifies — see *Accessibility correction* below.

## Related Issues

- Related Jira ticket: [AAASM-5149](https://lightning-dust-mite.atlassian.net/browse/AAASM-5149) (primary), [AAASM-5134](https://lightning-dust-mite.atlassian.net/browse/AAASM-5134), [AAASM-5152](https://lightning-dust-mite.atlassian.net/browse/AAASM-5152) (not implemented — see below)
- Related GitHub issues: N/A

## Not in this PR — AAASM-5152

Confirmed still valid on `main`, and confirmed **entirely outside this lane's file ownership**. The second empty/error-state family is consumed by `pages/PoliciesPage.tsx:8`, `pages/TraceViewPage.tsx:10`, `pages/LiveOpsPage.tsx:6` and `features/analytics/PanelErrorBoundary.tsx:2`; the fix is to repoint the first three at `components/EmptyState`/`ErrorState`. Every one of those files is owned by another concurrent lane, and `features/alerts/` has its own bespoke empty states (`EmptyStateNoAlerts`, `EmptyStateNoRules`) rather than the generic family — so there is no part of 5152 inside the app-shell surface. Reported rather than edited.

## Wire-contract correction — the badge could never have matched (review follow-up)

An adversarial review of this PR found that the fix as originally written was still unable to light the badge from real data, and the ticket's headline symptom was not reproducible.

`readAlertsPage` was doing `body.items as readonly Alert[]` — a cast, not a check — and the wire does not speak the dashboard's vocabulary. Per `AlertResponse` (`aa-api/src/routes/alerts.rs:376-395`) and `StoredAlert` (`aa-api/src/alerts/mod.rs:29-115`):

| | wire | dashboard |
|---|---|---|
| `status` | `unresolved` · `resolved` · `suppressed` | `FIRING` · `RESOLVED` · `SUPPRESSED` |
| `severity` | `info` · `warning` · `critical` | `CRITICAL` · `HIGH` · `MEDIUM` · `LOW` |

So `severity === 'CRITICAL' && status === 'FIRING'` matched nothing a live server sends. Every derived count was zero, and a *known* zero renders no badge — the rail asserting a healthy fleet while criticals fired. Same deception AAASM-5149 exists to remove, reached through schema drift instead of `?? []`. It also means **the ticket's stated symptom — "a resolved CRITICAL kept a red badge permanently" — cannot have occurred in production**: the badge was never lit by real data at all. The defect was real; the reported symptom was not. Noted on the ticket.

What changed:

- **`features/alerts/parseAlert.ts`** (new) canonicalises both vocabularies and raises `AlertShapeError` on anything else. Option (a) — normalise at the boundary — was taken over (b) changing the `Alert` type, because `Alert` is consumed by `pages/AlertsPage.tsx` and `pages/OverviewPage.kpis.ts`, i.e. outside this lane's ownership.
- **An unreadable payload is now `unavailable`, not empty.** `Array.isArray(items) ? … : []` was the same fail-open one level down. A single unreadable row rejects the page rather than being dropped, because silently shrinking a count is the confident-zero failure in miniature.
- **The WebSocket boundary got the same treatment.** `useAlertsStream` cast frames just as blindly; left alone, one pushed frame would write a raw wire row into the shared alerts cache and undo the parse live — the nav badge reads that cache.
- **Severity mapping is honest about its one judgement call.** `critical → CRITICAL` is exact and is the only mapping this ticket depends on. `warning → HIGH` / `info → LOW` is an ordinal alignment between a three-level backend ladder and a four-level dashboard one (`MEDIUM` is unreachable from the current wire); it is documented in one table and **wants product ratification**.
- **Fixtures now speak the backend's dialect.** This was the root cause of the suite being green while the badge was dead: fixtures written in the dashboard's own vocabulary cannot detect a mismatch with the wire. A regression guard drives `critical`/`unresolved` through the parse into `criticalFiringBadge` and asserts a count of 1 — verified load-bearing by inverting a mapping entry.

`openapi/v1.yaml` and `api/generated/**` are untouched; the wire contract itself is a separate ticket.

## Accessibility correction — WCAG AA over the mock (ADR-0027)

The rail's group headings render at 11px in `#6a6a60` on the `#0e0e0e` rail: **3.53:1**, below the WCAG 2.1 AA floor of 4.5:1 for normal text. That value comes from the authoritative mock, so "match the mock" and "meet AA" gave opposite answers and nothing recorded said which wins.

Product call: the accessibility floor wins, and the spec is corrected at source.

- `--shell-nav-section-text` → **`#7b7b71`**, measuring **4.52:1** against `#0e0e0e` and **4.66:1** against v2's darker `#0a0907` dark-mode rail, so it clears in both themes. It is the smallest step on the same warm-grey ramp that clears 4.5:1 — `#7a7a70` reaches only 4.45:1 — so the design intent survives as far as it can.
- **`design/v2/hi-fi/styles.css` is corrected too**, with an inline `ACCESSIBILITY CORRECTION` note, so the next surface built from the mock cannot reintroduce it. `design/v1` is historical and untouched.
- **[ADR-0027](docs/src/adr/0027-accessibility-floor-overrides-visual-spec.md)** records the general precedence rule. It qualifies ADR-0025 (which makes v2 authoritative) rather than superseding it, and conflicts with nothing: ADR-0017 has no shell/rail section and no ADR previously mentioned WCAG at all.
- **Pinned by a test that runs.** `AppShell.contrast.test.ts` recomputes the ratio from the shipped token, in the **unit** suite — the natural home, `tests/e2e/theme-visual.spec.ts`, is in a Playwright suite no CI job executes.

Full rail audit: `--shell-nav-text` 15.44:1, `--shell-nav-text-muted` 11.15:1, `--shell-nav-text-dim` 5.44:1 all pass. **`--shell-nav-num` (route numbers) is the one remaining 3.53:1** and is knowingly left — it also renders on the `#1f1f1f` hover fill, where even the corrected grey reaches only 3.86:1, so clearing AA needs a hover-state decision rather than a value bump. Recorded as the single accepted risk in ADR-0027.

### Two disclosed side-effects of adding that test

1. `vite.config.ts` gains `test.css: true`. Vitest stubs CSS imports and that silently extends to `?raw`, which resolves to `''` rather than failing — so a test asserting over stylesheet text passes vacuously. The contrast guard has to see the real file.
2. Enabling it revived `OverviewPage.test.tsx`'s colour-token assertion, dead since it was written, which then tripped on its own header comment (prose containing "white"/"black"). Comments are stripped before matching. **This is a one-line touch in another lane's test file**, disclosed here so it can be backed out trivially if that lane objects.

## Follow-up left for the owner of `src/styles.css`

The `--shell-nav-*` tokens are declared at `src/styles.css:47-54` and consumed only by `AppShell.css`. Rather than edit a shared file that five concurrent lanes are also touching, the hi-fi values are declared on `.appshell` — beside the shell accent tokens already scoped there for the same stated reason — which wins the cascade for every consumer. The eight `:root` declarations are now superseded, and `--shell-nav-active-border` has no remaining reference; both want deleting by whoever owns that file next.

The Policy badge keeps its own `?? []` fail-open (a failed policies query still reads as "no inactive policies"). That is the same class of defect as 5149 but is not what 5149 describes, so it is left alone and flagged rather than silently widened into this PR.

The alerts list is still fetched one page at a time (`per_page` defaults to 50) and the badge `known()`s a count over that window, while `useAlertsPageQuery` — which exposes `total` — exists for exactly this kind of caller. Pre-existing; AAASM-5123 fixed it on the Alerts page but not on the rail. Reported, not widened into this PR.

## Testing

**Unit** — four cases added to `AppShell.chrome.test.tsx`. Three are red against the pre-fix shell, verified by reverting `AppShell.tsx` and re-running: a failed query rendered no badge, an in-flight query rendered no badge, a resolved CRITICAL rendered one. The fourth is a guard that holds either way, pinning the Policy badge to a plain count so the shared `NavBadge` cannot start decorating a route whose number carries no provenance. The alerts mock now mirrors the query-result shape, so a test can no longer describe "the request failed" by merely omitting `data`.

**E2E** — `tests/e2e/review-aaasm-5149.spec.ts`, both themes, console errors asserted `[]`. The outage is injected with `page.route` (a `fetch` shim installed in-page is never consulted — `openapi-fetch` captures `globalThis.fetch` at module load) and auth is seeded via `addInitScript`. Six of the eight cases are red against `main`'s `AppShell`, verified by checking it out, rebuilding and re-running; the two that pass either way assert that a loaded, quiet fleet stays silent. Screenshots in `dashboard/verify/5149/`.

Gates, each run once after the final change: `pnpm test` **2638 passed / 255 files** · `pnpm type-check` clean · `pnpm lint` clean · `pnpm generate:api` then `git status --porcelain` empty (no schema drift — `openapi/` untouched).

The e2e now runs 10 cases (was 8) and ships its config: `tests/e2e/review-5149.config.ts`, port 4605, following `review-5124.config.ts`. Re-run against a clean `rm -rf dist` rebuild with a checked exit code: 10/10.

## Checklist

- [x] Code follows project style guidelines (`cargo fmt`, `cargo clippy`) — no Rust touched; dashboard lint/type-check clean
- [x] All CI checks passing — 16/16 non-skipped green; SonarCloud quality gate passed with **0 new issues**, 0 security hotspots, 100% coverage on new code; Codecov: all modified lines covered

[AAASM-5149]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Adversarial review — PR #1722 (AAASM-5149 + AAASM-5134)

Reviewed independently against `remote/main` (branch is 4 ahead, 0 behind). Every claim in the PR body was re-derived rather than relayed. **Verdict: not ready as-is — no BLOCKER, but M1 needs an answer before merge.** The change itself is sound and a real improvement.

### Gates

| Gate | Result |
|---|---|
| CI | 16 passed / 0 failed |
| SonarCloud | Quality Gate passed — 0 new issues, 0 accepted, 0 security hotspots, 100.0% coverage on new code, 0.0% duplication |
| `AppShell.chrome.test.tsx` | 18/18 pass (re-run locally) |
| `tests/e2e/review-aaasm-5149.spec.ts` | 8/8 pass (re-run locally, port 4605) |

---

## The three judgement calls

### 1. No `role="alert"` on the absent badge — *premise sound, conclusion overreached* (MINOR)

The premise is correct: the rail is persistent chrome, and `role="alert"` is assertive, so it would interrupt on every cold boot. Rejecting `alert` is right.

But that argument only rules out the *assertive* role. It does not rule out `role="status"` — which is exactly what this programme's own `StatusState.roleFor()` (`components/truthfulness/StatusState.tsx:42-49`) assigns to every non-`unavailable` state, and which per the ARIA spec does **not** announce content already present when the region is registered. It would have been silent on cold boot and audible on a mid-session transition into `unavailable`. That option isn't addressed anywhere in the PR or the code comment.

Is the absence announced by some other means? **Yes, but only on navigation.** I checked: `.truth-sr-only` (`truthfulness.css:18-27`) is clip-hidden, not `display:none`, so it contributes to the accessible name — and the badge sits inside the `NavLink`, so a screen reader reaching the item hears *"Alerts Unavailable — the request for this value failed. HTTP 503, link"*. Verified in the DOM.

Net: **not an accessibility regression** — the state is reachable and correctly described. But an alerts outage that begins mid-session is silent for a screen-reader user who does not traverse the rail, and the stated reason doesn't justify going all the way to *no* live region.

### 2. CSS layering — **claim verified in full**

I enumerated every `--shell-nav-*` reference in the repo:

- **Declared** in two places only: `src/styles.css:47-54` (`:root`, the superseded slate/blue set) and `src/components/AppShell.css:9-31` (`.appshell`, the hi-fi set + the three pre-existing accents).
- **Consumed** at 21 sites, **all in `AppShell.css`, all under a `.appshell__*` selector** — `__nav`, `__nav-brand`, `__nav-brand-sub`, `__nav-section`, `__nav-link`(+`:hover`, `--active`, `--active:hover`), `__nav-num`, `__nav-star`, `__nav-badge--absent`, `__nav-foot`, `__nav-foot-dot`. No consumer anywhere else in `src/`, `tests/`, or `design/`.
- `AppShell.tsx:202` renders `<div className="appshell">` as the ancestor of every one of those nodes, and there is **no `createPortal` in the file** — so custom-property inheritance reaches all of them. `.appshell` beats `:root` for every descendant irrespective of specificity (they match different elements; the nearest ancestor's declaration wins).

So: **no consumer sits outside `.appshell`**, and none silently keeps the old palette.

Also confirmed: `--shell-nav-active-border` has **zero** `var()` references once the diff drops the only one, and `:root[data-theme="dark"]` (`styles.css:219+`) does **not** redeclare any `--shell-nav-*`. Dead tokens correctly left for the owning lane rather than deleted.

**The 5134 refutation is correct, and on evidence independent of the test it cites.** `--ink` is `#0e0e0e` in `:root` and inverts to `#f0ede4` under `:root[data-theme="dark"]`; the comment at `styles.css:219-224` states the rail is intentionally *not* overridden in dark. Taking the ticket literally would have flipped the rail to near-white in dark mode. Using `#0e0e0e` is right. (But see F2 — the test cited as the guardrail is currently dead.)

### 3. AAASM-5152 not implemented — **claim verified**

Importers of `components/states` are exactly, and only:

- `pages/PoliciesPage.tsx:8`
- `pages/LiveOpsPage.tsx:6`
- `pages/TraceViewPage.tsx:10`
- `features/analytics/PanelErrorBoundary.tsx:2`

Nothing in `components/AppShell.*` or `features/alerts/**`. `features/alerts/` does use its own `EmptyStateNoAlerts` / `EmptyStateNoRules` (`AlertsFeedBody.tsx:66,69`, `AlertRulesTable.tsx:129`), not the generic family. The dead `policy` copy entry at `components/EmptyState.tsx:52-63` is confirmed dead — `'policy'` appears only in the union type at `:7` and in the table itself. **No part of 5152 was reachable from this lane; correctly skipped, correctly reported.**

One thing for whoever picks 5152 up: `components/truthfulness/StatusState.tsx:1-16` says the generic pair *now delegates* to `StatusState`, so a later lane has already begun converging the two families. 5152's target behaviour may need restating before it's actioned.

---

## Findings

### MAJOR

**M1 — `known()` certifies a count that the live wire contract cannot produce, and the new `status` predicate widens the silent-zero surface.**

`readAlertsPage` (`features/alerts/api.ts:99-106`) does `body.items as readonly Alert[]` — a blind cast, no validation, no normalisation. What `GET /api/v1/alerts` actually returns (`openapi/v1.yaml` → `AlertResponse`, generated from `aa-api/src/routes/alerts.rs:376-395`):

- `status` is `"unresolved"` on capture / `"resolved"` after resolve — **there is no `"FIRING"`**;
- `severity` is lowercase (`"critical"`, `"warning"`) — **there is no `"CRITICAL"`**;
- the schema carries no `rule_id`, `rule_name`, `first_fired_at`, `resolved_at` or `destination_ids` at all.

Against the real backend `criticalFiringCount` therefore returns `0` for **every** payload — and a *known* zero renders no badge, so the rail reads "nothing critical is happening" while criticals fire. That is the same deception class 5149 exists to remove, reached through schema drift instead of `?? []`.

Being precise about blame:

- **Not a regression.** The pre-PR `severity === 'CRITICAL'` never matched either, so production behaviour is unchanged.
- **Not this PR's invention.** The divergence is feature-wide — `OverviewPage.kpis.ts:148`, `AlertsPage.tsx:45`, `urlFilters.ts:10`, `AlertFilterBar.tsx:9`, `AlertStatsStrip.tsx:43` all use the same uppercase vocabulary, and `endpoints.ts:1-6` documents the feature as deliberately built ahead of unshipped Stories AAASM-1385/1386/1387/1388/1389. But `list` **is** shipped and **is** consumed live today, so this one path is real present-day drift.
- **What this PR does add:** wrapping the result in `known()` is, in this programme's vocabulary, an explicit certification of provenance. Certifying a number derived from an unvalidated cast of an unverified payload is a claim the code cannot support — `known()` is exactly the assertion that should not be made here without a guard.

Two knock-on consequences worth stating plainly:

1. **The ticket's headline premise is not reproducible in production.** "A CRITICAL resolved weeks ago kept a red badge on the nav item permanently" cannot have happened — the badge was never lit by real data, because `"critical" !== 'CRITICAL'`. The defect is real in the *code*; the *symptom* described on AAASM-5149 is not.
2. **No test in this PR can detect any of this.** Every unit and e2e fixture is hand-written in the dashboard's aspirational vocabulary, so the suite is self-consistent and blind to the contract it actually talks to.

Asked for: a follow-up ticket covering `GET /api/v1/alerts` shape normalisation (or a validating parse at `readAlertsPage`), and — at minimum in this PR — an inline note at the `known()` recording that its provenance claim is only as good as the unvalidated cast beneath it.

### MINOR

**m1 — no live region on the absent badge.** See judgement call 1. `role="status"` on the badge container is the option the stated reasoning does not rule out, and matches `StatusState.roleFor()`.

**m2 — contrast regression on the rail, PR-introduced.** `--shell-nav-section-text` and `--shell-nav-num` are both `#6a6a60` on `#0e0e0e` → **3.53:1**, below WCAG AA's 4.5:1 for text under 18.66px (both are `0.6875rem` ≈ 11px). Previously `#cbd5e1` at `opacity: .7/.65` over `#1e293b` ≈ 5–6:1. The route numbers are arguably decorative; the **group headings (`.appshell__nav-section`) are navigational structure** and shouldn't be. The values come straight from the mock, which is authoritative per ADR-0025 — so flagging as **PRODUCT-DECISION-REQUIRED** rather than asking for a unilateral deviation.

**m3 — undisclosed behaviour change on background-refetch failure.** `certainFromQuery` tests `isError` first, and `useAlertsQuery` retains the last successful `data` on a failed refetch (`refetchOnWindowFocus` is on by default). So a transient blip after a good load now replaces a real count with `—`, not just "in flight or failed on first load" as the Breaking Changes note says. Defensible under the truthfulness rules — showing stale data as current is its own lie — but it should be in the note.

**m4 — the shipped e2e spec has no committed config.** `playwright.config.ts` has `testDir: 'tests/e2e'`, so `review-aaasm-5149.spec.ts` *is* picked up by the default suite — but pinned to port 4173 and carrying `OUTAGE_SETTLE_MS = 30_000`, so it will contend with other lanes' preview servers and slow the default run. Sibling lanes commit a per-ticket config (e.g. `tests/e2e/review-5124.config.ts`); this one didn't. To run it I added `dashboard/tests/e2e/review-5149-adv.config.ts` (port 4605, `testMatch` pinned to this spec, `reuseExistingServer`) following that pattern — left **untracked**, not committed; adopt or discard as you prefer.

### OUT-OF-SCOPE-FOLLOW-UP

**F1 — Policy badge `?? []` (AAASM-5186): genuinely out of scope, confirmed.** 5149's text is specifically the alerts nav badge. The PR touches the policy branch only to satisfy the new return type (`known(inactivePolicies)`); behaviour is byte-identical, the inline comment is honest about the remaining fail-open, and a unit test pins the policy badge to a plain count so `NavBadge` can't start decorating it. Correctly deferred, not a convenient omission.

**F2 — `tests/e2e/theme-visual.spec.ts` is currently dead, and it is the evidence the 5134 refutation cites.** `seed()` at `:132-140` writes `aa_token` into **localStorage**; `src/auth/tokenStorage.ts` reads **sessionStorage only** (moved there by AAASM-4322). The app therefore never authenticates. I ran the *"nav rail + code/terminal palette stay dark in both themes"* test on this branch: it times out at `getByTestId('appshell')` — the shell never mounts, so the rail assertion is never reached. Dashboard Playwright is not in a CI job either, so nothing caught it.

So "the rail is dark in both themes, asserted today by `theme-visual.spec.ts`" is a true statement about the test's *intent* and a false one about it being *enforced*. The refutation itself still stands on `styles.css:219-224` and the actual dark-block token values — but the guardrail is not there. Pre-existing, not caused by this PR; needs its own ticket.

**F3 — the badge counts one page and `known()`s the total.** `fetchAlertsPage` requests `/api/v1/alerts` with no pagination params and the API defaults to `per_page: 50`. `useAlertsPageQuery` exists precisely "for callers that must be able to say how much of the fleet the page covers" — a count badge is such a caller, but the rail uses `useAlertsQuery`. Pre-existing; AAASM-5123 fixed this on the Alerts page but not on the rail.

### NIT

**n1 —** `.appshell__nav-link` gains `margin: 1px 0.5rem`; the mobile `.appshell__nav--open` rail isn't covered by the new spec.

---

## Side effects (AppShell renders on every page)

**(a) Known zero still renders no badge — confirmed.** `badgeFor` returns `null` for `isKnown(criticalAlerts) && value === 0`, so the pre-PR `criticalAlerts || null` behaviour is preserved exactly. Pinned by a unit test *and* by two e2e cases (`alerts-quiet-rail-{light,dark}`) that assert `nav-badge-alerts` has count 0 while the rail is demonstrably mounted. A real zero does **not** now render a marker.

**(b) Palette vs `theme-visual.spec.ts` — no regression, but unverifiable by that suite.** On paper the new values satisfy it: `#0e0e0e` has relative luminance ≈ 0.055, well under the test's 0.35 threshold, and since the dark block declares no `--shell-nav-*`, light and dark resolve identically as the test requires. I could not confirm by execution — see F2, the test cannot mount the app on this branch or any other.

**(c) Query wiring — clean.** `features/alerts/api.ts` is not in the diff; the `useAlertsQuery(DEFAULT_ALERT_FILTERS)` call site is unchanged. The change is render-only: fetch cadence, retry policy, `queryKey`, `placeholderData` and cache behaviour are untouched for every other consumer. Subject to m3, which is a rendering consequence, not a query one.

---

## Frontend validation

Ran the repo-local `@playwright/test` (not the MCP server) on port 4605 via the added config: **8 passed, 0 failed, 17.9s**. Screenshots regenerate to `dashboard/verify/5149/` — `alerts-outage-{light,dark}`, `alerts-outage-rail-{light,dark}`, `alerts-firing-only-{light,dark}`, `alerts-firing-only-rail-{light,dark}`, `alerts-quiet-rail-{light,dark}`, `rail-palette-{light,dark}`. I restored the committed PNGs afterwards; the worktree is clean apart from the untracked config.

**Does the spec prove the claim, or assert around it? It proves it.** It fails `GET /api/v1/alerts` with a real 503 at the network layer via `page.route` (correctly, not via a `fetch` shim), waits out the real `QueryClient` retry chain rather than reconfiguring the app to fail faster, and — the part that matters — clones the chip and strips `.truth-sr-only` before asserting the sighted text is exactly `—`, so a digit cannot hide behind the screen-reader sentence. It then asserts the SR sentence separately, asserts a known zero stays silent, and asserts `border-left-width: 0px` on the active pill.

The "6 of 8 red against `main`" claim is consistent by inspection of `main`'s expression: 2 outage cases (main renders no badge at all) + 2 mixed-count cases (main counts 3, not 1) + 2 palette cases (main is `#1e293b`) = 6 red; the 2 quiet cases pass either way, exactly as stated.

Its one blind spot is M1: the fixtures are written in the dashboard's own aspirational vocabulary, so no amount of green here says anything about the shape the backend actually sends.

---

## Summary

| Severity | Count |
|---|---|
| BLOCKER | 0 |
| MAJOR | 1 (M1) |
| MINOR | 4 (m1–m4; m2 is also PRODUCT-DECISION-REQUIRED) |
| OUT-OF-SCOPE-FOLLOW-UP | 3 (F1–F3) |
| NIT | 1 |

Scope: 5149 is covered and deliberately exceeded (the ticket asked only for the status predicate; removing the `?? []` fail-open is a justified widening, consistent with the truthfulness programme). 5134 is covered with a correct, well-evidenced refutation of the ticket's literal `var(--ink)`. 5152 is correctly outside this lane. All three judgement calls hold up; calls 2 and 3 verified in full, call 1 partially.

**Not ready to merge as-is.** Nothing here is unsafe and the diff is a genuine improvement, but M1 should be answered first — either a validating parse at `readAlertsPage`, or a follow-up ticket plus an inline note that the `known()` is only as trustworthy as the unvalidated cast under it. F2 also needs a ticket: the test the 5134 refutation leans on does not currently run.

Not approving and not merging — review only.

— Claude Code

Chisanan232:
## Review follow-up — what changed since the review above

Nine commits on top of the reviewed tip (`b00f258d` → `58ecbc2f`). The MAJOR finding (M1) is fixed, the contrast finding is fixed under a product ruling that reversed my earlier "leave it", and one MINOR is disclosed rather than changed. Still **not approving and not merging** — review only.

### M1 fixed — the badge could never have matched real data

The finding stands and was worse than "a provenance claim the cast can't support": with `severity === 'CRITICAL' && status === 'FIRING'` compared against a wire that sends `critical` / `unresolved`, the badge counted **zero for every live payload**, and a known zero renders no badge. The rail was still asserting a healthy fleet while criticals fired — the exact deception 5149 exists to remove, reached through schema drift instead of `?? []`.

- **`features/alerts/parseAlert.ts`** (new) validates and canonicalises at the boundary, raising `AlertShapeError` on anything it cannot read, so an unreadable payload lands on `unavailable` — the same place a failed request lands — instead of on an empty fleet.
- **Option (a), not (b)**, and the reason is ownership, not taste: changing the `Alert` type would reach `pages/AlertsPage.tsx` and `pages/OverviewPage.kpis.ts`, outside this lane. Both vocabularies are accepted so the rule-engine Stories (AAASM-1385…1389) can land without a flag day.
- **The `?? []` sibling one level down is gone too** — `Array.isArray(items) ? … : []` turned a malformed envelope into an empty fleet. A single unreadable row now rejects the page rather than being dropped, because silently shrinking a count is the same failure in miniature.
- **The WebSocket boundary was a second blind cast.** Left alone it would have written raw wire rows into the shared alerts cache and undone the parse live — the badge reads that cache. Without this the fix comes apart the moment a frame arrives.
- **Fixtures now speak the backend's dialect**, which was the root cause of a green suite over a dead badge. A regression guard drives `critical`/`unresolved` through the parse into `criticalFiringBadge` and asserts 1; verified load-bearing by inverting a `WIRE_STATUS` entry (3 tests red, including the count).

Two things the maintainer should weigh rather than take from me:

1. **`warning → HIGH` / `info → LOW` is an ordinal alignment, not a ratified mapping.** Only `critical → CRITICAL` is exact and load-bearing here. `MEDIUM` is unreachable from the current wire. One table, documented, **wants product ratification**.
2. **The ticket's headline symptom is not reproducible.** "A resolved CRITICAL kept a red badge permanently" cannot have happened in production — the badge was never lit by real data. The defect was real; the reported symptom was not. Worth correcting on AAASM-5149 so the next reader doesn't hunt for it.

### Contrast — reversed by product ruling, now fixed (ADR-0027)

My review classified `#6a6a60` at 3.53:1 as PRODUCT-DECISION-REQUIRED and deferred to the mock. That call came back the other way: **WCAG AA takes precedence over the mock colour value.**

- `--shell-nav-section-text` → **`#7b7b71`**: **4.52:1** against the `#0e0e0e` rail, **4.66:1** against v2's darker `#0a0907` dark-mode rail. Clears in both themes, and is the smallest step on the same warm-grey ramp that clears 4.5:1 (`#7a7a70` reaches 4.45:1), so the design intent survives as far as it can.
- **`design/v2/hi-fi/styles.css` corrected at source**, not just downstream — otherwise the spec keeps specifying a value we refuse to ship. `design/v1` left alone as historical.
- **[ADR-0027](docs/src/adr/0027-accessibility-floor-overrides-visual-spec.md)** records the general rule. It *qualifies* ADR-0025 rather than superseding it, and conflicts with nothing — ADR-0017 has no shell/rail section, and no ADR previously mentioned WCAG, contrast, or accessibility at all.
- Full rail audit: `--shell-nav-text` 15.44:1, `--shell-nav-text-muted` 11.15:1, `--shell-nav-text-dim` 5.44:1 — all pass. **`--shell-nav-num` is the one remaining 3.53:1**, knowingly left: it also renders on the `#1f1f1f` hover fill where even the corrected grey reaches only 3.86:1, so clearing AA needs a hover-state decision, not a value bump. Single accepted risk in ADR-0027.

### A dead assertion found while pinning the ratio

The contrast guard had to read the shipped stylesheet, which exposed that **vitest stubs CSS imports and that extends to `?raw` — it resolves to `''` instead of failing.** Any test asserting over stylesheet text was passing vacuously. `vite.config.ts` now sets `test.css: true`.

That revived `OverviewPage.test.tsx`'s colour-token assertion, dead since it was written, which then tripped on its own header comment (prose containing "white"/"black"). Comments are stripped before matching. **That is a one-line touch in another lane's test file** — disclosed so it can be backed out trivially if that lane objects. Measured blast radius of `css: true` was exactly that one test across the full 2638-test suite.

### m3 disclosed, not changed

`certainFromQuery` tests `isError` before `data`, so a transient refetch failure after a good load now replaces a real count with `—`. Correct under the truthfulness rules — a stale count shown as current is its own false claim — but wider than the PR originally said, so the Breaking Changes section now says it. Narrowing it would mean editing `lib/truthfulness/query.ts`, shared by 15+ call sites across eight lanes; that is a truthfulness-policy decision, not a side effect of this PR.

### Unchanged findings

- **AAASM-5186 (Policy badge `?? []`)** — re-confirmed genuinely out of scope.
- **One-page counting** — the badge still `known()`s a count over a 50-row window while `useAlertsPageQuery` exists for callers that must state coverage. Pre-existing (AAASM-5123 fixed the Alerts page, not the rail), now noted in the PR body.
- **`tests/e2e/theme-visual.spec.ts` is dead** — see below. Confirmed, deliberately **not** fixed here.

### F2 — confirmed, with the three answers

**(a) How long, and what broke it.** Commit `d68a0d63` "✨ (dashboard/auth): Add tokenStorage helper backed by sessionStorage" (AAASM-4322), **2026-07-08**. The spec was authored 2026-06-05 (`21748010`) and was correct when written; the auth migration invalidated it ~19 days ago. It is **not one spec** — **31 e2e files** seed `aa_token` into `localStorage` and were all broken by that one commit.

**(b) CI.** CI never runs it. **No workflow in `.github/workflows/` references Playwright at all** — the dashboard e2e suite is not executed anywhere in CI, so nothing caught this and nothing will. The breakage is invisible by construction, which is also why the new contrast guard was put in the unit suite.

**(c) Is the one-line fix enough?** **Yes, for this spec.** I patched `seed()` to `sessionStorage` in a scratch copy and ran the "nav rail + code/terminal palette stay dark in both themes" test: it passes in 281ms. No further rot behind the first failure. Then reverted — the spec is committed exactly as found.

A useful side-effect of that experiment: it independently confirms side-effect (b) from my review. **The new `#0e0e0e` rail satisfies the theme-visual contract** (identical in both themes, luminance below the 0.35 threshold) — something I could not verify at review time because the spec cannot mount the app.

### Verification

Gates, each once after the final change: `pnpm test` **2643 passed / 255 files** · `pnpm type-check` clean · `pnpm lint` clean · `pnpm generate:api` + `git status --porcelain` empty (no schema drift; `openapi/` and `api/generated/**` untouched).

E2E: **10/10** (was 8) on port 4605, against a `rm -rf dist` rebuild with a **checked exit code** — a failed build leaves `vite preview` serving a stale bundle, which makes a mutation check falsely read as "tests aren't load-bearing". The spec now ships its config (`tests/e2e/review-5149.config.ts`), which it previously lacked. Screenshots refreshed in `dashboard/verify/5149/`, including the two new unreadable-payload captures.

Codecov flagged 3 uncovered lines on the first push (patch 94.44%); `parseAlert.ts` is now fully covered and the stream's drop-on-unreadable arm is exercised. The lines still uncovered in `useAlertsStream.ts` (88-89, 97-98, 113) are pre-existing reconnect/socket/JSON-parse paths, not part of this change. Sonar passed throughout — 0 new issues, 0 hotspots, 0% duplication.

— Claude Code

PR #1723, opened 2026-07-26
[AAASM-5141] 🐛 (dashboard): Report the real modified-rule count and open the shipped simulator

Chisanan232:
## Description

Two truthfulness defects on the policy-editor surface, both confirmed present on `main` before fixing.

**AAASM-5141 — the footer reported the policy's rule total as the modified-rule count.**
`PolicyEditorOverlay.tsx:99` rendered `${draft.rules.length} rule(s) modified`, so editing one field of a nine-rule policy claimed nine rules were modified. That line is the pre-save blast-radius number — what an operator reads to decide whether simulating is worth it — so it overstated the impact of every edit. It now counts real per-rule changes against the snapshot the overlay already builds for the per-rule dirty dot, via a new `rulesChangedCount` helper. The footer number and the visible dirty dots therefore cannot disagree. A metadata-only edit reports `0`, which is truthful: no rule changed.

`rulesChangedCount` is ported from `design/v2/hi-fi/policy-editor.jsx:559` but keyed by each rule's stable `id` rather than by list position. The mock diffs positionally and adds the length difference, which double-counts every rule that shifts up after a mid-list removal and miscounts a reorder; the shipped editor already carries stable ids (and already builds the id map), so identity is available and is what an operator means by "this rule changed".

**AAASM-5142 — the Simulate button denied a shipped feature existed.**
`PolicyEditorOverlay.tsx:88-94` toasted `Simulate impact: coming soon.` while `PoliciesPage.tsx:521` already mounted a fully working `PolicySimulatePanel` — the ratified single-request dry-run (ADR-0017 item 6) — behind the page-header button. The editor button now opens that same panel, keeping the existing `errors > 0` guard ahead of it. `onSimulate` is a **required** prop rather than optional, so the button can never render with no production path behind it. This is not the deferred replay `SimulateModal` (AAASM-5094).

Ownership of the open flag stays on `PoliciesPage`, which already drives the panel from its header button, rather than being lifted into `PolicyEditorOverlayContainer` as the ticket's wording suggested — two owners would mean two mounted dialogs and two sources of open/closed truth.

**One regression this wiring introduced, fixed in the same PR.** The simulator is a native modal `<dialog>` in the top layer, but its Escape keydown still bubbles to the document-level listener `OverlayHost` installs. With the editor now able to open it, a single Escape dismissed the simulator *and* the editor behind it, discarding the operator's unsaved draft. Verified in a real browser: reverting the guard turns the Playwright Escape test red in both themes. Escape now belongs to the topmost surface only.

No product semantic owned by a Proposed ADR (0019–0026) was decided here, and nothing recorded as Accepted in ADR-0017 was reverted — item 6 ratifies exactly the Simulate v0 this PR wires up.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

`PolicyEditorOverlay`'s new `onSimulate` prop is required, but the component has exactly one caller (`PolicyEditorOverlayContainer`), which is updated in the same commit.

## Related Issues

- Related Jira tickets: [AAASM-5141](https://lightning-dust-mite.atlassian.net/browse/AAASM-5141), [AAASM-5142](https://lightning-dust-mite.atlassian.net/browse/AAASM-5142)
- Related GitHub issues: N/A

## Testing

Every test was verified to genuinely fail without its fix by reverting the source change and watching it go red, then restoring:

| Reverted change | Red tests |
|---|---|
| footer back to `draft.rules.length` | 5 (all of `footer blast radius (AAASM-5141)`) |
| `onSimulate()` back to the "coming soon" toast | 5 (1 overlay + 4 page) |
| Escape guard removed | 1 unit + 2 e2e (light and dark, in a real browser) |

- `rulesChangedCount.test.ts` — 9 cases: edits, add, remove, mid-list removal (where the mock's positional diff over-reports), pure reorder, empty snapshot, unchanged.
- `PolicyEditorOverlay.test.tsx` — footer count across edit / second edit / add / remove / metadata-only, asserting the count matches the number of dirty dots on screen; Simulate calls `onSimulate` and never says "coming soon"; a validation error blocks it.
- `PoliciesPage.test.tsx` — the editor's footer button opens the real panel, the editor and draft survive underneath, an invalid draft opens nothing, and Escape dismisses the simulator alone then the editor.
- `tests/e2e/review-aaasm-5141.spec.ts` (port 4594, light **and** dark, console errors asserted `[]`) — drives both regressions through the real overlay against a four-rule policy in the editor's own `spec.rules` schema, including a real simulate round-trip returning a `deny` verdict with the unsaved draft intact behind it. Evidence: `dashboard/verify/5141/` (22 screenshots).

Gates, run once each on the final tree:

```
pnpm test        → Test Files 254 passed (254) · Tests 2635 passed (2635)
pnpm type-check  → clean (exit 0)
pnpm lint        → clean (exit 0)
pnpm generate:api && git status --porcelain → no schema diff
pnpm exec playwright test --config tests/e2e/review-5141.config.ts → 10 passed (5.2s)
```

No Rust crate is touched, so no Rust suite was run.

## Checklist

- [ ] All CI checks passing

[AAASM-5141]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Adversarial review — PR #1723 (AAASM-5141 / AAASM-5142)

**Verdict: ready to merge. No BLOCKER, no MAJOR.** Three MINORs and two NITs below, all
comment/test-documentation accuracy or latent-risk notes — none change behaviour and none
need to be fixed before merge.

I did not write this PR and I have not taken its reasoning on trust: both load-bearing
claims were re-derived from the code and re-verified in a real browser, including by
deleting the fix and watching the failure return.

### Gate state

| Gate | Result |
|---|---|
| CI | **16 passed / 0 failed** |
| SonarCloud | **Quality Gate passed** — 0 new issues, 0 accepted, 0 security hotspots, 100.0% coverage on new code, 0.0% duplication |
| `vitest` (editor dir + PoliciesPage + OverlayHost + PolicySimulatePanel) | **182 passed, 0 failed** |
| PR Playwright suite, **port 4606** | **10/10 passed (5.0s)**, light + dark |

Playwright port override: I ran a reviewer copy of `tests/e2e/review-5141.config.ts` with
`PORT` 4594 → 4606, absolute `testDir` / `webServer.cwd` (the copy lives outside the repo),
and `reuseExistingServer: false`. No repo file was modified. Screenshots regenerated into
`dashboard/verify/5141/` and then restored to the committed versions — worktree left
pristine (`git status --porcelain` empty).

---

### Judgement call 1 — deviating from AAASM-5142's "lift `setSimulateOpen`" instruction

**The deviation is correct on the merits.** Verified against the wiring rather than the
ticket text: `PoliciesPage.tsx:377` already owns `simulateOpen`, the page-header
▸ Simulate at `:476` already drives it, and `:542` mounts the single
`<PolicySimulatePanel>`. Lifting the flag into `PolicyEditorOverlayContainer` would have
forced either a **second** `PolicySimulatePanel` mounted inside the container — two native
`<dialog>`s, two `showModal()` calls into the top layer, two simulate clients — or
hoisting the panel as well, which is what the page already does. The ticket's wording was
written from the audit, not from the wiring; following it literally would have produced
the two-owners defect the author describes. Reasoning confirmed, not relayed.

One process point, not a merge blocker: the deviation is recorded in the PR body and in a
code comment at `PoliciesPage.tsx:37-41`, but **not on AAASM-5142 itself**. A one-line Jira
comment ("implemented with ownership on `PoliciesPage`; not lifted — single panel
instance") would close the audit-trail loop per `.claude/rules/06-audit-trail.md`.

### Judgement call 2 — the Escape guard

**Correct, complete for every reachable path, and it does *not* create the inverse bug.**

*Is the regression real?* I deleted `if (simulateOpen) return` from `PoliciesPage.tsx:431`,
rebuilt, and re-ran. Both Escape e2e tests went red at exactly
`expect(policy-editor-overlay).toBeVisible()` — one Escape does dismiss the simulator **and**
the editor behind it. Restored with `git checkout HEAD --`. The regression was not invented.

*Is the ordering safe?* The guard reads React state from a **document** `keydown` listener,
so it only holds if keydown reaches the document before the flag flips. It does: the browser
dispatches keydown first and processes the close request afterwards, and
`PolicySimulatePanel.tsx:128-135` `preventDefault()`s `cancel` and routes closure through
`onClose`. React therefore stays the sole owner of `simulateOpen` and the `<dialog>` never
closes out from under it — there is no window where the flag is stuck `true`.

*Could it now swallow an Escape that should close the editor?* **No.** The PR's own e2e only
proves the second Escape with a *clean* draft, so I wrote my own probe against a **dirty**
draft — the case where a wrongly-swallowed Escape would cost the operator most:

- Escape #1 → simulator closes alone, editor survives, footer still reads
  `1 rule(s) modified` (draft intact).
- Escape #2 → correctly raises **"Discard unsaved changes?"** — not swallowed.
- Reopen-after-Escape probe → no stuck state, panel reopens from both entry points.
- With the guard removed, that same probe fails.

*Scope of the guard* — see NIT 4: it is broader than "Escape".

---

### Scope vs tickets

Both fully covered, nothing extra. **5141**: helper added, footer wired, metadata-only edit
reports `0`. **5142**: button wired to the shipped panel with the `errors > 0` guard retained
ahead of it, and `onSimulate` made **required** so the button cannot render with no
production path — the right shape for the "no action without a successful path" rule.

I checked ADR-0017 item 6 verbatim: the single-request dry-run **is** the ratified v0 and the
replay `SimulateModal` **is** explicitly deferred to AAASM-5094. This wires the correct
component and reverts no Accepted evolution; no Proposed ADR (0019–0026) semantics are
decided. Note both tickets cite `design/v1/hi-fi/` — the author correctly worked against
`design/v2/hi-fi/policy-editor.jsx`, which is authoritative per ADR-0025.

### Side effects

**(a) `rulesChangedCount` — no reachable false 0.** Verified in a real browser, not just in
unit tests: edit → 1, add → 1, remove → 1, metadata-only → 0. Duplicate gets `nextRuleId()`,
so it has no snapshot and counts 1. Id collisions are impossible — `draftFromPolicy` assigns
`nextRuleId()` to **every** rule rather than reusing the YAML id, so `originalRuleById` can
never collapse two rules into one entry and under-count. Pure reorder is the only
change-that-reports-0, and there is no reorder affordance (see MINOR 3).

**(b) `onSimulate` required — nothing else breaks.** Exactly one production render site
(`PoliciesPage.tsx:90`), updated in the same commit, plus the test file. `types.ts:116`
`PolicyEditorOverlayProps` is the separate `useOverlay` payload union and is unaffected.
`tsc` clean.

**(c) Two stacked dialogs — sane.** The editor is a portal `div[role=dialog][aria-modal]` in
the overlay mount; the simulator is a real `<dialog>` + `showModal()` in `document.body`, so
it enters the top layer above the editor regardless of DOM order and gets the native focus
trap and `::backdrop` free, focusing its first field on open. I verified inertness
empirically: a forced mouse click at the editor Cancel button's coordinates while the
simulator is up never reaches Cancel — it lands on the dialog backdrop and dismisses the
simulator. Scroll lock unchanged. Cosmetic caveat: the editor keeps `aria-modal="true"` while
the simulator is above it, so two elements claim modality at once — pre-existing pattern, not
introduced here.

**Cross-PR with #1720:** the five new `PoliciesPage.test.tsx` tests use the file's shared
`Wrapper` and the same `new-policy-btn` click as ~12 tests already in that file. They add no
new dependency on the permissive `ALL_SCOPES` default, so #1720's `Wrapper`/`GrantScopes`
rewrite affects them exactly as it affects the existing tests. No contrary evidence found.

---

### Findings

**MINOR 1 — a fabricated number inside a correct test.** `rulesChangedCount.test.ts:55-62`
justifies its assertion with "the positional diff in the hi-fi mock reports 3 here (one
length difference plus two positional mismatches)". Recomputing the actual mock
(`design/v2/hi-fi/policy-editor.jsx:559-564`) for that exact input: `n = |2-3| = 1`, then
`min(2,3) = 2` positions compared of which only index 1 mismatches → the mock reports **2**,
and there is **one** positional mismatch, not two. The assertion (`toBe(1)`) and the
conclusion (the mock over-reports) are both right; the arithmetic attached to them is not.
Flagging it precisely because it is the house failure mode — a true observation with an
invented supporting number.

**MINOR 2 — an invariant stated more strongly than it holds.** `rulesChangedCount.ts:22`
claims the footer total and the visible dirty dots "can never disagree". They necessarily
disagree on **deletion**: remove a rule from a 4-rule policy and the footer reads
`1 rule(s) modified` while **zero** dirty dots are on screen, because the removed card is
gone. (`RuleCard` also returns early for `rule.unknown` before rendering the dot — a second
exception.) The count is right; the invariant as written is false. Suggest softening to
"every dot on screen is counted" — which is true.

**MINOR 3 — a latent false 0 that the test suite currently defends.** `rulesChangedCount`
returns 0 for a pure reorder, and `rulesChangedCount.test.ts:64` enshrines that as the
correct answer. Unreachable today — `useDraft` exposes only update/add/duplicate/remove/reset
and there are no drag handles — so this is not a live defect. But rule order **is** evaluation
order, and `serializeDraft.ts:66-67` derives the persisted rule id positionally
(`R${idx+1}-${resource}-${action}`), so a reorder would rewrite every rule's id in the saved
YAML while the footer claimed `0 rule(s) modified`. If a reorder affordance ever ships, that
test will actively defend the wrong answer. Worth renaming to record it as an accepted
limitation, or counting index changes. Follow-up ticket, not this PR.

**NIT 4 — the guard is broader than the PR body says.** It lives in `attemptCloseEditor`, so
while `simulateOpen` it suppresses *every* dismiss route `OverlayHost` owns — Escape, backdrop
click, backdrop Enter/Space — and, because `PoliciesPage.tsx:69` passes
`onClose={onRequestClose}`, the editor's Cancel button too. The PR body's "Escape now belongs
to the topmost surface only" is narrower than what the code does. Nothing is actually dead
(top-layer inertness, verified above), so this is correct by construction rather than by
design; it is a condition on one specific sibling rather than "am I the topmost surface", and
a third stacked surface would need another flag.

**NIT 5** — `changedRules` is recomputed unmemoized every render
(`PolicyEditorOverlay.tsx:104`), `JSON.stringify`-ing every rule per keystroke. Same order of
cost `RuleCard` already pays per card. Noted, not worth changing.

**OUT-OF-SCOPE-FOLLOW-UP** — the editor's `aria-modal="true"` staying set beneath the
simulator (side effect (c)) and the reorder limitation (MINOR 3).

I have not approved and have not merged, per my brief.

— Claude Code

PR #1724, opened 2026-07-27
[AAASM-5190] 🐛 (dashboard): Key the approval verdict lookup by Map and drop the silent pending default

Chisanan232:
## Description

`ApprovalDetailDrawer` derived its head verdict chip with a plain-object lookup keyed by the wire value `approval.status`:

```ts
const STATUS_VERDICT: Record<string, Verdict> = { pending: 'pending', approved: 'allowed', rejected: 'denied' }
STATUS_VERDICT[status] ?? 'pending'
```

Two defects in one line.

**1. Prototype-chain fallthrough.** An object literal inherits from `Object.prototype`, so a status of `constructor`, `toString`, `valueOf`, or `__proto__` resolved to the *inherited member* rather than `undefined`. The `??` guard never fired and a function object flowed onward as a `Verdict` — reproduced by the reverted-code run, which crashes inside `VerdictChip` at `VERDICT_META[verdict].label`. `ApprovalResponse.status` is typed in `openapi/v1.yaml` as a bare `string` with no enum, so nothing upstream constrains what can arrive here.

**2. A silent `'pending'` default.** Any unrecognised-but-harmless status was reported as PENDING — a positive claim that the approval is still awaiting a human, indistinguishable on screen from one that genuinely is.

Fixed by converting the lookup to a `Map` (no inherited keys) and returning `Certain<Verdict>` with no default. An unrecognised status now renders the shared `AbsenceMarker`.

**Why `unknown` and not another truthfulness state:** the request *succeeded* and carried a value, so it is not `unavailable`; a status exists, so something did evaluate this approval and `not-evaluated` would be false; the backend does supply the field, so `not-supported` (which tells the operator to stop looking) is wrong. `unknown` — "we asked and could not determine the value" — is exactly the situation: we received a value we cannot interpret. The raw status reaches the operator through the marker's tooltip and screen-reader sentence rather than being swallowed.

This follows the precedent set by PR #1721 / AAASM-5109 for the identical shape in `features/trace/decision.ts` — same `Map`, same `Certain<T>` return, same "no default" rule, same `AbsenceMarker` rendering that `DecisionExplainer` already uses for an underivable verdict. No second convention was invented.

**Sweep:** `statusVerdict` was the only lookup in this file keyed by wire data. The remaining uses of `approval.status` are equality comparisons (`status === 'pending'` for the SLA section, `status !== 'pending'` for the footer's `disabled`), which carry no prototype hazard — and the footer's form is already the fail-closed direction, so an uninterpretable status leaves the decision buttons disabled.

Scope is deliberately narrow: the component is a pre-build awaiting AAASM-5095 and is not restructured here.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

All three documented statuses map exactly as before. Only previously-undefined behaviour changes.

## Related Issues

- Related Jira ticket: [AAASM-5190](https://lightning-dust-mite.atlassian.net/browse/AAASM-5190)
- Related GitHub issues: N/A

## Testing

Three new cases in `ApprovalDetailDrawer.test.tsx`, each verified to fail against the pre-fix component (restored via `git checkout remote/main -- <path>`, run, restored):

| Test | Reverted-code result |
|---|---|
| Inherited keys `constructor` / `toString` / `valueOf` / `__proto__` do not resolve to a verdict | ✗ ×4 — `TypeError: Cannot read properties of undefined (reading 'label')` in `VerdictChip` |
| Unrecognised status renders the `unknown` absence, not PENDING | ✗ — found `<span data-testid="verdict-chip" data-verdict="pending">⏸ PENDING</span>` |
| Every recognised status still maps exactly as before | ✓ before and after (this is the no-regression guard — it must pass in both directions) |

`5 failed | 11 passed` reverted → `16 passed` with the fix.

**Browser verification.** A route-level e2e is not possible: the drawer has no production mount (its approval-review flow is backend-blocked on AAASM-5095) and fabricating one would assert coverage that does not exist. Storybook-based verification *was* practical, following the existing `verify-aaasm-5075.spec.ts` pattern — `tests/e2e/verify-aaasm-5190.spec.ts` on port 4617 drives both head states and asserts on the rendered DOM (marker visible with `data-truth-state="unknown"`, raw status in the tooltip, zero verdict chips), plus the recognised-pending control. 4/4 passed; screenshots in `dashboard/verify/5190/` for both themes. Unit coverage remains the primary evidence.

Gates, each run once from `dashboard/`: `pnpm test` → 263 files / 2815 tests passed · `pnpm type-check` → clean · `pnpm lint` → no issues · `pnpm generate:api` → `git status --porcelain` empty.

## Checklist

- [x] Code follows project style guidelines (`cargo fmt`, `cargo clippy`) — dashboard-only change; `pnpm lint` / `pnpm type-check` clean
- [x] Documentation updated if behaviour changed — the WHY is recorded on the `Map` and on `statusVerdict`
- [ ] All CI checks passing

[AAASM-5190]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232 (commented):
# Adversarial review — PR #1724 (AAASM-5190)

**Reviewed head SHA: `30e4be491f543d733a523d0587e7fbee7c8b17e6`** (branch `v0.0.1/AAASM-5190/fix/drawer_proto`, 5 commits, 9 files, +215/−8, base `main`).
This is a **COMMENT review, not an approval** — I did not author this PR and I am not merging it.

## Verdict: **READY TO MERGE**

The defect is real, the fix closes it categorically, the evidence is genuine, and nothing was weakened to get green. Every finding below is MINOR / NIT / PRE-EXISTING — none blocks. I re-derived every claim rather than taking the PR body's word for it, and two of the PR's statements turned out to be *under*-evidenced rather than wrong (MINOR-1).

---

## Jira scope coverage — AAASM-5190

The ticket asks for exactly three things. All three are implemented in production code, none is promised only in a fixture, story, or comment:

| AC (from ticket) | Where | Status |
|---|---|---|
| Convert the lookup to a `Map` (no prototype-chain fallthrough) | `ApprovalDetailDrawer.tsx:49` `VERDICT_BY_STATUS = new Map<string, Verdict>` | ✅ |
| Drop the silent `'pending'` default for an explicit absence from the truthfulness vocabulary | `:77-81` `statusVerdict()` returns `Certain<Verdict>`, no default | ✅ |
| Follow the PR #1721 / AAASM-5109 precedent | Same `Map`, same `Certain<T>`, same no-default rule, same `AbsenceMarker` — verified against `features/trace/decision.ts` (see cross-PR) | ✅ |
| "Do not delete this component" | Component fixed in place, not deleted | ✅ |

**The narrow scope is legitimate, not concealment.** The ticket's own "Fix — in place, do NOT delete the component" section explicitly forbids restructuring, and records that a prior deletion recommendation was withdrawn after verification. Nothing in the ticket is left unimplemented.

I verified the ticket's premise independently rather than assuming it: `openapi/v1.yaml:4511-4513` types `ApprovalResponse.status` as `type: string` with **no `enum`**, and it is in the `required` list. So an unmodelled fourth value genuinely can arrive — the bug is reachable, not hypothetical.

---

## ADR assessment (with statuses)

A **Proposed** ADR is evidence only, never implementation authority. Classified accordingly:

| ADR | Status (verbatim) | Bearing on this PR |
|---|---|---|
| **ADR-0017** Dashboard Design-Parity — Ratified Evolutions | **Accepted** | **Item 11** reads: *"Shipped uses a single generic right-side `TraceDrawer` shell; the mock's per-variant `ApprovalDetailDrawer` is superseded."* The ticket's reading — that this is a statement about **the design mock**, not a decision to delete this file — is **textually correct**: item 11's grammatical subject is *the mock's* drawer, and ADR-0017's framing rule is that shipped `dashboard/src/` is authoritative over the mock. Item 11 does not authorise deleting shipped code. See ADR-DECISION-1 for the residue. |
| **ADR-0026** Seven Open Dashboard Product-Semantics Decisions | **Proposed, except Decision 2 — Accepted 2026-07 (option (A))** | **D2 is the accepted part and it supports this PR**: *"show only states the current projection can actually emit · include unconfigured / unknown where applicable · … do not preserve aspirational states merely to match an old mock."* Rendering `unknown` for an uninterpretable status is squarely inside that. ADR-0026's shared-context rule — *"The default answer to every question below, absent a product decision, is `—`"* — is exactly what this PR now renders. The other six decisions remain Proposed and are not relied on. |
| **ADR-0027** The Accessibility Floor Overrides the Visual Specification | **Accepted** | Binding: no surface ships below 4.5:1 normal / 3:1 large. **Verified against the actual rendered pixels — passes in both themes** (numbers below). ADR-0027 also establishes that contrast assertions live in the **unit** suite, "deliberately not in the Playwright e2e suite, which no CI job currently executes" — which corroborates PRE-EXISTING-3. |
| **ADR-0019** Agent Trust-Score Derivation | **Proposed** (requires product sign-off) | Not applicable. Evidence only. |
| **ADR-0020** Rolling vs Calendar Monthly Budget Windows | **Proposed** (requires product + architecture sign-off) | Not applicable. Evidence only. |
| **ADR-0021** Topology Enforcement-Mode Mutation | **Proposed** (requires architecture + security sign-off) | Not applicable. Evidence only. |
| **ADR-0022** Agent-Detail Config Projection | **Proposed** (recommendation half requires product sign-off) | Not applicable. Evidence only. |
| **ADR-0023** Is `aa-api` Meant to Carry a Policy Cascade? | **Proposed** (requires architecture sign-off) | Not applicable. Evidence only. |
| **ADR-0024** Semantics of an Empty or Unavailable Policy Cascade | **Proposed** (requires product + architecture sign-off) | Closest doctrinal parallel — *"permission is never inferred from the absence of policy data"*, and a reporting surface *"can only fail closed by declining to answer."* This PR does exactly that. **Cited as corroboration only, not authority**, because it is Proposed. |
| **ADR-0025** (visual-fidelity baseline) | **Proposed** | Item 4 would require future light/dark evidence to be captured against the **v2** prototype. **Proposed, so not binding** — and inapplicable anyway: these screenshots are of shipped code (which ADR-0017 makes authoritative), not a mock-fidelity claim. |
| **ADR-0015** (enforcement-side equivalent) | **Accepted** | §4: *"failure can never be laundered into a grant."* This PR is the presentation-layer analogue. Consistent. |

**No ADR governs "never key a lookup by wire data with an object literal."** That rule currently exists only as code precedent in two files. See ADR-DECISION-2.

---

## CI / Sonar / Codecov evidence — all re-fetched against `30e4be49`

- **Check-runs on `30e4be49`: 16 success, 23 skipped, 0 failed, 0 cancelled.** Confirmed independently via `/commits/30e4be49.../check-runs`. **0 check-runs on any other SHA**; `30e4be49` is the last of the PR's 5 commits.
- The **relevant** jobs genuinely ran rather than being skipped into a false green: `Dashboard tests + coverage` ✅, `Dashboard (type-check + lint)` ✅, `Dashboard build` ✅, `Dashboard codegen drift check` ✅, `aa-cli build compat (dashboard embed)` ✅, `CodeQL` + `Analyze (javascript-typescript)` ✅, `codecov/patch` ✅. The 23 skips are all Rust/proto/SDK/eBPF jobs correctly routed off by the paths filter for a dashboard-only diff.
- **SonarCloud**: Quality Gate passed · 0 New issues · 0 Accepted · 0 Security Hotspots · **100.0% Coverage on New Code** · 0.0% Duplication. Comment updated `02:39:24Z` vs head committed `02:24:23Z` — **the analysis postdates the head, so it is not stale**. Confirmed.
- **Codecov**: *"All modified and coverable lines are covered by tests."*
- No prior reviews exist; the only PR comments are from `codecov[bot]` and `sonarqubecloud[bot]`.

---

## Test evidence — reproduced locally, not taken on trust

**The new tests are genuine regression tests.** I restored the pre-fix component from `remote/main`, re-ran, and restored:

```
pre-fix  (remote/main component + new tests):  5 failed | 11 passed (16)
post-fix (as committed):                       16 passed (16)
```

Exactly the `5 failed | 11 passed` → `16 passed` the PR body claims. Failure mode on the reverted code matches the ticket's repro — a `data-verdict="pending"` chip rendered for `status: 'escalated'`, and a `TypeError` inside `VerdictChip` for the inherited keys. Worktree restored to `30e4be49`, `git status --porcelain` empty.

**I proved the underlying defect independently** rather than inferring it. All eight `Object.prototype` member names fall through the old `STATUS_VERDICT[status] ?? 'pending'`, and `Map.get` returns `undefined` for every one:

| wire `status` | old object lookup + `??` | `Map.get` |
|---|---|---|
| `constructor` | `Object` (function) | `undefined` ✅ |
| `toString` / `valueOf` / `hasOwnProperty` / `isPrototypeOf` / `propertyIsEnumerable` / `toLocaleString` | function | `undefined` ✅ |
| `__proto__` | `[object Object]` | `undefined` ✅ |
| `escalated` | `'pending'` ← the silent lie | `undefined` ✅ |

**The `Map` closes prototype pollution categorically**, for all eight names — not just the four the tests enumerate.

---

## Regression assessment

- **`git diff --numstat remote/main..30e4be49` on the test file: `47 0`** — 47 additions, **zero deletions**. I grepped the diff for removed lines (`^-[^-]`): none. **No existing test was weakened, deleted, or had an assertion relaxed.**
- **Every recognised status still maps exactly as before**, and this is pinned by the new `it.each`: `pending→pending`, `approved→allowed`, `rejected→denied`. That triple passes both pre- and post-fix, which is what makes it a real no-regression guard rather than a tautology.
- Only previously-undefined behaviour changes. `VerdictChip`, `Verdict`, and the `ApprovalActions` footer contract are untouched.
- `pnpm generate:api` drift check green — no schema churn.

---

## Playwright / visual evidence

Run **myself**, not read from the PR body. Isolated config on port **4621** with `reuseExistingServer: false` so a fresh Storybook was compiled — **4/4 passed**. (The stale-`dist` hazard does not apply here: this harness serves `storybook dev`, which compiles from source, not `vite preview` over a prebuilt `dist`. Worth noting the PR's own config uses `reuseExistingServer: true`, which *would* be exposed to a stale server left running by another lane on :4617 — not a defect in the committed evidence, since I reproduced it on a clean server.)

**The committed screenshots are genuine and current.** I regenerated all four from `30e4be49` and compared:

| file | SHA-256 | pixel diff |
|---|---|---|
| `recognised-pending-{light,dark}.png` | **identical** | **0 / 1,254,960 subpixels** |
| `unrecognised-status-{light,dark}.png` | **identical** | **0 / 1,254,960 subpixels** |

Byte-for-byte identical to my independent regeneration. Restored afterwards; tree clean.

**WCAG AA — measured from the actual rendered pixels** (ADR-0027 floor, 4.5:1 normal text):

| element | colour on ground | ratio | |
|---|---|---|---|
| light — `—` glyph | `#8a5a00` on `#ffffff` | **5.93:1** | PASS |
| light — `UNKNOWN` label | `rgb(149,106,24)` on `#ffffff` | **4.82:1** | PASS |
| dark — `—` glyph | `#e6b86b` on `#1c1a16` | **9.46:1** | PASS |
| dark — `UNKNOWN` label | `rgb(210,168,98)` on `#1c1a16` | **7.87:1** | PASS |

Both themes verified as genuinely light and dark (distinct grounds, `#ffffff` vs `#1c1a16`). The label is the tightest because `.truth-absent__label` carries `opacity: .9`; on the drawer's white ground it clears the floor with room. *(Advisory for whoever reuses `AbsenceMarker` on a tinted ground: the same label on `--paper` `#f5f4f0` computes ≈4.42:1 and would **fail**. Not this PR's problem — the drawer is on `--paper-2` — but the margin is thinner than it looks.)*

---

## Business-logic assessment — is `unknown` the right state?

**Yes.** I checked the reasoning against all six `TruthState` definitions in `lib/truthfulness/absence.ts` rather than accepting the PR's three-way argument:

| state | definition | verdict |
|---|---|---|
| `unavailable` | "The request failed; we have no answer at all" | ❌ the request **succeeded** and carried a value |
| `not-evaluated` | "No evaluation has been performed, so there is no verdict" | ❌ a status **exists** — something evaluated it; we just cannot read the answer. Claiming otherwise would be a false statement about the governance queue |
| `not-supported` | "The backend genuinely cannot provide this; waiting will not help" | ❌ the backend **does** supply the field (`required` in the schema); this would wrongly tell the operator to stop looking |
| `unconfigured` | "capability exists but nothing configured for it yet" | ❌ not a configuration question |
| `demo` | fixture/illustrative | ❌ this is live wire data |
| **`unknown`** | **"We asked and could not determine the value"** | ✅ **exact fit** |

The PR's stated reasoning holds and is complete. `unknown` carries tone `caution` (amber), which is the right register — it is an anomaly the operator should notice, not a fault.

**The raw status does reach the operator**, through two independent channels, neither of which swallows it:
- **tooltip**: `title="Unknown — Unrecognised approval status \"escalated\""` (asserted by both the unit test and the e2e spec)
- **screen-reader sentence**: `.truth-sr-only` renders `"Unknown — this value could not be determined. Unrecognised approval status \"escalated\"."`

---

## Security assessment

- **Prototype pollution: closed, categorically.** `Map` has no inherited string keys; verified for all 8 reachable `Object.prototype` names, not just the 4 tested.
- **No fabricated success/verified/allowed state.** The change strictly *removes* a confident claim (`'pending'`) and replaces it with an explicit absence. There is no direction in which this PR invents a safer-looking answer.
- **No injection.** The raw wire status reaches only a `title` attribute and text nodes, both React-escaped. No `dangerouslySetInnerHTML` anywhere in the path.
- **Footer stays fail-closed.** `disabled={approval.status !== 'pending'}` — an uninterpretable status (`'escalated'`, `'constructor'`, …) is `!== 'pending'`, so approve/reject remain disabled. Correct — but see **MINOR-1**, because nothing actually *asserts* this.
- **No test-only behaviour leaking to production.** The `testId` prop on `AbsenceMarker` is a pre-existing display-layer affordance of the shared component; no fixture, mock, or demo path is reachable from production code in this diff.

**In-file sweep — the PR's claim is accurate.** I re-ran it rather than trusting it. The only surviving uses of `approval.status` in the file are `=== 'pending'` (SLA gate, `:179`) and `!== 'pending'` (footer, `:206`) — string equality, which carries no prototype hazard. **No lookup keyed by wire data survives in this file.** `buildRows` reads fields directly; `key={row.label}` uses hardcoded labels. **This PR did not repeat #1721's boundary error inside its own file.**

---

## Cross-PR notes

- **#1721 / AAASM-5109 precedent is genuinely in place**, not merely cited: `features/trace/decision.ts` has *both* wire-keyed tables as `Map`s (`VERDICT_BY_DECISION:89`, `VERDICT_BY_OPERATION:123`), with the same rationale in the docblock. This PR reuses that exact shape — same `Map`, same `Certain<T>`, same no-default rule, same `AbsenceMarker` rendering. **No second convention was invented.** Verified.
- **`VerdictChip` is safe** and did not need changing: it indexes `VERDICT_META` by its `verdict: Verdict` prop, and all three call sites feed it a value that came out of a `Map`-backed deriver.
- **However — the class is not eradicated repo-wide.** See PRE-EXISTING-1. This PR is correctly scoped and I am **not** asking it to grow; but #1721 and #1724 are now two consecutive per-file sweeps of the same defect, and the pattern keeps regenerating.

---

# Findings

### MINOR-1 — the fail-closed footer is the one safety claim nothing verifies
The PR body leans on this in its sweep paragraph ("the footer's form is already the fail-closed direction, so an uninterpretable status leaves the decision buttons disabled"). The behaviour **is correct** in code. But:
- **No unit test asserts it.** `ApprovalDetailDrawer.test.tsx` contains no `disabled` assertion at all.
- **The screenshots cannot evidence it either.** The stories omit the `GrantScopes` wrapper the unit tests use, so `usePermissions().canWrite` is `false` and `ApprovalActions` computes `actionsDisabled = disabled || busy || !canWrite` → disabled in **both** stories. I confirmed this at the pixel level: the footer region is **byte-identical** between `recognised-pending-light.png` and `unrecognised-status-light.png` (3617px of `rgb(138,138,138)` in each). The captures show a disabled footer regardless of status, so they neither confirm nor refute the claim.

Net: the governance-relevant half of this fix — *an uninterpretable status must not be actionable* — currently rests on code reading alone. A future "tidy-up" rewriting `status !== 'pending'` into `status === 'approved' || status === 'rejected'` would silently **enable** approve/reject for an uninterpretable status, and no check would catch it. One line in the new `describe` block closes this permanently:

```ts
it('leaves the decision footer disabled for an uninterpretable status', () => {
  renderWithClient(
    <ApprovalDetailDrawer approval={makeApproval({ status: 'escalated' })} onClose={vi.fn()} />,
  )
  expect(screen.getByRole('button', { name: /approve/i })).toBeDisabled()
})
```
(wrapped in `GrantScopes`/`WRITE_SCOPES` as the sibling tests already are, so the assertion is about the status and not about scope).

### MINOR-2 — the inherited-key net covers 4 of 8 reachable prototype names
`it.each(['constructor','toString','valueOf','__proto__'])` omits `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, and `toLocaleString`, all four of which I confirmed fall through the pre-fix lookup identically. **There is no behavioural gap** — the `Map` closes all eight by construction — so this is completeness of the regression net, not a defect. Adding the four names to the existing `it.each` array is a four-token change and makes the test document the whole hazard.

### NIT-1 — the screen-reader sentence is unasserted
The test pins `title` contains `escalated`, but not the `.truth-sr-only` sentence. Since the `title` tooltip is **mouse-only** (the marker is a non-focusable `<span>`), the SR sentence is the only non-mouse path to the raw status — arguably the more important of the two to pin. `expect(marker).toHaveTextContent('escalated')` would cover both.

### NIT-2 — `UnrecognisedStatus` story sets `expires_at: ''`
Mirrors the existing `Decided` story, so it is consistent and harmless — the SLA block is gated on `status === 'pending'` and would be hidden either way. Noting only because it means the evidence does not surface a real (correct, pre-existing, unchanged) consequence: an uninterpretable status also **hides the SLA countdown**, so the operator loses the deadline as well as the verdict. That is the right call under the truthfulness rule — we cannot claim it is still counting down — but it is a behaviour worth being deliberate about when 5095 lands.

### PRE-EXISTING-OUT-OF-SCOPE-1 — the defect class survives in 12 other places (**does not block this PR**)
This PR's file is clean and its scope is correct per the ticket. But a full sweep of `dashboard/src` for the identical shape found **12 remaining instances**, three of which are more severe than the one just fixed:

- **`components/agentDetail/AgentDecisionStream.tsx:26,42-43`** — closest sibling. `VERDICT_BY_LABEL[row.decisionLabel]` where `AgentDecisionResponse.decisionLabel` is a bare wire `string`. `"constructor"` → truthy `Object` → `VERDICT_META[Object]` is `undefined` → **`.colorVar` throws and unmounts the Traffic tab**.
- **`features/audit/logs.ts:376-378,431-432`** — `DETAIL_SUMMARISERS[kind]` where `kind` comes from `JSON.parse` of the audit `payload`. A wire string **selects a function that is then invoked**: `kind:"constructor"` → `Object(detail)` returns the detail object, which is then wrapped as `known(summary)` typed `Certain<string>` — an object masquerading as an audit summary.
- **`features/alerts/parseAlert.ts:51,71,111,121`** — `.toLowerCase()` narrows the exploitable set to `constructor`/`__proto__`, but both still return truthy inherited members, **bypassing the documented `throw`**, and the corrupted value flows into `StatusBadge`, `SeverityBadge`, `AlertCardFeed`, `OverviewPage`, and both `AlertList` sort comparators.

Plus nine MEDIUM/LOW (`PolicySimulatePanel`, `RoleCapabilityCards`, `AgentRegistryList`, `TeamMembersCard`, `TeamOrphanDetail`, `TeamDetailPage`, `AuditLogPage`, `features/agents/api.ts`, `features/audit/export.ts`). Happy to hand over the full annotated list with wire paths.

**Recommend one Epic** rather than a third per-file PR — the file-at-a-time approach has now run twice and the class keeps reappearing.

### PRE-EXISTING-OUT-OF-SCOPE-2 — `AbsenceMarker`'s tooltip is unreachable by keyboard
The marker is a non-focusable `<span>` with `cursor: help`; `title` fires on hover only. A **sighted keyboard-only operator** cannot reach the raw status — screen-reader users can (via `.truth-sr-only`), and mouse users can. Property of the shared AAASM-5173 component, affects every call site, not introduced here.

### PRE-EXISTING-OUT-OF-SCOPE-3 — the e2e evidence can never re-run
`grep -rn playwright .github/workflows/` returns **zero hits** — no CI job executes any Playwright spec (independently corroborated by ADR-0027's note). So `verify-aaasm-5190.spec.ts` is one-shot local evidence that cannot regression-detect. Additionally, the default `playwright.config.ts` (`testDir: 'tests/e2e'`, no `testMatch`) *would* pick this spec up and fail it against `vite preview` on :4173, which does not serve Storybook. **This PR conforms exactly to an established convention** — 87 spec files and 24 per-ticket standalone configs already do this — so it is not this PR's to fix, and the PR is honest that "unit coverage remains the primary evidence." Flagging the systemic issue only.

### ARCHITECTURE-DECISION-REQUIRED-1 — no ADR governs the wire-keyed-lookup rule
No ADR mentions prototype pollution, `__proto__`, `Object.create(null)`, or untrusted-key lookups. The rule now exists only as code precedent in two files, with the rationale **duplicated verbatim in two docblocks** — which is how conventions drift. Given PRE-EXISTING-1's 12 survivors, this warrants either a short Accepted ADR or, better, a lint rule (`no-restricted-syntax` on `Record<string, …>` used as a lookup) that makes the class mechanically unable to regenerate. Non-blocking.

### ARCHITECTURE-DECISION-REQUIRED-2 — ADR-0017 item 11 leaves this component's disposition unrecorded
The ticket's reading of item 11 is textually correct (its subject is the mock). But item 11 does ratify the generic `TraceDrawer` shell as the authoritative container, and the shipped `ApprovalDetailDrawer.tsx` has **no route and no lazy import** — its own docblock says the remaining affordances are backend-blocked on AAASM-5095. Fixing rather than deleting is unambiguously right *for this ticket*; the disposition question (fold into `TraceDrawer` vs. mount as-is) belongs to **AAASM-5095**. Flagged so it is not lost, since the earlier deletion recommendation was withdrawn without recording the alternative. Non-blocking.

---

## Unresolved risks

1. **The fail-closed footer is unpinned** (MINOR-1) — the only genuinely load-bearing gap in this PR. Everything else is documentation or completeness.
2. **The defect class is live in 12 other files**, 3 of them crash-or-corrupt severity (PRE-EXISTING-1). Fixing this drawer does not reduce that exposure.
3. **No mechanical guard** stops the next `Record<string, X>` keyed by wire data from landing (ADR-DECISION-1).

## Recommended follow-up tickets (none filed yet — say the word and I'll open them)

| # | Type | Summary |
|---|---|---|
| A | **Epic + Bugs** | Eradicate wire-data-keyed object-literal lookups across `dashboard/src` — 12 sites, 3 HIGH (`AgentDecisionStream`, `audit/logs.ts`, `alerts/parseAlert.ts`). Full annotated list available. |
| B | Task | Add an ESLint rule (or ADR) forbidding `Record<string, …>` as a lookup keyed by wire data; require `Map`. Closes the class mechanically. |
| C | Subtask on **AAASM-5095** | Decide `ApprovalDetailDrawer`'s disposition against ADR-0017 item 11's generic `TraceDrawer` ratification, and mount or fold accordingly. |
| D | Task | Make `AbsenceMarker`'s detail keyboard-reachable (focusable + `aria-describedby`, or a popover) — shared AAASM-5173 component, all call sites. |

---

**Summary.** The defect was real and reachable, the `Map` closes it categorically for all eight inherited keys, `unknown` is demonstrably the correct vocabulary member, every recognised status still maps exactly as before, no existing test was weakened (47 additions / 0 deletions), the committed screenshots are byte-identical to an independent regeneration, and the changed text clears the ADR-0027 contrast floor in both themes. My only substantive ask is **MINOR-1**: add one assertion so the fail-closed footer — the safety property the PR's own rationale depends on — is pinned by something other than a code reading.

*Reviewed head SHA: `30e4be491f543d733a523d0587e7fbee7c8b17e6`. Comment review only — no approval, no merge.*

Chisanan232 (commented):
# Follow-up — MINOR-1 closed

**New head SHA: `9aa7213ffdb2e75f198a81a9f5390f3732339b95`.** My earlier review was against `30e4be491f543d733a523d0587e7fbee7c8b17e6`, so that SHA is now stale.

**Delta since the reviewed SHA is only the MINOR-1 test addition — nothing else.** Verified:

```
$ git diff --stat 30e4be49..9aa7213f
 dashboard/src/components/trace/ApprovalDetailDrawer.test.tsx | 17 +++++++++++++++++
 1 file changed, 17 insertions(+)
```

One commit (`9aa7213f`), one file, test-only, zero production-code change. No source file, story, config, or screenshot was touched. Everything in my original review still stands unmodified.

## What MINOR-1 was

The head's honesty was only half the fix. The other half — an uninterpretable status must not be *actionable* — was correct in code (`disabled={approval.status !== 'pending'}`) but **asserted nowhere**, and the Storybook captures could not evidence it either: the stories omit `GrantScopes`, so `actionsDisabled = disabled || busy || !canWrite` is dominated by `!canWrite` and the footer renders disabled in **both** stories. I confirmed that at the pixel level — the footer region is byte-identical between `recognised-pending-light.png` and `unrecognised-status-light.png` (3617px of `rgb(138,138,138)` in each).

Now pinned in the unit suite, where `renderWithClient` already grants `WRITE_SCOPES`, so `actionsDisabled` collapses to exactly the status gate under test.

## Mutation-verified, not revert-verified — please record it that way

Reverting the AAASM-5190 source fix leaves this assertion **passing** (`1 passed | 16 skipped`), because `disabled={approval.status !== 'pending'}` was never part of that fix. It is a characterisation guard for pre-existing behaviour, not a regression test for the fix, so a revert cannot exercise it.

The check that proves it load-bearing is a **mutation of the guard itself** — the exact plausible-looking refactor the test exists to catch:

```diff
-  disabled={approval.status !== 'pending'}
+  disabled={approval.status === 'approved' || approval.status === 'rejected'}
```

→ **1 failed** at `expect(screen.getByTestId('approval-approve-btn')).toBeDisabled()`. Restored afterwards; tree clean.

## Evidence on `9aa7213f`

- **Check-runs: 16 success, 23 skipped, 0 failures** — the same profile as the reviewed SHA. `Dashboard tests + coverage` ✅, `Dashboard (type-check + lint)` ✅, `Dashboard build` ✅, `Dashboard codegen drift check` ✅, `codecov/patch` ✅, `SonarCloud analysis` ✅.
- **SonarCloud on the new head**: Quality Gate passed · **0 New issues** · 0 Accepted · 0 Security Hotspots · **100.0% Coverage on New Code** · 0.0% Duplication. Comment updated `04:32:54Z`, i.e. **after** this commit — not a stale carry-over from the previously reviewed SHA.
- **Codecov**: "All modified and coverable lines are covered by tests."
- Local gates, each once: `pnpm test` → 263 files / **2816** tests passed (2815 + 1) · `pnpm type-check` clean · `pnpm lint` clean · `pnpm generate:api` → no drift.

## Verdict unchanged: READY TO MERGE

MINOR-1 was the only substantive ask in my review and it is now closed. MINOR-2 (the inherited-key `it.each` covers 4 of the 8 reachable prototype names) and NIT-1/NIT-2 remain open and remain non-blocking — the `Map` closes all eight by construction, so those are completeness of the regression net, not defects.

The PRE-EXISTING findings are being tracked separately as their own Epic (18 wire-keyed lookup sites across the dashboard, including two that are **worse than what this PR fixed** and reachable in production, where this drawer has no production mount). None of that blocks this PR.

Still a **comment review, not an approval** — the review-required gate is untouched.

*Reviewed head SHA: `9aa7213ffdb2e75f198a81a9f5390f3732339b95` (previously `30e4be491f543d733a523d0587e7fbee7c8b17e6`).*

PR #1726, opened 2026-07-27
[AAASM-5184] ♻️ (dashboard): Stop rendering the unnamed-team group as a team

Chisanan232:
## Description

Topology treated the group holding team-less agents as if it were a team. `mapGraph` coerced a null `team_id` to `''` (`mapGraph.ts:80`), and every consumer downstream then read that empty string as a team name, producing three visible false statements: the header counted a team that does not exist, the sidebar offered a filter row with an empty label, and the canvas drew a nameless cluster.

**Which layer owns this defect — the dashboard, not the backend.** The Topology page calls `GET /api/v1/topology` (`dashboard/src/features/topology/api.ts:63`), whose `TopologyGraphResponse` carries only `nodes` + `edges` — no team list and no `team_count` (`openapi/v1.yaml`). `project_graph_nodes` (`aa-api/src/routes/topology.rs:405-441`) emits a flat node vec and passes `team_id: Option<String>` through untouched; it never groups. The grouping and the count are built entirely client-side at `TopologyPage.tsx:53-54`. The backend's two `team_count` computations (`topology.rs:579` for `/topology/overview`, `:961` for `/topology/stats`) already skip team-less agents via `if let Some(tid) = &r.team_id`, and neither feeds this page. So this is genuinely the dashboard half, not a duplicate of the backend fix.

**Before → after** (two agents, one on team `support`, one with `team_id: null`):

| | before | after |
|---|---|---|
| header | `· 2 agents · 2 teams` | `· 2 agents · 1 team` |
| sidebar filter rows | `◎All teams`, `○` (blank), `○support` | `◎All teams`, `⚠ Unclaimed`, `○support` |
| sidebar stats | — | `⚠ 1 unclaimed` |
| canvas cluster | unlabelled box | `⚠ UNCLAIMED`, warn-toned |
| cluster click | nothing (inert since AAASM-5140) | opens its own detail panel |
| JSON export | `"team": ""` | `"team": null` |

A team-less agent is a **real, known state**, so it is named and rendered rather than given an absence marker — `NO_DATA` would claim we could not determine the team, when we know precisely that no team claims it. `features/topology/unclaimed.ts` holds the single definition, and its membership predicate **defers to `isOrphanAgent`** from `features/teams/orphans.ts` (AAASM-5157) rather than forking a second `team_id == null` test — a unit test asserts the two agree for every input. The group key is `__orphan__`, matching `design/v2/hi-fi/topology.jsx` (authoritative per ADR-0025) at `:190`, `:664`, `:928`.

Scope call on the design's `__orphan__` spec: the cluster treatment (`:190`), detail panel (`:664-686`), `⚠ N unclaimed` stat (`:952`) and orphan-excluding `teamCount` (`:928`) all ship. The **"Suspend all unclaimed" bulk action (`:765`) deliberately does not** — it is a mutation, not part of this truthfulness defect, and warrants its own ticket.

The detail panel omits the cross-team edge count for this group: an edge touching an agent with no team is not a boundary crossing, which is the gateway's own rule (`aa-api/src/routes/topology.rs:259`, mirrored by `/topology/edges`'s `is_cross_team`). Reporting a number there would have the panel contradict both the gateway and the sidebar's fleet-wide counter.

### Dependency — Lane B / AAASM-5181-5182

AAASM-5182 fixes the **backend** half of the same defect class in `aa-api/src/routes/topology.rs`. This PR touches **no Rust** and no OpenAPI schema, so the two do not conflict and neither blocks the other. Confirmed above that the two fixes address different, non-overlapping surfaces — the Topology page never reads a server-side team count — so there is no risk of two places deciding what "no team" means.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

`TopologyNode.team` no longer carries `''` for a team-less agent, but it is an internal view-model field with no external consumers; the JSON export is unwound back to the wire's `null`.

## Related Issues

- Related Jira ticket: [AAASM-5184](https://lightning-dust-mite.atlassian.net/browse/AAASM-5184)
- Stacked context: AAASM-5157 (PR #1715) established the `unclaimed` vocabulary this reuses; AAASM-5140 made the cluster inert and explicitly deferred the naming decision here.

## Testing

**Revert check** — every new test was verified load-bearing by reverting the source and watching it go red (`git checkout <ref> -- <paths>`, never `git stash`):

- Unit: with the seven consumer files reverted, **11 tests failed**, each on the specific claim it protects (header count, filter-row label, unclaimed stat, cluster label, panel copy, mapper, export). Restored → `264 files / 2834 tests` pass.
- E2E: a surgical mutation (`teamCount` counts the unclaimed group; label blanked) that still type-checks, with a **verified `pnpm build` exit 0 and a fresh `dist`** — so the pass was not read off a stale preview — turned **4 of 6** e2e cases red. The 2 that stayed green are the "every agent has a team" controls, correctly unaffected. Restored → 6/6 pass.

Playwright: repo-local `@playwright/test` on port 4612, driven from the wire shape (`team_id: null`), light **and** dark, console errors asserted `[]`. Screenshots committed to `dashboard/verify/5184/`.

Gates: `pnpm test` ✅ · `pnpm type-check` ✅ · `pnpm lint` ✅ (no issues) · `pnpm generate:api` → `git status --porcelain` empty ✅.

## Review follow-up — `fc3b6f44`

Adversarial review of `e85eae92` found one defect of the same class this PR closes, plus two smaller items. Fixed here.

### The blank-team hole (fixed)

`isOrphanAgent` lifts `team_id` through `certain`, which treats **only the exact empty string** as missing. So `team_id: "   "` was read as a real team: `teamKeyOf({team_id:'   '})` returned `'   '`, and all three false statements this PR fixes came straight back — inflated `N teams`, a filter row whose label renders blank, a nameless cluster.

That input is reachable. `AgentRegistry::validate_tenant_id` (AAASM-4190) rejects control characters only, which is exactly why **#1728 added `team_of` with `.trim().is_empty()`**. Left as-is, the two layers would have disagreed about what "no team" means the moment #1728 landed — the outcome the *Dependency* section above claims is impossible.

Fixed at the **shared definition** (`features/teams/orphans.ts`), not locally in topology, so the one-definition property this PR is built on is what got repaired:

```ts
isAbsent(certain(node.team_id?.trim(), 'unconfigured'))
```

`lib/truthfulness` is untouched — `'   '` is a real string and `certain` is right to say so for values in general; this is a domain rule about team ids. A real id is returned **unchanged**, not trimmed, so the group still keys on exactly what was registered — the same choice `team_of` makes. Blast radius is `selectOrphanAgents` → `TeamsPage`, where the trim also lines the unclaimed list up with the post-#1728 backend; no other consumer.

Covered at every level: `isOrphanAgent` (spaces / tab / newline, plus a padded id staying claimed), `teamKeyOf`, the mapper, the "these two agree for every input" test now spanning whitespace and padded ids, and a new light+dark e2e case driving `team_id: "   "` off the wire.

### SVG export leaked the sentinel (fixed)

`buildGraphJson` unwinds the group key to `null` because an export leaves the app and must not assert a team id no API returned. `serializeGraphSvg` serialises a clone of the live DOM, which still carried `data-team="__orphan__"` — the same fabrication in the other file format. The attribute is now dropped from the clone (not blanked: `data-unclaimed="true"` already carries the fact, and an empty `data-team` would be precisely the blank-team claim this ticket ends). Live DOM untouched, asserted.

### Coverage (fixed)

Codecov reported **92.85% patch — 4 lines missing + 1 partial, all `TopologyGraph.tsx`**, while Sonar reported "100.0% Coverage on New Code". Both are true under their own definitions; flagging so they are not conflated. The uncovered lines were real and reachable: `TeamCluster`'s keyboard handler and the pan-drag click suppression. Now covered by unit tests — Enter and Space activate the cluster, other keys don't, and a click concluding a pan does not select. Every line and branch in the diff region of `TopologyGraph.tsx` is covered locally.

### Known limitation (accepted, no defence built)

`UNCLAIMED_TEAM = '__orphan__'` shares a namespace with real team ids, and `validate_tenant_id` rejects control characters only — so `team_id` is **caller-supplied and could be `__orphan__`**. A tenant registering agents under that id gets its real team rendered as "⚠ Unclaimed", dropped from `N teams`, and given the "no team-scoped policy or budget applies" copy. Display-only, same-tenant, no authorization or data-exposure consequence. Recorded rather than defended: the alternative (widening `TopologyNode['team']` to `string | null`) pushes the same either/or into the force layout, the filter state and every `Map<string, …>` keyed by team, for a collision no registry has produced.

### Open, not settled in this PR

- **"unclaimed" names different populations across the FE/BE boundary.** This page counts every agent with no team at any depth; `/topology/stats.orphan_count` counts no-team **and `depth > 0`**; `/topology/overview.standalone_root_agents` counts no-team **and `depth == 0`**. The design's own panel copy (`design/v2/hi-fi/topology.jsx:686`) says the group mirrors `orphan_count`. The definition shipped here is the deliberate one from AAASM-5157, but the divergence is real and awaiting a decision — flagged, not resolved.
- **Design-tone divergences.** The cluster, filter row and panel note ship `var(--warn)` where `design/v2` uses danger red (`:190`, `:790`, `:686`); only the header stat (`:952`) matches the mock's tone exactly. Warn reads as the governance gap it is rather than a fault, but it is a divergence and ADR-0017 is the mechanism for recording one. **ADR-0025, which makes `design/v2` authoritative, is itself *Proposed*** — the mock's standing here rests on `design/README.md` ("Use **v2** as the current visual spec"), not on that ADR.
- **The unclaimed cluster still draws a team-shaped budget bar** (a sum of per-agent limits, as for real teams) while the panel says no team-scoped budget applies. Not false, but it is the one remaining place the group is presented exactly like a team. Filed rather than changed — it is a visual/product call, not a truthfulness bug.

### Re-verified at `fc3b6f44`

`pnpm test` **264 files / 2849 tests** ✅ · `pnpm type-check` ✅ · `pnpm lint` ✅ · `pnpm generate:api` → `git status --porcelain` empty ✅ · `rm -rf dist && pnpm build` **exit 0** with a fresh `dist` confirmed before serving, then the 5184 Playwright pass **8/8** (light + dark, including the new blank-team case). Screenshots regenerated against that build and committed.

## Checklist

[AAASM-5184]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232 (commented):
## Adversarial review — PR #1726 (AAASM-5184)

**Reviewed head SHA: `e85eae9214690313bccc50019fc375240442b920`** (branch `v0.0.1/AAASM-5184/fix/topology_blank_team`, base `main`, merge-base `dab8a70b0a27e5ce8363003b42a7144a8d601491`, 25 files, +823/−86, 9 commits). I did not author this PR. This is a **COMMENT review** — not an approval, not a merge.

### Verdict

**No BLOCKER.** The defect is real, the fix is in the right layer, the tests are load-bearing (I verified that by reverting, not by reading the claim), and every visible statement I could measure is now true. Two **MAJOR** findings stand between this and an approval, and the first is the same defect class this PR exists to close, so it belongs in this PR rather than a follow-up.

- **MAJOR-1** — a whitespace-only `team_id` still becomes a phantom team here, and post-#1728 the backend disagrees.
- **MAJOR-2 / ARCHITECTURE-DECISION-REQUIRED** — "unclaimed" now names two different populations across the FE/BE boundary, undisclosed.
- 5 MINOR, 4 NIT, 1 PRODUCT-DECISION-REQUIRED (tracking), 1 security-adjacent note.

---

### Did the "dashboard owns it, not the backend" analysis survive an independent check? — **Yes.**

I re-derived it from the source rather than from the PR body:

| claim | verified |
|---|---|
| `GET /api/v1/topology` returns only `nodes` + `edges` | `openapi/v1.yaml` → `TopologyGraphResponse.required = [nodes, edges]`. No team list, no `team_count`. ✅ |
| `project_graph_nodes` never groups | `aa-api/src/routes/topology.rs:405-441` — maps records → flat `Vec<AgentNode>`, sorts by id, enriches `policy_count` / `effective_permissions` / `budget`. `team_id` is passed through by `AgentNode::from(record)` and never read. ✅ |
| grouping + count are client-side | `TopologyPage.tsx:53-57` (`teams` from `new Set(allNodes.map(n => n.team))`, `teamCount`), seeded by `mapGraph.ts:80`. ✅ |
| the backend `team_count` sites don't feed this page | `get_overview` (`team_map`) and `get_stats` (`team_sizes.len()`); `features/topology/api.ts` issues only the `/topology` graph query. ✅ |
| not a duplicate of #1728 | #1728 touches exactly one file (`aa-api/src/routes/topology.rs`) and only `get_overview`, `get_stats`, the four cache keys, and the new `team_of` helper. It does **not** touch `get_graph` / `project_graph_nodes`. No file overlap, no semantic overlap on this page. ✅ |

So the layer split is correct and two layers are **not** both deciding what this page renders. The residue is that the two lanes still hold *different* definitions of "no team" — MAJOR-1 and MAJOR-2 below. That's a consequence of the split, not evidence against it.

---

### MAJOR-1 — a whitespace-only `team_id` still becomes a phantom team, and it disagrees with #1728

`teamKeyOf` defers to `isOrphanAgent`, which is `isAbsent(certain(node.team_id, 'unconfigured'))`. `certain` (`dashboard/src/lib/truthfulness/absence.ts:186`) treats **only** `null`, `undefined` and the exact empty string as missing — it does not trim. So:

```
teamKeyOf({ team_id: '   ' })            === '   '        // not __orphan__
mapTopologyGraph(...).nodes[0].team      === '   '
```

I ran that in this worktree at the head SHA (temporary probe spec, deleted; tree clean afterwards) — the assertion `expect(teamKeyOf(node({team_id:'   '}))).toBe(UNCLAIMED_TEAM)` fails with `expected '   ' to be '__orphan__'`.

Consequence: for `team_id: "   "` **all three** false statements this PR fixes come straight back — `N teams` is inflated by one, the sidebar offers a filter row whose label is blank whitespace, and the canvas draws a nameless cluster. And that input is reachable: `AgentRegistry::validate_tenant_id` (AAASM-4190) rejects control characters only, which is precisely why **PR #1728 added `team_of` with `.trim().is_empty()`** and documented it in that helper's own docstring.

That makes the PR body's assertion — *"there is no risk of two places deciding what 'no team' means"* — not hold once #1728 lands: the backend will group `"   "` as no team while this page groups it as a team named `"   "`.

Fix it at the shared definition (`certain`, or `isOrphanAgent`) rather than only in `teamKeyOf`, so the single-definition property the module is built around survives. `mapGraph.test.ts` covers `''`; add `'   '` and `'\t'`.

### MAJOR-2 / ARCHITECTURE-DECISION-REQUIRED — "unclaimed" now names two different populations

Three definitions coexist for one word:

| surface | rule |
|---|---|
| dashboard sidebar `⚠ N unclaimed` (this PR) | no team, **any** depth (`isOrphanAgent`, AAASM-5157) |
| `GET /topology/stats.orphan_count` | no team **and `depth > 0`** (`topology.rs:951-955`; #1728 only adds the trim) |
| `GET /topology/overview.standalone_root_agents` | no team **and `depth == 0`** |

The authoritative design's own panel copy says so explicitly — `design/v2/hi-fi/topology.jsx:686`: *"These agents have no team_id and depth > 0 … **Mirrors backend TopologyStats.orphan_count**."* This PR deliberately ships the broader, depth-agnostic definition (which I think is the **right** call — 5157 reasoned it through, and depth has nothing to do with whether team-scoped policy can reach an agent). But the effect is that the Topology screen and `/topology/stats` will report **different numbers for the same fleet**, an operator comparing them has no explanation, and the PR body doesn't disclose the deviation at all while presenting the design spec as shipped.

Not a blocker — the shipped number is the more complete one and this page never reads `orphan_count`. But it needs (a) a note in the PR/ticket, (b) an ADR recording which definition is canonical, and ideally (c) a backend follow-up exposing an `unclaimed_count` that matches. As written, the PR satisfies its own single-definition principle *within* the frontend and quietly breaks it across the boundary.

---

### ADR assessment (with statuses)

| ADR | Status | Bearing |
|---|---|---|
| **0017** Dashboard Design-Parity — Ratified Evolutions | **Accepted** | The mechanism this PR needed and didn't use: it records where the *shipped implementation* is authoritative over the mock and annotates the mock with a supersession note. See MINOR-2. |
| **0025** `design/v2/` Is the Authoritative Visual Spec | **Proposed** — "requires product/design sign-off … before it is treated as binding on future audits" | The PR cites it four times as *"authoritative per ADR-0025"*, i.e. as implementation authority. A Proposed ADR is evidence, never authority. The mock's real standing comes from `design/README.md` ("Use **v2** as the current visual spec") — cite that and note 0025's status. |
| **0026** Seven Open Dashboard Product-Semantics Decisions | **Proposed, except Decision 2 — Accepted 2026-07 (option A)** | States outright: *"Nothing in this ADR is implemented and nothing may be implemented from it"* except D2. D2 (capability legend / unknown state) doesn't govern this change. Reinforces that product-semantics choices need recorded sign-off — see PRODUCT-DECISION-REQUIRED. |
| **0027** The Accessibility Floor Overrides the Visual Spec | **Accepted** | AA floor on normal text. **Measured and met** — table below. |

The 0025 tension the ticket flagged is real and worth stating plainly: **0025 is the ADR that makes `design/v2` authoritative, and 0025 itself is Proposed.** This PR's design-fidelity argument therefore rests on `design/README.md`, not on 0025.

---

### Jira scope coverage — AAASM-5184

Ticket status is still **To Do** (should be In Progress / In Review). Components `agent-assembly` ✅. No Fix version set. The "PR opened" comment is present per the audit-trail convention ✅.

The description names five things `design/v2` specifies for the `__orphan__` group and says **"Which of those ship is a product call and should be settled on this ticket."**

| design item | shipped |
|---|---|
| distinct cluster treatment (`:190`) | ✅ (tone differs — MINOR-2) |
| dedicated detail panel + own copy (`:664-686`) | ✅ (copy differs — better; MAJOR-2 / MINOR-2) |
| warn-toned `⚠ N unclaimed` stat (`:952`) | ✅ exact, including the `> 0` gate |
| `teamCount` excluding orphans (`:928`) | ✅ |
| "Suspend all unclaimed" bulk action (`:765`) | ❌ excluded |

All three of the ticket's reported symptoms are fixed and covered. The AAASM-5157 reuse requirement — *"Topology must reuse that rather than fork a second definition"* — is honoured in substance (see NIT-2 for the caveat) and there is a test asserting the two predicates agree for every input, which I ran.

**On the exclusion (PRODUCT-DECISION-REQUIRED, tracking = MINOR-5):** excluding the bulk action is **legitimate**, not hidden scope. It is a mutation, it is the only one of the five that changes state rather than what the screen asserts, and shipping *less* than the spec is the conservative direction for a truthfulness defect. What's missing is the record: the ticket asked for a product call, the PR made it unilaterally citing a Proposed ADR, and **no follow-up ticket exists** — a Jira sweep across AAASM for `unclaimed` / `Suspend all` returns only 5183 (a different Teams-page defect) and the closed 5157/5080/5077/5071. So a specified design item is now unimplemented with nothing tracking it. File it.

---

### CI / Sonar / Codecov evidence

- **All checks on `e85eae92`: 16 success, 23 skipped, 0 failed. Zero checks on any other SHA.** Green: Dashboard tests + coverage, Dashboard build, Dashboard (type-check + lint), Dashboard codegen drift check, aa-cli build compat (dashboard embed), CodeQL ×4, Detect changed areas, CI Success, codecov/patch, SonarCloud ×2.
- **SonarCloud analysed SHA = `e85eae92`** (check completed `03:05:03Z`, comment `03:05:05Z`; head committed `02:55:43Z`) — **the analysis postdates the head commit**, so the gate is not stale. Quality Gate passed · 0 New issues · 0 Accepted · 0 Security Hotspots · 100.0% Coverage on New Code · 0.0% Duplication.
- **Codecov** `codecov/patch` on `e85eae92`: success, **"92.85% of diff hit (target 70.00%)"** — 4 lines missing + 1 partial, all in `TopologyGraph.tsx` (83.33%). Note the Codecov *comment* is timestamped `02:42:38Z`, i.e. **before** the head commit, but the check-run on head reports the same figure, so it isn't misleading. Flagging only so nobody reads Sonar's "100.0% Coverage on New Code" as "no uncovered new lines" — the two tools measure different line sets and both are true.
- **Sonar-smell refactor verified as behaviour-preserving.** `git diff 231b86c8..e85eae92` is a pure extraction: `TeamCluster` component + `clusterLabel()` + `filterMark()`. Every attribute, handler, tooltip string, class name and testid is identical to the inline version; `selected` is still `selectedTeam === c.team`; the inline `selectable = onTeamClick !== undefined` is now expressed per-attribute as `onTeamClick ? … : undefined`. The three claimed smells map 1:1 to the three extractions (cognitive complexity in `clusters.map`, nested ternary in `aria-label`, nested ternary in the sidebar bullet). Both **dark** screenshots were regenerated in that commit (2–4 byte PNG deltas), so `dashboard/verify/5184/` matches the refactored code, not the pre-refactor code.

---

### Test evidence (run by me, at `e85eae92`, tree clean before and after)

- **`pnpm test` → 264 files / 2834 tests passed, 0 failed.** Matches the PR body's figure exactly.
- **Revert spot-check — the "11 tests" claim is true, and I did not take it on faith.** `git checkout dab8a70b -- ` the seven consumer files (`mapGraph.ts`, `exportGraph.ts`, `types.ts`, `TopologyPage.tsx`, `TopologySidebar.tsx`, `TopologyGraph.tsx`, `TeamDetailPanel.tsx`), then re-ran the topology + page suites: **exactly 11 failed / 229 passed across 5 failing files**, each on the specific claim it protects — mapper (`team_id → team`), blank-id mapper, JSON export `team: null`, sidebar unclaimed stat, sidebar filter-row label, cluster label, header count (×2, incl. the all-unclaimed → `0 teams` case), page filter-row label, page unclaimed stat, unclaimed detail panel. Files restored, `git status --porcelain` empty.

### Regression assessment

Diffed **every** changed test against the merge-base. **Nothing weakened.**

- `TopologyGraph.test.tsx` (+28/−15) — the two AAASM-5140 tests asserting the cluster was *inert* were replaced with strictly stronger assertions (named label, `role=button`, `onTeamClick` fires with the sentinel). That's a deliberate, documented supersession: 5140's stopgap existed **only** because the group was keyed `''` and `TopologyPage` read that as falsy, so the click opened nothing. With a real key the affordance has a successful path, so restoring it is correct.
- `TopologySidebar.test.tsx` — `STATS` literal gained the now-required `unclaimed: 0`; no assertion touched.
- `mapGraph.test.ts` — **tightened**: `team: ''` → `team: UNCLAIMED_TEAM`, plus a new `expect(...).not.toBe('')` and a new blank-`team_id` case.
- `exportGraph.test.ts`, `unclaimed.test.ts`, `TopologyPage.test.tsx` — additions only.
- No deletions, no `skip`/`todo`, no loosened matchers, no threshold or config changes. No Rust, no OpenAPI, no generated schema touched (codegen drift check green).

---

### Playwright / visual evidence (my own lane — port **4623**, repo-local `@playwright/test`, not the MCP server)

**Stale-`dist` hazard handled explicitly:** `rm -rf dashboard/dist && pnpm build` → **exit 0** (`tsc --noEmit && vite build`), `dist/index.html` confirmed present before serving, own config with `reuseExistingServer: false`. Nothing was read off a stale preview.

**14/14 passed, light and dark.** Verified:

- header `· 2 agents · 1 team` (was `2 teams`); filter row `⚠ Unclaimed`; `⚠ 1 unclaimed` stat; cluster label `⚠ Unclaimed`; real `support` cluster still drawn under its own name.
- **`__orphan__` never appears anywhere in the rendered `body` innerText** — asserted against the whole document, not just the elements under test.
- **Keyboard + semantics:** cluster has `role="button"`, `tabindex="0"`, `aria-label` matching `/no team/i`; `.focus()` + `Enter` opens the panel with `data-unclaimed="true"`.
- **Genuine zero is a real answer:** an all-unclaimed fleet reads `2 agents · 0 teams` (not `1 team`) with `2 unclaimed`, and the cluster is still drawn. `[]` survives as an answer — `realTeams([UNCLAIMED_TEAM]) === []`.
- **No fabricated success or zero state:** an all-teams fleet renders no unclaimed stat, no unclaimed cluster and no unclaimed filter row; an empty fleet (`nodes: []`) and a **500 on `/api/v1/topology`** both render no unclaimed affordance at all — no invented group, no invented count.
- Filtered-to-unclaimed keeps the fleet-wide header (`2 agents · 1 team`) — pre-existing semantics, the header is fleet-wide by design — and still never shows the sentinel.
- Console errors `[]` and zero `pageerror` in every case, both themes.

**ADR-0027 (Accepted) — WCAG 2.1 AA measured in-page** (relative-luminance formula, foreground vs first opaque ancestor background):

| element | light | dark |
|---|---|---|
| `topology-stat-unclaimed` (11px/400) | `#8a5a00` on `#ffffff` — **5.93:1** | `#e6b86b` on `#1c1a16` — **9.46:1** |
| unclaimed filter row (10px/400) | **5.93:1** | **9.46:1** |
| cluster label (11px/700) | **5.93:1** | **9.46:1** |
| panel warn note (11.2px/400) | `#8a5a00` on `#f5e6c4` — **4.80:1** | `#e6b86b` on `#382810` — **7.73:1** |

All ≥ 4.5:1 → **AA normal-text floor met in both themes**, none relying on the large-text exception. Committed evidence (`dashboard/verify/5184/`, 6 PNGs, light+dark × topology / panel / all-unclaimed) is consistent with the code I ran.

*Not covered, and I don't think it needs to be:* loading / unavailable / unknown / unconfigured are unchanged by this diff — the stat is gated on `> 0` and the group only exists when a node carries no team. I verified the two states this diff **could** have broken (empty, error) fabricate nothing.

---

### Business-logic assessment

The central judgement call is right, and it's the one most likely to have been got wrong: **a team-less agent is treated as a real, known state, not an absence.** `unclaimed.ts` reaches for neither `NO_DATA` nor `unknown` / `unavailable`, and the module doc states why — "we could not determine the team" is a *weaker and different* claim than "no team claims it", and the second is what's actually known. Correct. The group is therefore named and rendered like any other grouping, and only the **team count** stops including it. That's the minimal true change.

Consistency sweep across the five surfaces the ticket named: header count ✅, filter row ✅, cluster label ✅, cluster detail panel ✅, JSON export `"team": null` ✅ (with a good rationale — exporting the sentinel would hand a consumer a team id no API asserted). One surface was missed — MINOR-3 (SVG) — and one is arguably inconsistent — MINOR-4 (budget bar).

Dropping the cross-team-edge count for this group is right and well-argued: `is_cross_team` (`topology.rs:259`) requires a team on **both** endpoints, so an edge touching a team-less agent is not a boundary crossing. Reporting a number there would contradict both the gateway and the sidebar's fleet-wide counter.

**Performance — no O(n²).** The grouping is `[...new Set(allNodes.map(n => n.team))]` (O(n)), `countUnclaimed` is one `filter` (O(n)), `realTeams` is O(teams), `teamLayout` builds one `Map` in a single pass (O(n)). The only super-linear work on this page (`computeHierarchy`, `TeamDetailPanel.crossTeamCount`) is pre-existing and untouched. No regression on large graphs.

### Security assessment

**No authorization or cross-org implication.** No new endpoint, field, query parameter, or authorization branch. The grouping is computed entirely from the node set `/api/v1/topology` already returned, which is filtered by `record_visible_to` server-side; the unclaimed group can only ever contain agents the caller could already see. No secrets or PII in the new copy; the committed screenshots are synthetic fixtures. Cache-key / tenant-isolation concerns belong to #1728, not here.

**One security-adjacent note (MINOR, display-only).** `UNCLAIMED_TEAM = '__orphan__'` shares a namespace with real team ids, and `validate_tenant_id` rejects only control characters — so a caller who registers agents under `team_id: "__orphan__"` gets a **real** team rendered as `⚠ Unclaimed`, dropped from `N teams`, and given the panel copy *"no team-scoped policy or budget applies to them"* — a wrong governance statement about a team that does have team-scoped policy. Same-tenant, display-only, no data exposure. The docstring accepts the collision but frames it as *"a collision no registry has produced"*; `team_id` is caller-supplied at registration, so it should be framed as **caller-controllable**, which changes the risk calculus even if the conclusion stays. Cheap mitigation: a sentinel that cannot be a valid tenant id (e.g. one containing a character `validate_tenant_id` rejects).

---

### Remaining findings

**MINOR-1 — Proposed ADR used as implementation authority.** Four citations of *"authoritative per ADR-0025"*; 0025 is Proposed and says so. Re-anchor on `design/README.md` and note 0025's status.

**MINOR-2 — design-fidelity divergences presented as compliance, with no ADR-0017 note.** The PR says the cluster treatment (`:190`), panel (`:664-686`) and stat (`:952`) "all ship". Three of four deviate in tone or copy:

| element | `design/v2/hi-fi/topology.jsx` | shipped |
|---|---|---|
| cluster outline + label | `#b8291e` (danger red), dash `3 3` | `var(--warn)` amber, dash `3 3` |
| filter row (`:790`) | `var(--danger)` | `var(--warn)` |
| panel note (`:686`) | `var(--danger-bg)` / `var(--danger)` | `--warn-bg` / `--warn` |
| panel title (`:673`) | `⚠ team: __orphan__` | eyebrow `no team` + `⚠ Unclaimed` |
| header stat (`:952`) | `var(--warn)` | `var(--warn)` ✅ |

I think warn-over-danger is the **better** choice ("not a fault, but a governance gap" — the PR's own words) and the panel title is a clear improvement over a spec that would re-assert `team:` for a group that is not a team. But ADR-0017 (Accepted) exists precisely to record "shipped implementation authoritative over the mock" with a supersession note on the mock. Use it, or match the spec — don't ship a divergence described as parity.

**MINOR-3 — the SVG export still leaks the sentinel.** `serializeGraphSvg` clones the live DOM, which carries `data-team="__orphan__"` from `TeamCluster`. The PR fixed exactly this class of leak for JSON, and its own rationale — *"would hand a downstream consumer a team id that no API ever asserted"* — applies verbatim. Visible SVG text is fine (`⚠ Unclaimed`); it's the machine-readable attribute.

**MINOR-4 — the unclaimed cluster still draws a team-shaped budget bar.** `TeamBudgetBar team="Unclaimed" spent limit` renders in the same form as a real team's, while the panel one click away says *"no team-scoped policy or budget applies to them."* The aggregate is a sum of per-**agent** limits (`TopologyGraph.tsx:373-396`), identical in kind for real teams, so it isn't literally false — but it is the one remaining place the group is presented exactly like a team, and the consistency sweep didn't cover it. `TeamBudgetBar` is shared with the Costs page, so a bar labelled "Unclaimed" now appears in team-bar company.

**MINOR-5 — no follow-up ticket for the excluded bulk action** (see Jira section).

**NIT-1 — filter-row order is decided by the sentinel's spelling, not the label.** Measured: `["All teams", "Unclaimed", "ops", "support", "zeta"]` — `__orphan__` happens to sort first under ICU collation. Rename the sentinel and the row moves; and by label, "Unclaimed" would sort last, after "zeta". Pin it explicitly (first or last) rather than relying on the key.

**NIT-2 — `teamKeyOf` does not purely defer to `isOrphanAgent`.** `teamId && !isOrphanAgent(node)` layers a second absence test on top of the one the docstring says is the sole authority. It happens to be protected (the `''` case in the agreement test would fail if the two diverged), so this is a documentation/intent mismatch rather than a bug — either drop the `teamId &&` or say what it guards.

**NIT-3 — dead work.** `TeamDetailPanel`'s `crossTeamCount` `useMemo` still runs for the unclaimed group even though the section is no longer rendered — O(n+e) per selection.

**NIT-4 — one new test isn't load-bearing against the pre-fix code.** `TopologyGraph.test.tsx` → "makes the unclaimed group selectable now that its panel opens" still passes under the base-revision revert, because it feeds `team: UNCLAIMED_TEAM` into the view model directly while the old gate was `c.team !== ''`. The e2e covers the real chain from the wire shape, so this is a coverage-shape remark, not a gap.

**PRE-EXISTING-OUT-OF-SCOPE (no action asked of this PR):** the fleet-wide header count not narrowing under a team filter; `computeHierarchy` / `crossTeamCount` complexity; the error-state rendering on `/topology`.

---

### Cross-PR notes — relationship to #1728 (AAASM-5181 / 5182)

- **No conflict, and neither blocks the other.** #1728 touches one file, `aa-api/src/routes/topology.rs`; #1726 touches no Rust and no OpenAPI. Zero file overlap.
- **Not duplicate work.** #1728 fixes `get_overview`'s `team_map` + `standalone_root_agents` and `get_stats`'s `team_sizes` (plus injective, length-prefixed cache keys for AAASM-5181). None of those feed the Topology page, which reads only `/topology`'s graph. The PR's non-duplication claim holds — I verified it against both diffs rather than accepting it.
- **But the two lanes now disagree about the thing they were both fixing** — this is the cross-PR risk that did materialise, in a narrower form than "two layers deciding what this page shows":
  1. **MAJOR-1** — #1728's `team_of` normalises a **whitespace-only** `team_id` to no-team (`.trim().is_empty()`); this PR's `teamKeyOf` does not. Same concept, two rules, opposite answers for `"   "`.
  2. **MAJOR-2** — `/topology/stats.orphan_count` (unchanged by either PR) counts no-team-**and-`depth > 0`**; this PR's sidebar counts no-team at any depth. Two numbers, one word, no reconciliation.
- **#1728 does not apply `team_of` to the graph projection** either, so `/api/v1/topology` can still emit `team_id: "   "` even after it merges. That's what makes MAJOR-1 live rather than theoretical, and it's worth a comment on #1728 as well — the two PRs should agree on the definition before either merges.
- Merge order is otherwise unconstrained.

---

### Unresolved risks

1. `"   "` → phantom team on this page, diverging from the backend post-#1728 (MAJOR-1).
2. "Unclaimed" means two different populations FE vs BE, with no canonical record (MAJOR-2).
3. `__orphan__` is a caller-registrable team id, so the "Unclaimed" treatment is spoofable within a tenant (display-only).
4. The `__orphan__` sentinel still reaches a downloadable artifact via the SVG export.
5. A specified design item ("Suspend all unclaimed") is unimplemented with nothing tracking it.
6. Design-tone and panel-copy divergences are unrecorded, so a future audit will re-litigate them.

### Follow-up tickets to file

1. **Trim whitespace in the shared "no team" definition** — dashboard; fix in the shared predicate, add `'   '` / `'\t'` cases. *Recommend folding into this PR, not deferring: it is the same defect this PR closes.*
2. **Reconcile "unclaimed" across the dashboard, `/topology/stats.orphan_count` and `/topology/overview.standalone_root_agents`; record the canonical definition in an ADR.**
3. **"Suspend all unclaimed" bulk action** (`design/v2/hi-fi/topology.jsx:765`) — the deferred half of this ticket's product call.
4. **Unwind `__orphan__` out of the SVG export.**
5. **ADR-0017 supersession note** for the warn-vs-danger tone and the panel-copy deviation.
6. **Decide the unclaimed cluster's budget-bar treatment** (drop it, or label it as a per-agent sum).

Also worth doing on the ticket itself: move AAASM-5184 off **To Do**, and set a Fix version.

---

*Reviewed at `e85eae9214690313bccc50019fc375240442b920`. All commands run read-only against a clean worktree; the two temporary artifacts I created (a whitespace probe spec and a port-4623 Playwright config/spec) were deleted and `git status --porcelain` was empty afterwards. No approval, no merge.*

Chisanan232 (commented):
## Follow-up review — PR #1726, now at `433891a4b460c3142bbe7846965abffda57f9964`

Continuing the adversarial review of `e85eae92`. Still a **COMMENT** review — no approval, no merge. Ten commits added on top; the base of the comparison is unchanged (`dab8a70b`).

**MAJOR-1 is fixed. The SVG leak is fixed. The coverage gap is closed. MAJOR-2 remains open by design — it is a decision, not a defect, and it has been escalated rather than settled in this PR.**

---

### MAJOR-1 — resolved, and resolved in the right place

The finding was that `certain()` treats only the exact empty string as missing, so `teamKeyOf({team_id: '   '})` returned `'   '` — reinstating all three false statements this PR exists to remove, and putting the dashboard in direct contradiction with **#1728's `team_of`**, which trims. That was not a hypothetical: I ran it, and the assertion failed with `expected '   ' to be '__orphan__'`.

The fix landed at **the shared definition** — `features/teams/orphans.ts`, one line — rather than locally in topology:

```ts
isAbsent(certain(node.team_id?.trim(), 'unconfigured'))
```

This is the right place, and it matters that it is: the single-definition property is what this PR is built on, so patching around it in `unclaimed.ts` would have left `isOrphanAgent` and `teamKeyOf` disagreeing about the same input — a second fork of the concept created by the fix for the first one. Fixing the predicate keeps `teamKeyOf`'s deferral honest.

Three details I checked rather than assumed:

- **`lib/truthfulness` is untouched.** Correct call. `'   '` *is* a real string, and a general-purpose absence primitive should say so; "a blank team id is not a team" is a domain rule about one field. Widening `certain` would have silently changed every other `certain(…)` call site in the dashboard.
- **A real id is returned unchanged, not trimmed** — so ` support ` still keys on exactly what was registered, which is the same choice `team_of` makes ("The id is returned unchanged when it is not blank"). The two layers now agree on both halves of the rule, not just the empty half. There is a test pinning this (`keeps a padded real team id exactly as registered`).
- **Blast radius enumerated and bounded.** `isOrphanAgent` has exactly two consumers: `selectOrphanAgents` → `TeamsPage` (the unclaimed list + census reconciliation), and `teamKeyOf` → `mapGraph`. I checked every other open PR for overlap — #1730 touches `features/teams/api.ts` but neither `orphans.ts` nor `TeamsPage.tsx`, and changes no `AgentNode` field. No other lane is affected.

Worth noting the Teams-page side-effect is a **second latent bug quietly fixed**: post-#1728 the backend excludes a blank-team agent from `overview.teams`, so before this change `reconcileAgentCensus` would have counted it in neither `teams` nor `orphans` — a non-zero `unaccountedFor` and a spurious census notice. The trim closes that too.

`teamKeyOf` was also reordered so `isOrphanAgent` is the leading term; the remaining `teamId` test is type narrowing that can only be falsy once the predicate has already said so, and the comment now says exactly that instead of claiming a purity the code didn't have.

**Coverage of the fix** is at every level the defect could re-enter: `isOrphanAgent` (spaces / tab / newline, plus a padded id staying claimed), `teamKeyOf`, `selectOrphanAgents`, the mapper, the "these two agree for every input" test extended to whitespace and padded ids, and a new light+dark e2e case driving `team_id: "   "` off the wire. That last one is the one I'd have insisted on — it is the only test that proves the whole chain, and it is the shape the gateway can actually emit.

### SVG export — resolved

`serializeGraphSvg` now strips `data-team` from elements already carrying `data-unclaimed="true"`. Removing rather than blanking is the right encoding for the reason given: an empty `data-team` would be precisely the blank-team assertion the ticket exists to end. Tests cover the sentinel's absence, a real team's attribute surviving, and — the one I'd have asked for — that the **live DOM is not mutated** by exporting.

### Codecov — resolved, and the uncovered lines were real

92.85% patch (4 missing + 1 partial, all `TopologyGraph.tsx`) against Sonar's "100.0% Coverage on New Code". Both figures were true under their own definitions, which is why they shouldn't be conflated. The uncovered lines were **reachable, not dead**: `TeamCluster`'s `handleKeyDown` body and the pan-drag click suppression. I had already exercised the keyboard path in Playwright, so the gap was in unit coverage, not in behaviour.

Now covered by tests that assert behaviour rather than chase lines: Enter and Space activate the cluster, other keys do not, and a click that merely concluded a pan-drag does not select while the next clean click still does. I re-ran coverage scoped to the topology suites and confirmed **zero uncovered lines and zero uncovered branches in the diff region** (`TopologyGraph.tsx` 25-120). The remaining uncovered lines in that file (566-600) are pre-existing pan/zoom code outside this diff.

### MAJOR-2 — correctly left open

"Unclaimed" still names three different populations: this page counts no-team at any depth; `/topology/stats.orphan_count` counts no-team **and `depth > 0`** (`aa-api/src/routes/topology.rs:951-955`); `/topology/overview.standalone_root_agents` counts no-team **and `depth == 0`** (`:595-597`). The design's own panel copy (`design/v2/hi-fi/topology.jsx:686`) says the group mirrors `orphan_count`.

One thing I got sharper on re-reading: the backend's two fields **partition** the dashboard's one — they are differently sliced, not contradictory, and nothing currently reports a wrong number. But the slices are operator-visible: `aasm topology stats` renders `Orphans | <n>` (`aa-cli/src/commands/topology/render/table.rs:143`) and `aasm topology overview` lists the standalone roots (`:38-42`), so an operator comparing the CLI with this page sees two different counts for one fleet, under two words that mean the same thing to them.

This is a naming-and-reporting decision that predates this PR (AAASM-5157 set the dashboard's definition; this PR only puts a number on screen), the resolution is a backend change in another lane, and narrowing the dashboard to match `orphan_count` would **re-open AAASM-5157's bug**. It is escalated for a decision rather than settled here, and the PR body now records the divergence explicitly. I would not block on it.

### Also now recorded in the PR body

- The `__orphan__` sentinel being a **caller-registrable** `team_id` — reframed from "a collision no registry has produced" to what it is: caller-supplied input. Display-only, same-tenant, no authz or data-exposure consequence, and deliberately not defended. That is the right disposition; I flagged it so the accepted risk is accurate, not so a defence gets built.
- The remaining design-tone divergences (`--warn` where `design/v2` uses danger at `:190`, `:790`, `:686`) and that **ADR-0025, which makes `design/v2` authoritative, is itself Proposed** — so the mock's standing rests on `design/README.md`, not that ADR. ADR-0017 (Accepted) is the mechanism for recording a ratified divergence; still recommended, not blocking.

### Follow-ups filed

- **AAASM-5206** (Bug, sev-low, rc.7) — the unclaimed cluster's team-shaped budget bar contradicting the panel copy. Filed rather than fixed here: the treatment is a visual/product call and `TeamBudgetBar` is shared with the Costs page.
- **AAASM-5207** (Story, rc.7) — `Suspend all unclaimed` (`design/v2/hi-fi/topology.jsx:765`), the design item this PR deliberately excluded. It had no ticket; now it does, so the exclusion is tracked rather than dropped.

Both linked to AAASM-5184.

---

### Verification at `fc3b6f44` (run by me, worktree clean before and after)

| gate | result |
|---|---|
| `pnpm test` | **264 files / 2849 tests passed**, 0 failed (was 2834 at `e85eae92`) |
| `pnpm type-check` | exit 0 |
| `pnpm lint` | exit 0, `--max-warnings 0` |
| `pnpm generate:api` → `git status --porcelain` | no schema drift |
| `rm -rf dist && pnpm build` | **exit 0**, fresh `dist/index.html` confirmed present before serving |
| 5184 Playwright pass (own config, port 4612, fresh `dist`) | **8/8** light + dark, including the new blank-team case |

Screenshots regenerated against that build and committed — the six PNGs shifted by 1-42 bytes (re-render, not a visual change), and I read `topology-unclaimed-light.png` back to confirm it still shows `· 2 agents · 1 team`, `⚠ 1 unclaimed`, `◎ All teams / ⚠ Unclaimed / ○ support`, and two named clusters.

Gates were re-run in full after the Sonar fix at `433891a4` (test / type-check / lint / build / e2e all as above); `pnpm generate:api` was clean at `fc3b6f44` and the `.dataset` change touches no schema, with CI's codegen drift check green on the final head confirming it.

Commit granularity is right: ten commits, one logical unit each, fix and its tests separated. Pre-commit and pre-push hooks ran on every one (`cargo doc` included on push) — no `--no-verify`, no force-push, no `git stash`.

### CI on `fc3b6f44`

**All green on `433891a4`: 15 success, 23 skipped, 0 failed** (38 check-runs, paginated — the unpaginated view truncates at 30 and undercounts).

- **SonarCloud analysed SHA = `433891a4`**, comment `04:52:41Z` vs head committed `04:42:34Z` — **analysis postdates the head**. **Quality Gate passed · 0 New issues · 0 Accepted · 0 Security Hotspots · 100.0% Coverage on New Code · 0.0% Duplication.** Confirmed independently against the SonarCloud API (`/api/issues/search?pullRequest=1726&resolved=false` → `total: 0`), not just read off the comment.
- **This needed a fix, and it is worth recording why.** The intermediate head `fc3b6f44` came back with **1 New issue** — `typescript:S7761`, *"Prefer `.dataset` over `removeAttribute(…)`"*, MAJOR maintainability, at `exportGraph.ts:53` — introduced by my own SVG-sentinel fix. The Quality Gate still *passed*, so it would have been easy to leave; it was fixed rather than accepted. `delete el.dataset.team` with the query typed as `SVGElement` (`querySelectorAll` otherwise yields bare `Element`, which has no `dataset`). Same behaviour, asserted by the existing export tests — no suppression, no threshold change.
- **Codecov: `codecov/patch` on `fc3b6f44` reported `100.00% of diff hit (target 70.00%)`** — up from the **92.85%** (4 missing + 1 partial) that prompted the coverage work. `433891a4` changes one line inside a function that was already fully covered; at the time of writing Codecov has not yet re-reported on it, so the 100.00% figure above is the last one it published, on the commit that contains all the substantive change.
- `CodeQL` — "No new alerts in code changed by this pull request". `Dashboard (type-check + lint)`, `Dashboard build`, `Dashboard tests + coverage`, `Dashboard codegen drift check`, `aa-cli build compat (dashboard embed)`, `CI Success` all green.

---

### Standing verdict

The truthfulness defect is fixed, the fix is now whole (the whitespace hole was the same defect wearing a different input), the tests are load-bearing — I re-verified the "11 tests" claim at `e85eae92` by reverting the seven consumer files and watching exactly 11 go red — and every visible statement I could measure is true in both themes at AA contrast.

What remains is one escalated decision (MAJOR-2), two filed follow-ups, and a recommendation to record the design-tone divergence under ADR-0017. None of those is a reason to hold the fix.

**Reviewed head SHA: `433891a4b460c3142bbe7846965abffda57f9964`.** No approval, no merge — that call is Bryant's.

Chisanan232 (commented):
## Delta review — `e85eae92` → `433891a4` (10 commits)

**Scope:** this is a *delta* review only. The review at `e85eae9214690313bccc50019fc375240442b920` stands unchanged; everything below is `git diff e85eae92..433891a4` and nothing else. I did not author these commits. **No approval, no merge** — comment only.

**Verdict on MAJOR-1: genuinely resolved, correctly placed, and consistent with #1728.** Two new MINORs below (one of them cross-PR, owned by #1728), one attribution correction, and one qualifier on the partition claim. Nothing blocking.

---

## MAJOR-1 — whitespace-only `team_id` → verified resolved

`isOrphanAgent` is now `isAbsent(certain(node.team_id?.trim(), 'unconfigured'))`.

**`lib/truthfulness/` is byte-untouched — not just in the delta, in the whole PR.**
`git diff $(git merge-base main HEAD)..433891a4 -- dashboard/src/lib/truthfulness/` is empty; the 27-file PR filelist contains no `lib/` path. `certain` still folds exactly `null | undefined | ''` (`absence.ts:186`). The domain rule lives at the field, which is the right place for it.

**A real id is returned unchanged, not trimmed — confirmed.**
`teamKeyOf` returns the *raw* `node.team_id`, never the trimmed value; the trim only ever decides emptiness. `' acme '` → `certain('acme')` is present → `isOrphanAgent` false → the key is `' acme '`, byte-for-byte. Two tests pin this and both are new in the delta:
- `orphans.test.ts` — *"does not trim a real team id into a different team"* (`' platform '` stays claimed)
- `unclaimed.test.ts` — *"keeps a padded real team id exactly as registered"* (`teamKeyOf(' support ') === ' support '`)

**Matches #1728's `team_of` on both halves.** Read directly from `925d07f1`:
```rust
fn team_of(record: &AgentRecord) -> Option<&str> {
    record.team_id.as_deref().filter(|team| !team.trim().is_empty())
}
```
| input | `team_of` (#1728) | `isOrphanAgent`/`teamKeyOf` (this PR) |
|---|---|---|
| `None` / `null` / `undefined` | `None` | unclaimed |
| `""` | `None` (`"".trim().is_empty()`) | unclaimed |
| `"   "` / `"\t"` / `"\n"` | `None` | unclaimed |
| `" acme "` | `Some(" acme ")` — **untrimmed** | `" acme "` — **untrimmed** |

Both halves agree, *and* both preserve the registered id rather than normalising it. `team_of` is applied at all three of its backend sites (`topology.rs:636` overview `team_map`, `:666` `standalone_root_agents`, `:1025` stats `team_sizes`/`orphan_count`).

**Load-bearing beyond Topology — worth stating explicitly.** `reconcileAgentCensus` (`orphans.ts:96-101`) computes `grouped = Σ overview.teams[].agent_count + orphans.length`, mixing a **backend** figure with a **frontend** one. Post-#1728 without this trim, a `"   "`-team agent falls out of *both* sides — excluded from `teams[]` by `team_of`, excluded from `orphans` by the untrimmed predicate — so `unaccountedFor` becomes 1 and the Teams page would falsely announce it is hiding an agent from itself. The trim is what keeps that census honest. Fixing at the shared definition rather than locally in topology was the right call for this reason, not only for the one-definition property.

**Blast radius — verified independently, both claims hold.**
On this head the only production consumers of `isOrphanAgent` are the two named:
- `selectOrphanAgents` (`orphans.ts:49`) → `TeamsPage.tsx:48`
- `teamKeyOf` (`unclaimed.ts:69`) → `mapGraph.ts:84`

`reconcileAgentCensus` does **not** call it (it reads the backend's `teams[]`), and `TeamOrphanDetail.tsx` imports only the `AgentCensus` *type*. Across **all 11 currently-open PRs** (`1724, 1725, 1727, 1728, 1729, 1730, 1731, 1050, 1051, 1053`): zero touch `orphans.ts`, `unclaimed.ts`, `mapGraph.ts`, `exportGraph.ts`, `TeamsPage.tsx` or any topology component, and `git diff main...HEAD | grep -E 'isOrphanAgent|teamKeyOf|selectOrphanAgents|UNCLAIMED_TEAM|serializeGraphSvg'` is empty in every one. #1730 does touch `features/teams/api.ts`, but its whole diff there is `+12/-0` adding `monthly_spend_usd: number | null` to **`TeamListRow`** — it does not touch `AgentNode`, `orphans.ts` or `TeamsPage.tsx`. Claim confirmed.

**The `teamKeyOf` reorder cannot change behaviour.** `teamId && !isOrphanAgent(node)` → `!isOrphanAgent(node) && teamId` is a conjunction consumed only for its truthiness by a ternary; `&&` is commutative there. The only asymmetry is short-circuit evaluation: the new order always calls `isOrphanAgent`, where the old order skipped it when `teamId` was falsy. `isOrphanAgent` is pure and total on that input (`''.trim()` is fine, `?.` absorbs `null`/`undefined`), so nothing observable moves. The narrowing comment is accurate — the residual `teamId` test can only be falsy when `isOrphanAgent` already returned `true`.

---

## SVG export sentinel — verified **against the real rendered graph**, not just the unit fixture

`delete el.dataset.team` removes the attribute; it does not blank it. `querySelectorAll<SVGElement>('[data-unclaimed="true"][data-team]')` only matches the unclaimed cluster, and `data-unclaimed={unclaimed ? 'true' : undefined}` (`TopologyGraph.tsx:82`) means React omits the attribute entirely for real teams, so they cannot match.

I did not take that on the unit test's word. I drove a real export end-to-end (repo-local `@playwright/test`, port **4633**, against a `dist` rebuilt with a verified exit code — see Build/e2e below), clicked *export graph → Export SVG*, captured the download, and read it:

```
LIVE_BEFORE = ["g|support|unclaimed=undefined", "DIV|support|unclaimed=undefined",
               "g|__orphan__|unclaimed=true",   "DIV|Unclaimed|unclaimed=undefined"]
EXPORTED_data-team          = ["support", "support", "Unclaimed"]
EXPORTED_contains_sentinel  = false
LIVE_AFTER  = ["g|support", "DIV|support", "g|__orphan__", "DIV|Unclaimed"]
```

- `__orphan__` appears **nowhere** in the exported file. ✅
- Removal, not blanking — no `data-team=""` in the output. ✅
- Real teams survive (`data-team="support"` intact, twice). ✅
- **Live DOM untouched** — `g|__orphan__` is still there after the export. ✅

**The live-DOM test would catch a mutation.** `exportGraph.test.ts` holds a reference to the *live* `cluster` element, calls `serializeGraphSvg(svg)`, then asserts `cluster.getAttribute('data-team')` still equals `UNCLAIMED_TEAM`. Had the loop been written against `svg` rather than `clone`, that attribute would be gone and the assertion fails. It is a real guard, not a line-executor.

### MINOR-1 (new) — the export still carries a `data-team` for the unclaimed group

See the probe above: the exported SVG contains `data-team="Unclaimed"`. It comes from the shared `TeamBudgetBar` nested inside the cluster's `<foreignObject>` (`TopologyGraph.tsx:108` → `TeamBudgetBar.tsx:71`), which carries no `data-unclaimed`, so the `[data-unclaimed="true"][data-team]` selector cannot reach it.

This is **not** a re-assertion of the sentinel and it is not a false number — `Unclaimed` is the display label, already on screen as visible text. But two things follow:

1. The export is internally inconsistent: the group's cluster `<g>` deliberately has no `data-team`, while a descendant of that same cluster does. A consumer scraping `data-team` out of the SVG still gets the group as a team-shaped entry.
2. `expect(out).not.toContain('data-team')` in *"does not write the unclaimed sentinel into the exported file"* passes **only because the fixture is a bare `<g>` with no budget bar**. It does not hold for the real export. The other two assertions in that test (`not.toContain(UNCLAIMED_TEAM)`, `toContain('data-unclaimed="true"')`) are both genuinely true of the real export — it is specifically the `data-team` assertion that overstates its reach.

Adjacent to **AAASM-5206**, whose root-cause section already names this exact line (`<TeamBudgetBar team={label} …>`), so a fix there very likely subsumes it. Flagging so the export half is not lost: 5206 is written as a *visual* legibility bug and says nothing about the exported file. Suggest either narrowing that one assertion to the element under test, or noting the export surface on 5206.

### MINOR-2 (new, cross-PR — owned by #1728 / `aa-api`, not by this PR)

This PR's body states that once #1728 lands, the two layers agree about what "no team" means. For **grouping** that is now exactly true (table above). For **edge classification** it is not, and the gap is on the backend side:

`get_topology_graph` builds its team lookup from the **raw** field, not `team_of` (`topology.rs:1130` on `925d07f1`):
```rust
let teams_by_id: HashMap<[u8; 16], Option<String>> =
    records.iter().map(|r| (r.agent_id, r.team_id.clone())).collect();
```
`collect_graph_edges` then feeds that straight into `is_cross_team` (`:540`). So for an agent with `team_id: "   "`, `is_cross_team(Some("   "), Some("alpha"))` → `true`, even though `team_of` says that agent has no team and `is_cross_team(None, …)` is explicitly `false` (asserted by #1728's own `a_team_less_endpoint_is_never_cross_team`).

The dashboard defers to the server flag whenever present (`crossTeam.ts:34`, `if (edge.crossTeam !== undefined) return edge.crossTeam`) and `/api/v1/topology` always emits it — so the dashboard faithfully renders `⇆ N cross-team` counting an edge the very same page groups as unclaimed, and which `TeamDetailPanel` says is not a boundary crossing. Nothing to change in **this** PR; the one-line fix belongs in `aa-api` (`team_of(r)` in place of `r.team_id.clone()`), and it is small enough to fold into #1728 while that lane is open. Worth a comment on #1728 or a follow-up subtask.

---

## Sonar `typescript:S7761` — 0 new issues on this head, no behaviour change

Citing the **server**, not the bot comment. `GET /api/project_pull_requests/list`:

```
analysisDate : 2026-07-27T04:48:27+0000
commit.sha   : 433891a4b460c3142bbe7846965abffda57f9964   ← pinned to this head
qualityGate  : OK   bugs 0   vulnerabilities 0   codeSmells 0
```
`GET /api/measures/component?pullRequest=1726` → `new_violations 0`, `new_code_smells 0`, `new_bugs 0`, `new_vulnerabilities 0`, `new_security_hotspots 0`, `new_duplicated_lines_density 0.0`, `new_coverage 100.0` over `new_lines 640`.

The bot comment's `04:52:41Z` is a *notification* timestamp; the analysis itself ran at `04:48:27Z`, and the API pins it to `433891a4` by SHA — a stronger attribution than any timestamp comparison, since it does not depend on clock ordering.

`GET /api/issues/search?pullRequest=1726` returns 4 issues, **all `status=CLOSED, resolution=FIXED`**:
- `typescript:S7761` `exportGraph.ts` — created `04:34:58Z` (the `fc3b6f44` analysis), closed `04:48:27Z` (this head). Self-inflicted by the SVG fix and cleaned up in the same PR — good.
- `S3776` + 2× `S3358` (`TopologyGraph.tsx`, `TopologySidebar.tsx`) — closed `03:01:29Z`, i.e. pre-delta.

Behaviour: `delete el.dataset.team` and `removeAttribute('data-team')` both remove the attribute; the only other change is typing the query `SVGElement` so `.dataset` is reachable. The real-export probe above confirms the attribute is gone and nothing else moved. Agreed that leaving it would have been easy — gate was already green — and fixing it was the right call.

---

## Codecov — **the PR body is stale; Codecov *has* re-reported, and it agrees**

The body says Codecov had not re-reported on `433891a4`. That was true when written and is no longer. Current attribution, checked directly:

- `GET /api/v2/.../commits/433891a4…/` → `state: complete`, totals `files 333, coverage 90.63`
- `GET /api/v2/.../pulls/1726/` → `head_totals` **identical** to those (so head = `433891a4`), `updatestamp 2026-07-27T04:58:01Z` (16 min *after* the head commit), and:

```json
"patch": {"hits": 59, "misses": 0, "partials": 0, "coverage": 100.0}
```

**0 misses, 0 partials — the 92.85% → 100% claim is now independently corroborated by Codecov on this head, not only by Sonar.** (The stale-looking `04:35:13Z` PR comment update belongs to `fc3b6f44` at `04:28:04Z`; the bot updated its API record without re-editing the comment.)

The two previously-missing paths were real and reachable, and the new tests assert **behaviour**, not line execution:
- keyboard — `onTeamClick` called **with `UNCLAIMED_TEAM`** for `Enter` and `Space`; **not called** for other keys. Also asserts `tabindex="0"`, so the affordance and the handler are checked together.
- pan-drag — after a >threshold drag the click is swallowed, **and then a clean click still selects**. That second half is what makes it a real test: an implementation that always suppressed would pass the first assertion and fail this one.

---

## Regression, gates, and evidence

**Tests 2834 → 2849 — verified locally on the head:** `264 files / 2849 tests`, all passing.

**No assertion removed or relaxed.** `git diff e85eae92..433891a4 -- '*.test.*'` deletes exactly three substantive lines, and each is a *strengthening*:

| removed | replaced by | net |
|---|---|---|
| `it('never yields the empty string')` over `[null, undefined, '', 'support']`, `not.toBe('')` | `it('never yields a blank-looking key')` over `[null, undefined, '', '   ', '\t', 'support']`, **`.trim()`**`.not.toBe('')` | strictly stronger — superset of inputs, and `trim() !== ''` implies `!== ''` |
| `selectOrphanAgents(...)` → `['b','d']` | `['b','d','e']`, with `agent({id:'e', team_id:'   '})` added to the fixture | stronger |
| agreement-property loop over `[null, undefined, '', 'support', 'ops']` | `[null, undefined, '', '   ', '\t', '\n', ' support ', 'support', 'ops']` | stronger |

(The fourth deletion is the `@testing-library/react` import line gaining `fireEvent`.)

**CI — 38 check-runs, paginated.** `gh api --paginate .../check-runs?per_page=100` → 38 total (15 success, 23 skipped, **0 failing, 0 pending**). The author's count of 38 is right. Note for anyone re-checking: the *unpaginated* response also reports `total_count: 38` correctly — it is the `check_runs[]` array that truncates at 30, so read the array length, not the field. Combined commit **status** is `pending` only because there are zero legacy statuses on the SHA; every gate here is a check-run.

**Build was verified before anything was served.** `rm -rf dist && pnpm build` → **exit 0**, fresh `dist/` timestamped after the delete. No stale-preview risk. I then ran the committed 5184 suite from an out-of-tree config on port **4633** with `reuseExistingServer: false` (both 4612 and 4633 confirmed unbound first) — **8/8 pass**, matching the author's claim. Worth noting for future runs: the committed `review-5184.config.ts` pins 4612 with `reuseExistingServer: true`, which is precisely the configuration that would silently reuse a stale preview; the mitigation has to be the verified build, and here it was done.

**Screenshots match the current build — compared by content.** Running the suite regenerates all six, so I captured mine, diffed against `HEAD`, then restored (`git checkout -- dashboard/verify/5184/`; tree confirmed clean). Byte deltas 1–53; per-pixel comparison (all six `1280×736`, identical geometry):

| file | differing px | max channel Δ | rows touched |
|---|---|---|---|
| `all-unclaimed-dark` | 708 / 942080 (0.075%) | 42 | 316–417 |
| `all-unclaimed-light` | 793 / 942080 (0.084%) | 35 | 316–417 |
| `topology-unclaimed-dark` | 96 (0.010%) | 2 | 328–406 |
| `topology-unclaimed-light` | 158 (0.017%) | 4 | 328–406 |
| `unclaimed-panel-dark` | 36 (0.004%) | 2 | 328–405 |
| `unclaimed-panel-light` | 146 (0.016%) | 3 | 328–406 |

Every differing pixel falls inside the graph-canvas band; the header, sidebar and detail-panel regions are byte-identical. Visual check of the highest-delta pair confirms identical content — `· 2 agents · 0 teams`, `⚠ 2 unclaimed`, `⚠ Unclaimed` filter row, `⚠ UNCLAIMED` cluster label all unchanged; only d3-force settling positions differ. This is the known `verify/` re-render noise, not a content change.

---

## Tickets

Both exist, are linked to **AAASM-5184** (`Relates`, link ids `11017`/`11018` — reciprocal on 5184), and follow the type-specific schema.

- **AAASM-5206** — Bug, *"Unclaimed Topology cluster renders a team-shaped budget bar"*. Low / 2 pts / Component `agent-assembly` / Team Pioneer / Fix version `agent-assembly v0.0.1-rc.7` / assigned / labels `bug, dashboard, design-parity, finding, follow-up, sev-low`. Description carries Summary, Expected, Actual, Repro steps, Test environment, Impact, **Root cause** (names the exact line), AC, Regression risk. Correct Bug schema. It even flags the shared-with-Costs blast radius.
- **AAASM-5207** — Story, *"Suspend all unclaimed agents from Topology"*. Medium / 5 pts / same Component, Team, Fix version, assignee / labels `dashboard, design-parity, feature, follow-up`. Proper user voice ("As an operator … I want … so that …"), plus design rationale, approach, AC and Out-of-scope. Correct Story schema, and it correctly carries the deliberately-deferred `design/v2:765` bulk action.

**NIT** — both have `Sprint` (`customfield_10020`), `Start date` (`customfield_10015`) and `Due date` empty. Consistent with the rest of this lane (AAASM-5184 itself has no Fix version, Team or points), so this reads as a lane-wide backfill rather than anything these two got wrong.

---

## MAJOR-2 — not re-opened, and I agree with the author's analysis (with one qualifier)

Not implemented here, correctly — it is a **PRODUCT-DECISION-REQUIRED** item escalated to the engineer. I checked the load-bearing "2 + 2 = 4, differently sliced" claim rather than taking it on faith, and it **holds** for this PR:

Post-#1728 both backend fields apply the *same* `team_of` rule — `standalone_root_agents` is `depth == 0 && team_of(r).is_none()` (`:666`), `orphan_count` is `team_of(r).is_none() && depth > 0` (`:1026-1028`). Since `depth` is a non-negative integer, `{depth == 0}` and `{depth > 0}` are exhaustive and disjoint over the no-team set, so the two fields do partition the dashboard's single count. And the populations line up for the calls the dashboard actually makes: `/api/v1/topology` (`:1122-1127`) and `/topology/stats` (`:992-997`) both use `registry.list()` filtered by `record_visible_to` and nothing else, and `features/teams/api.ts:29,64` calls `/topology/overview` and `/topology` **with no query params at all**. So: differently sliced, not contradictory. I have no counter-example.

**Qualifier, so the claim is not over-generalised when the engineer decides.** The partition is a property of the dashboard's *parameterless* usage, not of the fields themselves. `/topology/overview` alone accepts `status`, `min_depth` and org scoping (`:606-627`), which `/topology/stats` and `/api/v1/topology` do not. Four team-less agents — two at `depth 0` (one `suspended`), two at `depth 1` — give an API consumer:

- dashboard-equivalent (`/api/v1/topology`): **4** unclaimed
- `/topology/stats.orphan_count`: **2**
- `/topology/overview?status=active.standalone_root_agents`: **1** (the suspended root is filtered out)

→ 2 + 1 = 3 ≠ 4. Not reachable from the dashboard today, but if the resolution of MAJOR-2 is "document the three definitions as a partition", the doc should say *under identical, unfiltered scope*, or the next consumer to pass `?status=` inherits a fresh contradiction.

---

## Findings

| # | Class | Finding |
|---|---|---|
| 1 | **MAJOR-1 — RESOLVED** | Whitespace `team_id` fixed at the shared definition; matches #1728's `team_of` on `None`/empty/whitespace **and** on preserving a padded real id verbatim; `lib/truthfulness` untouched; reorder provably behaviour-neutral; blast radius is the two claimed consumers and nothing in any open PR |
| 2 | **MINOR** | Exported SVG still carries `data-team="Unclaimed"` from the nested `TeamBudgetBar`; the unit test's `not.toContain('data-team')` holds only for its synthetic fixture. Adjacent to AAASM-5206 (which does not mention the export surface) |
| 3 | **MINOR** | `/api/v1/topology`'s `teams_by_id` uses raw `team_id`, not `team_of`, so `cross_team` still treats `"   "` as a team — the dashboard renders that flag verbatim. Cross-PR, **owned by #1728 / `aa-api`**, nothing to change here |
| 4 | **NIT** | AAASM-5206 / 5207 have no Sprint / Start date / Due date (lane-wide, not specific to these) |
| 5 | **PRODUCT-DECISION-REQUIRED** | MAJOR-2 — unchanged, correctly deferred. Partition claim verified; add "under identical, unfiltered scope" when it is written down |
| 6 | *correction, in the PR's favour* | Codecov **has** re-reported on `433891a4`: patch **100.0%, 0 misses, 0 partials**. The body's "not yet re-reported" is stale — worth updating so the coverage claim rests on Codecov's own number rather than on Sonar's differently-defined one |

Delta looks good. Not approving and not merging — that is the engineer's call, and MAJOR-2 is still open for them.

PR #1728, opened 2026-07-27
[AAASM-5181] 🐛 (aa-api): Key the topology cache by tenant and stop counting team-less agents as a team

Chisanan232:
## Description

Two defects in `aa-api/src/routes/topology.rs`, in the same file and the same
defect class: a projection that names something it does not actually
distinguish.

**AAASM-5181 (High, security) — the overview cache key omitted the org.** The
key named the caller's tenant tag but not the org the request selected. That tag
is the constant `admin` for every admin caller, so
`GET /api/v1/topology/overview?org_id=A` and `?org_id=B` collapsed onto one
entry: whichever arrived second was served the first org's overview for the rest
of the 1s TTL. The comment directly above the key asserted the opposite
guarantee. The org is now part of the key (`effective_org` — the caller's own org
for a non-admin, the requested one for an admin), which closes the read *and*
the poisoning direction together, since neither request can populate an entry the
other reads. The comment now states what the key delivers.

While fixing it I checked the four sibling caches in this file (tree, team,
lineage, stats) for the same omission. **None of them omits a response-shaping
parameter** — `get_team` accepts `?org_id` but never reads it, so it cannot shape
the body (noted inline; if it ever starts filtering by org, it belongs in that
key).

All five keys did, however, share a **second, subtler flaw**: they were
`|`-joined from caller-controlled parts. `AgentRegistry::validate_tenant_id`
(AAASM-4190) rejects control characters only, so `|` is legal inside an `org_id`
/ `team_id`, and `status` and the `{team_id}` / `{agent_id}` path segments are
free-form. An `org="acme"` + `team="x|team=y"` caller and an `org="acme|team=x"`
+ `team="y"` caller both rendered `org=acme|team=x|team=y` — one key, two
tenants, in both the read and the write direction. Components are now
length-prefixed so the key is decodable and therefore injective, and an absent
optional part no longer collides with a present empty one (`unwrap_or("")`
collapsed those, though they are different queries). This extends AAASM-4190's
own bucket-key decision to the response caches.

**AAASM-5182 — a team-less agent was counted as a team.** The registry accepts
`Some("")` for `team_id`, and both the overview breakdown and the stats aggregate
grouped on `team_id.is_some()`. One team-less agent therefore opened a team entry
of its own: `team_count` claimed a team that does not exist, and the group had no
name to render. Worse, such an agent appeared in *neither* the team breakdown nor
`standalone_root_agents`, existing only inside `total_agent_count`. Both sites now
group through `team_of`, which folds `None` and a blank id into one "no team"
answer where the grouping is built — not at the render layer. `total_agent_count`
is unchanged: it counts every visible agent either way.

The stats aggregate was fixed alongside the overview deliberately: leaving it
would have let `/topology/stats` and `/topology/overview` report different team
counts for one registry.

## Type of Change

- [x] 🐛 Bug fix
- [x] ♻️ Refactoring (the cache-key encoder)

## Breaking Changes

- [x] No

No schema change — `openapi/v1.yaml` and `api/generated/**` are untouched. The
response *values* change (a blank team no longer appears as a team; a blank-team
root now appears among the standalone roots), which is the point of the fix.

## Related Issues

- Related Jira ticket: [AAASM-5181](https://lightning-dust-mite.atlassian.net/browse/AAASM-5181) (High, security), [AAASM-5182](https://lightning-dust-mite.atlassian.net/browse/AAASM-5182)
- The frontend half of the blank-team defect is [AAASM-5184](https://lightning-dust-mite.atlassian.net/browse/AAASM-5184) and is **not** in this PR — `dashboard/**` is untouched here.

## ADR

**Action: `follow`** — no new ADR. This change makes the code deliver decisions
already on record, and does not move a boundary:

- **AAASM-3483** — the `/topology/*` tenant-scoping convention recorded in-code
  at `topology.rs:65-124`, mirroring AAASM-3139. The cache key is what failed to
  honour it.
- **AAASM-4190** (`aa-gateway/src/registry/store.rs:276-287`) — already decided
  that a tenant id must not "cause bucket-key collisions when the id is used as
  part of a composite key with a separator". It was applied to rate-limit buckets
  only; this applies the same decision to the response caches.
- **ADR 0015 §4** — decided that *legitimate absence* (explicitly: "a team-less
  agent") must be modelled as absence rather than collapsed into a value.

No conflict with any Accepted decision was found. A new ADR would be noise: all
five response caches in `aa-api` live in this one file, and the invariant is
documented on the `cache_key_of` helper.

## Testing

Nine tests in a new `overview_tests` module. **Verified against the unfixed
source**: 7 of the 9 fail without their fix. The two org-selector tests were then
re-run against a tree where *only* the `effective_org` key component is removed
and everything else is fixed — exactly those two fail, so they pin AAASM-5181
specifically rather than the key encoding.

The two that pass unfixed are deliberate and are not regression tests: one is a
property guard on the new key encoder, the other asserts a real team is still
grouped, counted and listed unchanged.

Both directions of the cache defect are covered — the org asked about second must
not read the first org's entry, and reversing the request order must not reverse
who leaks.

Gates (each run once, `-p aa-api`):

```
cargo nextest run -p aa-api      → 1040 tests run: 1040 passed (1 leaky), 0 skipped
cargo clippy -p aa-api --all-targets -- -D warnings → exit 0
cargo fmt --all --check          → exit 0
```

No dashboard suite: `dashboard/**` is untouched.

## Checklist

- [x] Documentation updated if behaviour changed (the false isolation comment is corrected; the new invariants are documented on `cache_key_of` / `opt_part` / `team_of`)

## Reported, not fixed

Out of this PR's scope; flagging rather than silently expanding it:

1. **`topology.rs:427` sets `policy_count` unconditionally** and **`:434` always
   emits a budget block with a non-optional `spend_usd`**, defaulting to `0.0` via
   `unwrap_or(0.0)` when the agent has no entry in the budget snapshot. An agent
   with no recorded spend is therefore indistinguishable from one that spent
   exactly `$0.00`. `limit_usd` next to it is correctly `Option`, with a comment
   saying `null` is used "rather than a misleading 0" — `spend_usd` does not get
   the same treatment. Not my tickets; needs its own.
2. **`cross_team` treats `Some("")` as a real team**, so an edge from a blank-team
   agent is flagged as a boundary crossing, contradicting `is_cross_team`'s own
   documented rule. I did **not** fix it here: `edges.rs::compute_cross_team` has
   the identical defect, and normalizing only this side would create exactly the
   disagreement between the graph and the edge list that the doc comment says must
   never happen. It needs one ticket covering both call sites.
3. **`get_team` accepts `?org_id` and ignores it** — the parameter is advertised
   in its OpenAPI `params(TopologyFilterParams)` but never read. Harmless for the
   cache (it cannot shape the body), but it is a documented filter that silently
   does nothing.

[AAASM-5181]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aa-api/src/routes/topology.rs
+++ b/aa-api/src/routes/topology.rs
@@ -109,18 +109,63 @@ fn caller_has_no_tenant_scope(caller: &AuthenticatedCaller) -> bool {
     !caller.scopes.contains(&Scope::Admin) && caller.tenant.org_id.is_none() && caller.tenant.team_id.is_none()
 }
 
+/// Join cache-key components into a string that only an equal component list
+/// can produce.
+///
+/// Every component here is caller-controlled: the registry validates a tenant id
+/// against control characters *only* (`AgentRegistry::validate_tenant_id`,
+/// AAASM-4190), so `|` is legal inside an `org_id` / `team_id`, and `status` and
+/// the `{team_id}` / `{agent_id}` path segments are free-form request input.
+/// Joining those raw lets two *different* requests assemble one key — an
+/// `org="acme"` + `team="x|team=y"` caller and an `org="acme|team=x"` +
+/// `team="y"` caller both rendered `org=acme|team=x|team=y` — which is a
+/// cross-tenant read and equally a cross-tenant write, since either caller can
+/// be the one that populates the shared entry. Length-prefixing each component
+/// makes the encoding decodable, hence injective, extending AAASM-4190's
+/// bucket-key defence to the response caches.
+fn cache_key_of(parts: &[&str]) -> String {
+    let mut key = String::new();
+    for part in parts {
+        if !key.is_empty() {
+            key.push('|');
+        }
+        key.push_str(&part.len().to_string());
+        key.push(':');
+        key.push_str(part);
+    }
+    key
+}
+
+/// Encode an optional cache-key component so an absent value and a present
+/// empty one cannot collide.
+///
+/// The two are different queries, not two spellings of one: `?org_id=` selects
+/// the records whose `org_id` is empty, while omitting it selects every record;
+/// likewise a caller scoped to `org_id: Some("")` is confined by
+/// [`record_visible_to`] where a caller with `None` is not.
+fn opt_part(value: Option<&str>) -> String {
+    match value {
+        Some(v) => format!("={v}"),
+        None => "-".to_string(),
+    }
+}
+
 /// Cache-key fragment that makes a cached topology response specific to the
 /// caller's tenant scope, so a tenant-scoped response is never served to a
 /// caller from a different tenant.
+///
+/// The tag alone is *not* sufficient to isolate a response: it is the constant
+/// `admin` for every admin caller, so any handler whose body also varies with a
+/// request parameter must name that parameter in its key too (AAASM-5181).
 fn tenant_cache_tag(caller: &AuthenticatedCaller) -> String {
     if caller.scopes.contains(&Scope::Admin) {
-        return "admin".to_string();
+        return cache_key_of(&["admin"]);
     }
-    format!(
-        "org={}|team={}",
-        caller.tenant.org_id.as_deref().unwrap_or(""),
-        caller.tenant.team_id.as_deref().unwrap_or(""),
-    )
+    cache_key_of(&[
+        "tenant",
+        &opt_part(caller.tenant.org_id.as_deref()),
+        &opt_part(caller.tenant.team_id.as_deref()),
+    ])
 }
 
 // ---------------------------------------------------------------------------
@@ -527,13 +572,12 @@ pub async fn get_overview(
 
     // AAASM-3483 — the tenant tag scopes the cache entry to the caller's tenant
     // so a tenant-scoped response is never served to a different tenant.
-    let cache_key = format!(
-        "{}|{}|{}|{}",
-        tenant_cache_tag(&caller),
-        params.status.as_deref().unwrap_or(""),
-        params.min_depth.unwrap_or(0),
-        params.show_budget.unwrap_or(false),
-    );
+    let cache_key = cache_key_of(&[
+        &tenant_cache_tag(&caller),
+        &opt_part(params.status.as_deref()),
+        &params.min_depth.unwrap_or(0).to_string(),
+        &params.show_budget.unwrap_or(false).to_string(),
+    ]);
     if let Some(cached) = state.topology_overview_cache.get(&cache_key).await {
         return (StatusCode::OK, Json((*cached).clone()));
     }
@@ -678,14 +722,13 @@ pub async fn get_tree(
         );
     }
 
-    let cache_key = format!(
-        "{}|{}|{}|{}|{}",
-        tenant_cache_tag(&caller),
-        root_id,
-        max_depth,
-        params.status.as_deref().unwrap_or(""),
-        show_budget,
-    );
+    let cache_key = cache_key_of(&[
+        &tenant_cache_tag(&caller),
+        &root_id,
+        &max_depth.to_string(),
+        &opt_part(params.status.as_deref()),
+        &show_budget.to_string(),
+    ]);
     if let Some(cached) = state.topology_tree_cache.get(&cache_key).await {
         return Ok((StatusCode::OK, Json((*cached).clone())));
     }
@@ -741,14 +784,16 @@ pub async fn get_team(
             .with_detail("Reading a team's topology requires admin scope or membership in that team"));
     }
 
-    let cache_key = format!(
-        "{}|{}|{}|{}|{}",
-        tenant_cache_tag(&caller),
-        team_id,
-        params.status.as_deref().unwrap_or(""),
-        params.min_depth.unwrap_or(0),
-        params.show_budget.unwrap_or(false),
-    );
+    // `params.org_id` is deliberately absent: unlike `get_overview`, this handler
+    // never reads it, so it cannot shape the body and naming it would only split
+    // the entry. If it ever starts filtering by org, it belongs in this key.
+    let cache_key = cache_key_of(&[
+        &tenant_cache_tag(&caller),
+        &team_id,
+        &opt_part(params.status.as_deref()),
+        &params.min_depth.unwrap_or(0).to_string(),
+        &params.show_budget.unwrap_or(false).to_string(),
+    ]);
     if let Some(cached) = state.topology_team_cache.get(&cache_key).await {
         return Ok((StatusCode::OK, Json((*cached).clone())));
     }
@@ -826,7 +871,7 @@ pub async fn get_lineage(
         );
     }
 
-    let cache_key = format!("{}|{}", tenant_cache_tag(&caller), agent_id_str);
+    let cache_key = cache_key_of(&[&tenant_cache_tag(&caller), &agent_id_str]);
     if let Some(cached) = state.topology_lineage_cache.get(&cache_key).await {
         return Ok((StatusCode::OK, Json((*cached).clone())));
     }
@@ -914,7 +959,7 @@ pub async fn get_stats(
     // they leak every tenant's agent counts. The tenant tag scopes the cache and
     // the visibility filter below confines the aggregation to the caller's tenant
     // (a non-admin with no tenant scope aggregates over an empty set → zeros).
-    let cache_key = format!("stats|{}", tenant_cache_tag(&caller));
+    let cache_key = cache_key_of(&["stats", &tenant_cache_tag(&caller)]);
     if let Some(cached) = state.topology_stats_cache.get(&cache_key).await {
         return (StatusCode::OK, Json((*cached).clone()));
     }
--- a/aa-api/src/routes/topology.rs
+++ b/aa-api/src/routes/topology.rs
@@ -1220,7 +1220,7 @@ mod graph_tests {
 
     /// A read-scoped caller confined to `org` / `team`. `None` / `None` with no
     /// admin scope is the unscoped caller the deny-by-default guard rejects.
-    fn reader(scopes: Vec<Scope>, org_id: Option<&str>, team_id: Option<&str>) -> RequireRead {
+    pub(super) fn reader(scopes: Vec<Scope>, org_id: Option<&str>, team_id: Option<&str>) -> RequireRead {
         RequireRead(AuthenticatedCaller {
             key_id: "k".to_string(),
             scopes,
@@ -1231,13 +1231,13 @@ mod graph_tests {
         })
     }
 
-    fn admin() -> RequireRead {
+    pub(super) fn admin() -> RequireRead {
         reader(vec![Scope::Admin], None, None)
     }
 
     /// Minimal registered agent. Only the fields the graph projection reads
     /// carry meaningful values.
-    fn record(id_byte: u8, name: &str, team_id: Option<&str>) -> AgentRecord {
+    pub(super) fn record(id_byte: u8, name: &str, team_id: Option<&str>) -> AgentRecord {
         AgentRecord {
             agent_id: [id_byte; 16],
             name: name.to_string(),
@@ -1273,7 +1273,7 @@ mod graph_tests {
         }
     }
 
-    fn state_with(records: Vec<AgentRecord>) -> AppState {
+    pub(super) fn state_with(records: Vec<AgentRecord>) -> AppState {
         let state = AppState::local_in_memory().expect("state builds");
         for r in records {
             state.agent_registry.register(r).expect("register");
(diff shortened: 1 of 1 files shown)


Chisanan232 (commented):
## Adversarial review — PR #1728

**Reviewed head SHA: `925d07f146a5f91ae460b643fb1a96f37a6bb15e`** (branch `v0.0.1/AAASM-5181/fix/topology_tenant`, 1 file, +333/−42, merge-base `dab8a70b` = current `main` tip).
Reviewer did not author this PR. **COMMENT only — not an approval, not a merge.**

### Verdict

**No BLOCKER and no MAJOR findings.** The security claim is real, the fix is correct, and I could **not** construct a key collision or a cross-org read against the patched code. Every load-bearing claim in the PR body that I could test independently held up — including both revert experiments, which reproduced the stated pass/fail counts *exactly*. The findings below are all MINOR/NIT/PRE-EXISTING, and the two that matter are **citation accuracy in the ADR section** and a **Jira scope mismatch on AAASM-5182**, neither of which requires a code change.

This is the strongest PR of the wave on evidence discipline: the author disclosed the hook-bypass on the first commit, disclosed three reported-not-fixed defects, filed tickets for all three, and correctly declined to half-fix the `cross_team` defect. I verified each of those disclosures rather than taking them.

---

### Security and tenant-boundary assessment (the core)

**I enumerated all five caches in `topology.rs` and derived each key by hand rather than accepting the PR's "the other four don't omit a response-shaping parameter" claim.** Result: the claim is true, and here is the working.

| Handler | Query type | Key components | Every body-shaping input present? |
|---|---|---|---|
| `get_overview` :593 | `TopologyFilterParams` (status, min_depth, show_budget, **org_id**) | tag, `effective_org`, status, min_depth, show_budget | ✅ complete |
| `get_tree` :750 | `TreeParams` (depth, status, show_budget — **no org_id**) | tag, root_id, `max_depth`, status, show_budget | ✅ complete |
| `get_team` :815 | `TopologyFilterParams` (**has org_id**) | tag, team_id, status, min_depth, show_budget | ✅ — `org_id` genuinely unread (verified :826–852) |
| `get_lineage` :899 | **no Query extractor at all** | tag, agent_id | ✅ complete |
| `get_stats` :987 | **no Query extractor at all** | "stats", tag | ✅ complete |

The tag itself is sound because `record_visible_to` (:99) short-circuits `true` for **any** admin — so an admin's own `tenant.org_id`/`team_id` provably cannot shape the body, which is what licenses the constant `"admin"` tag. For non-admins the tag carries both org and team, which is exactly the pair `record_visible_to` branches on. `get_tree` keying on the **capped** `max_depth` rather than raw `params.depth` is correct, not a gap: `?depth=50` and `?depth=10` produce byte-identical bodies.

**Injectivity — I tried to break it and failed.** I reimplemented `cache_key_of`/`opt_part` faithfully (using **byte** length, matching Rust `str::len()`) and brute-forced:

- `cache_key_of`: **41,371 component-tuples, 0 collisions** — alphabet chosen adversarially (`""`, `"|"`, `":"`, `"0"`, `"1"`, `"="`, `"-"`, `"|1"`, `"1:"`, `":a"`, `"a|b"`, `"10:x"`), arity 0–4.
- `opt_part`: 0 collisions; `None`→`"-"` vs `Some("")`→`"="` are distinct, and `Some("-")`→`"=-"` does not alias `None`.
- Full `get_overview` key shape, admin and non-admin, over 17 tenant ids including `"acme|team=x"`, `"x|team=y"`, `"5:admin"`, `"6:tenant"`, `"="`, `"-"`, `"1:a"`, `"0:"`, a 300-char id, U+00E9 and a 4-byte emoji: **9,180 distinct keys, 0 collisions between distinct response bodies.**
- Cross-handler: overview∩team, overview∩lineage, overview∩stats, team∩lineage all **∅**.

The encoding is injective for the right reason — it is *decodable*: decimal length up to the first `:` (digits can never contain `:`), then exactly N bytes, then `|` or end. Nesting `tenant_cache_tag`'s own `cache_key_of` output as a single component of the outer call is safe because the outer call length-prefixes it. Empty components, `None` vs `Some("")`, unicode and very long ids all behave.

One methodological note in the PR's favour: my first harness reported ~10,900 "collisions" — all of them were **admins differing only in their own tenant fields**, which correctly share an entry because their bodies are identical. Once identity was defined by what actually shapes the body, the count went to zero. Sharing there is a feature, not a leak.

**Authorization is at the right boundary.** The cache is *not* what stands between a non-admin and another org, and I confirmed this on every path: `caller_has_no_tenant_scope` fails closed first (:573 empty overview, :724 / :893 404-not-403 so existence can't be probed); `effective_org` is **forced** to the caller's own org for non-admins (:581–585) so `?org_id` is inert for them; `record_visible_to` then filters every record; `build_tree` re-checks on every recursively-discovered node (:238, AAASM-4819) and `get_lineage` breaks the ancestor walk at the first out-of-tenant node (:932); `get_team` gates on `can_access_team` **before** the cache read (:807). Authz precedes caching everywhere. Removing the cache entirely would not change who can read what.

**Poisoning is closed in both directions, and the test proves it.** See test evidence — the order-reversed test fails independently, so it is not redundant with its sibling.

**Lifecycle:** moka TTLs are overview 1s, tree/team/lineage 5s, stats 10s (`state.rs:372–386`). A scope change on an API key produces a *different key*, so a stale entry is not served to the changed principal — revocation is safe by construction, not by TTL. A registry mutation is stale for at most the TTL, unchanged by this PR.

---

### Business-logic assessment (AAASM-5182)

`team_of` (:66) is applied at **all three aggregation sites** — `team_map` (:636), `standalone_root_agents` (:666), and stats `team_sizes` (:1025). I grepped every `team_id` read in production code to confirm none was missed. The remaining raw reads are genuine passthroughs (`:273`, `:940`, `:950` echo what was registered — correct for a wire echo) plus the two deliberate deferrals below. Fixing stats alongside overview was the right call: leaving it would have made `/topology/stats` and `/topology/overview` disagree on `team_count` for one registry.

`total_agent_count` is correctly untouched — it counts every visible agent either way — and the blank-team root now appears in `standalone_root_agents` instead of vanishing from both the breakdown and the standalone list, which was the more serious half of 5182.

---

### Jira scope coverage

- **AAASM-5181** (Bug, High, `security`+`truthfulness-contract`, component `agent-assembly`, **no fix version**) — fully addressed. Worth recording: the *ticket* asserts "An Admin caller has an empty tenant tag", which is **false** at base — `tenant_cache_tag` returned the literal `"admin"`. The defect is real and the PR's own comment states the mechanism correctly ("the constant `admin` for every admin caller"). The PR is more accurate than the ticket it closes.
- **AAASM-5182** — ⚠️ **MINOR — scope mismatch.** Every file 5182's description names is dashboard code (`mapGraph.ts:79`, `TopologyPage.tsx:43-44/131`, `TopologyGraph.tsx:484`); this PR touches **zero** dashboard files. The backend grouping defect this PR actually fixed is real and worth fixing, but it is not what 5182 describes, and #1726 (AAASM-5184, "Frontend-only") is the PR that satisfies 5182's stated ACs. **5182 should not be closed on this PR** — either re-scope 5182 to the backend half or file the backend defect properly and let 5184/#1726 close 5182.
- **AAASM-5200 / 5201 / 5202** — all three **exist, are open (To Do), Medium, component `agent-assembly`**, created 2026-07-27, and each description genuinely matches the scope the PR body attributes to it. I re-verified 5201 and 5202 against HEAD independently (see below). NIT: all three carry **no fix version and no labels**.

---

### ADR assessment (with statuses)

PR declares action **`follow`** — no new ADR. **I accept `follow`**, but two of the three citations overstate their source.

| Cited | Status | Assessment |
|---|---|---|
| **ADR-0015 §4** (DLP trust boundary) | **Accepted** | ⚠️ **Partly supports the claim.** §4 *is* Accepted and *does* name "**a team-less agent**" verbatim. But §4 governs the **`PolicyContext` layer in the policy engine**, and the distinction it mandates is **legitimate absence vs. *resolution failure*** — i.e. don't collapse an *error* into `None`. That is the opposite direction from "don't collapse absence into a value like `""`". Applying it to the aa-api topology projection is a **reasonable analogy, not the ADR's ruling.** §4 does not govern this surface. |
| **AAASM-4190** (in-code, `store.rs:276-287`) | Ticket **Done** (rc.4) | ⚠️ **Verbatim-accurate but truncated in a way that broadens it.** The full sentence is "***Control characters* in the id could cause bucket-key collisions when the id is used as part of a composite key with a separator**". The recorded decision is a **control-character charset rule**; `validate_tenant_id` enforces exactly `char::is_control()` and nothing else. The PR drops the subject, turning a narrow charset rule into a general anti-collision principle. The *engineering* is right — `\|` really is legal, and I confirmed there is **no length limit, empty string is accepted, and ordinary spaces pass** — but "already decided" is doing more work than the source supports. |
| **AAASM-3483** (in-code convention, `topology.rs:79-89`) | Ticket **Done**, Highest | ✅ **Accurate.** The convention is exactly as cited, and 3483's own Actual section already named `get_team` as affected. |

**Uncited but relevant: ADR-0021** (`topology-enforcement-mode-mutation-safety`) — status **`Proposed — requires architecture + security sign-off before any implementation`**. It is the *only* ADR that discusses these caches (`":156 Topology responses are additionally cached (moka, 1–10 s TTLs"`, and `":332 cache invalidation … falls out of the same fix"`). Being **Proposed, it is evidence only and carries no implementation authority** — so it does not block this PR, and nothing here conflicts with it. It should still be cited, and 0021 will need updating once it moves, since this PR changes the key construction it describes. Also uncited: **ADR-0011 (Accepted)**, the one ADR that already defines a composite tenant key with a separator (`assembly.audit.<tenant>.<agent>`) — direct precedent.

No conflict with any **Accepted** decision was found. Introducing a shared key encoder across five caches is a small enough, well-localized decision that `follow` is defensible; the invariant is documented on `cache_key_of`. **Ask: correct the two citations in the PR body before merge** (no code change needed).

---

### Test evidence

Nine new tests in `overview_tests`. **I spot-checked both revert claims by actually reverting** (via a scripted rewrite + `git checkout -- <path>` restore; no `git stash`, worktree confirmed clean after each):

- **Revert A — only the `effective_org` key component removed, everything else fixed:** → **exactly 2 fail**, and they are exactly the two org-selector tests. Claim verified precisely. The failure output is a live demonstration of the vulnerability:
  `assertion left == right failed: the ?org_id=globex request must not be served acme's cached overview / left: ["team-acme"] right: ["team-globex"]`
- **Revert B — unfixed source + the new tests:** → **9 run, 2 passed, 7 failed.** The 2 that pass are **exactly** the two the PR named: `cache_key_components_cannot_be_reshuffled_into_the_same_key` (encoder property guard) and `a_real_team_is_unaffected`. Claim verified precisely.

Both poisoning directions are genuinely load-bearing: `the_org_that_populates_the_entry_does_not_decide_what_the_other_reads` fails **independently** in revert A, so it is not a duplicate of its sibling — it really does pin the write direction.

**No existing test was weakened.** I enumerated all 42 deleted lines: 6 = old `tenant_cache_tag` body, 2 = the old (false) isolation comment, 28 = the five old `format!` key blocks, 4 = the grouping lines, 4 = test-helper `fn` signatures widened to `pub(super)` with **bodies unchanged**. **Zero test assertions deleted or altered.** No fabricated success/zero/healthy state is introduced anywhere.

---

### Regression assessment

Run once each in a worktree at the exact head SHA:

```
cargo clippy -p aa-api --all-targets -- -D warnings   → exit 0
cargo nextest run -p aa-api                           → 1040 tests run: 1040 passed (1 leaky), 0 skipped, exit 0
```

Both match the PR body's stated numbers exactly. Additionally, **every one of the 4 commits is independently `rustfmt`-clean** (checked per-commit against the repo `rustfmt.toml`, including the base for control) — consistent with the author's statement that the `-c core.hooksPath=/dev/null` first attempt was soft-reset and the final history ran the hooks. I found no evidence of a bypassed commit surviving into the final history.

**No OpenAPI / generated-client drift and no wire-contract change.** One file changed; `openapi/v1.yaml` and `api/generated/**` untouched; CI's `OpenAPI spec drift check` and `OpenAPI spec lint (Spectral)` both green. The author correctly used inline `//` rather than `///` for the new rationale **inside** `#[utoipa::path]` handlers — `///` there would have folded into the OpenAPI description and drifted the spec. Response *values* change (a blank team is no longer a team), which is the intended fix, not a contract change.

---

### CI / Sonar / Codecov evidence — including the label-gating caveat

All check-runs below are on the reviewed head SHA `925d07f1`; **0 failures, 0 runs on any other SHA.**

⚠️ **Caveat worth recording, because the raw API is misleading here.** `Coverage` and `SonarCloud analysis` each appear **twice on this same SHA** — once `skipped` and once `success`:

| Check | conclusion | started → completed |
|---|---|---|
| Coverage | **skipped** | 2026-07-27T02:40:23Z |
| Coverage | **success** | 02:56:30Z → **03:19:47Z** (23 min) |
| SonarCloud analysis | **skipped** | 02:40:23Z |
| SonarCloud analysis | **success** | 03:20:03Z → **03:26:32Z** |

Both are gated on the `run-coverage` label (`ci.yml:378` coverage, `:1229` sonar), and neither has `types: [labeled]`, so the author added the label and closed/reopened the PR to force re-evaluation. **I verified the surviving analysis is real, not a phantom skip:**

- The successful `Coverage` job **ran for 23 minutes** — a skipped job completes instantly.
- `sonar` declares `needs: [changes, coverage, dashboard-test]` and started at **03:20:03Z, 16 seconds after Coverage finished at 03:19:47Z** — it consumed real LCOV, not a fallback.
- `codecov/patch` reported **success at 03:20:22Z**, immediately downstream of the real coverage upload.
- The SonarCloud PR comment (Quality Gate passed, 0 New issues, 100.0% Coverage on New Code) was updated **03:26:29Z**, matching the successful job and postdating the head commit (**02:33:07Z**).

So the "0 issues / 100% coverage" result is **genuine and covers the current head**. Anyone reading the rollup should not conclude from the `skipped` rows that Sonar never ran — and conversely, a reviewer on a PR where the label was *never* added would see an identical-looking `0 issues` via the API with no analysis behind it. The label label→close/reopen dance is a repo-wide CI ergonomics gap (adding `types: [labeled]` to `ci.yml` would fix it), not a defect of this PR.

The label is currently present on the PR. There are **no reviews on the PR** and `mergeStateStatus` is `BLOCKED` pending the required approval.

---

### Findings

**MINOR**
1. **ADR-0015 §4 citation is an analogy, not the ADR's ruling.** §4 (Accepted) governs `PolicyContext` **absence-vs-resolution-failure** in the policy engine; it does not decide wire representation of blank ids in aa-api. It *does* name "a team-less agent" verbatim, so the citation isn't fabricated — just over-extended. Reword in the PR body.
2. **AAASM-4190 quotation truncated in a broadening way.** Source scopes the rule to **control characters**; the PR renders it as a general separator/collision decision and rests part of its `follow` justification on that. Restore the subject of the sentence. (The underlying engineering finding — `\|` is legal, no length limit, `Some("")` accepted — is correct and I confirmed it in `validate_tenant_id`.)
3. **AAASM-5182 scope mismatch — do not close 5182 on this PR.** 5182's description names only dashboard files; this PR changes none. The backend defect fixed here is real but undescribed by any ticket; #1726 (5184, "Frontend-only") is what satisfies 5182 as written.
4. **Same defect class left in `build_policy_chain` (:334, :337), and it is in *none* of 5200/5201/5202.** `if let Some(team_id) = lineage.team_id` pushes `PolicyScope::Team("")` for a blank id, and `:334` does the same for `org_id` — directly contradicting that function's own doc at :325 ("*A tier appears only when the agent has that selector*"). This is precisely the "projection names something it does not distinguish" defect the PR set out to fix, in the same file, with `team_of` already available two hundred lines up. **Needs a follow-up ticket** (or folding into 5201's "one ticket, all call sites" framing).
5. **A third surface still groups on raw `team_id`:** `/topology/graph` (:1130) feeds `is_cross_team` (:540) from unnormalized ids. The PR argues overview+stats had to move together "so the two surfaces cannot report different team counts" — that reasoning applies to the graph too. Deferral is nonetheless **correct** (see below); flagging so the inconsistency is a known, time-boxed state rather than a surprise.

**NIT**
6. `cache_key_components_cannot_be_reshuffled_into_the_same_key` is documented as "the property the five topology cache keys rest on" but is **four hand-picked assertions**, not a property test. It happens to be the one test that survives a full revert, so it carries little regression weight. Consider `proptest`, or at minimum add the `None`-vs-`Some("")` and nested-tag cases.
7. `get_lineage` builds its key from the **raw** `agent_id_str` before `parse_agent_id`, so two hex spellings of one id occupy two entries. Harmless (the body echoes the raw string, so the bodies genuinely differ) and pre-existing — noting only because it is adjacent to the changed line.
8. Jira hygiene: 5181, 5182, 5200, 5201, 5202 all have **no fix version**; 5200/5201/5202 also have **no labels**.

**PRE-EXISTING-OUT-OF-SCOPE**
9. The five moka caches are built with a TTL but **no `max_capacity`**, and their keys embed caller-controlled, unbounded-length strings (`status`, `org_id`, `team_id`, raw path ids) — `validate_tenant_id` imposes no length limit. Unbounded key cardinality within the TTL window. Pre-existing; this PR adds one dimension (`effective_org`, admin-controlled) to the 1s overview cache, which is marginal. Worth a ticket on its own, not this one.
10. `get_overview`'s comment at :577-579 says an out-of-scope `?org_id` "yields no agents", but the code **forces** `effective_org` to the caller's own org, so a non-admin sending `?org_id=B` gets org A's data, not an empty result. Pre-existing (unchanged by this PR) and not a leak — but the comment is inaccurate and sits three lines above a comment this PR rewrote for exactly that kind of inaccuracy.

**Judgement on the three reported-not-fixed items — all three are genuinely out of scope and all three have real tickets:**
- **AAASM-5200** (`policy_count` unconditional, `spend_usd` via `unwrap_or(0.0)`): correctly deferred — the ticket itself notes the fix **is** an OpenAPI contract change needing dashboard sequencing. Out of scope for a cache-key PR.
- **AAASM-5201** (`cross_team` treating `Some("")` as a team): **leaving this unfixed is the right call, and I checked the reasoning rather than taking it.** `is_cross_team` (:318) is `matches!((Some(a), Some(b)) if a != b)`, so `Some("")` vs `Some("support")` → `true`; `edges.rs::compute_cross_team` has the identical shape. `is_cross_team`'s own doc (:314-317) says it mirrors `edges::compute_cross_team` "*so the graph and the edge list can never disagree about the same edge*". Normalizing one side alone would manufacture exactly that disagreement — **fixing one call site really is worse than fixing neither.** One ticket covering both sites is correct. Note the residual cost, which the PR does not state: `topology.rs` now carries **two** notions of "no team" (`team_of` for aggregation, raw `Some("")` for `cross_team`) until 5201 lands. That is documented and ticketed, so it is acceptable — but it should be said out loud in the PR body.
- **AAASM-5202** (`get_team` advertises `?org_id`, ignores it): correctly out of scope, and I independently confirmed it cannot shape the body, so it is an honesty gap and not a leak — which is exactly how the ticket frames it.

---

### Unresolved risks

- The `cross_team` / `edges.rs` inconsistency is **live until AAASM-5201 lands**, and until then `topology.rs` disagrees with itself about what a blank team is. Bounded, documented, ticketed — but it should be sequenced soon rather than aging.
- `build_policy_chain` (finding 4) is **not covered by any existing ticket**, so it will be silently forgotten unless one is filed.
- Nothing here protects against a *future* handler adding a body-shaping parameter and forgetting the key. `get_team`'s inline note is the right instinct but is prose, not a mechanism. A follow-up worth considering: make the key derive from the params struct so omission is a compile error rather than a review catch. Not this PR's job.

### Follow-up tickets

- **New ticket needed:** `build_policy_chain` treats `Some("")` `team_id`/`org_id` as a real selector, contradicting its own doc (`topology.rs:325`, :334, :337) — reuse `team_of`. Consider folding into **AAASM-5201** so all "blank id is a real team" call sites move at once, per 5201's own stated rationale.
- **New ticket suggested:** moka topology caches have no `max_capacity` with caller-controlled unbounded-length keys (finding 9).
- **Repo-wide, not this PR:** add `types: [labeled]` to `ci.yml` so `run-coverage` doesn't require a close/reopen to take effect.
- **Existing, confirmed correct:** AAASM-5200, AAASM-5201, AAASM-5202 — please add fix versions and labels.

### Cross-PR notes

- **#1726 (AAASM-5184, dashboard)** is the frontend counterpart. **No file overlap** — #1728 is `aa-api` only, #1726 is `dashboard/**` only — so they cannot conflict textually. Semantically they are complementary: #1726 derives clusters from `/topology/graph`, whose per-node `team_id` passthrough **this PR does not change**, so #1726's logic is unaffected by the backend change and both can land in either order. Note the transient window: after #1728 alone, `/topology/overview.team_count` excludes blank teams while `/topology/graph` still emits them, so a header count sourced from overview and a cluster list sourced from graph can differ by one until #1726 lands.
- No other open PR in the wave touches `aa-api/**`, so the "aa-api touchers must be sequenced off clean main" constraint does not bind here.

---

**Bottom line:** the security fix is correct and independently verified — I could not construct a key collision or a cross-org read against this code. No code change is required to merge on my findings; please correct the two ADR/ticket citations and file the `build_policy_chain` follow-up. Merge decision belongs to a code owner; this review is a comment only.

Chisanan232 (commented):
## Follow-up on the cross-PR finding from #1726's delta review

**Head SHA unchanged: `925d07f146a5f91ae460b643fb1a96f37a6bb15e`.** No commit, no push — see why below. Still a comment only; not an approval, not a merge.

### 1. The finding is real

Confirmed at `aa-api/src/routes/topology.rs:1129-1130`:

```rust
let teams_by_id: HashMap<[u8; 16], Option<String>> =
    records.iter().map(|r| (r.agent_id, r.team_id.clone())).collect();
```

Raw `r.team_id.clone()`, not `team_of`. That map feeds `is_cross_team` at `:540`, so `is_cross_team(Some("   "), Some("alpha"))` → `true` while `team_of` classifies that same agent as unclaimed. The described contradiction on the rendered page is genuine.

### 2. I did not fix it — it is **not** independent of AAASM-5201

This was the decisive check, and it comes out the other way:

```rust
// aa-api/src/routes/edges.rs:170-176   (compute_cross_team)
let team_map: HashMap<AgentId, Option<String>> = ids
    .into_iter()
    .map(|id| {
        let team = state.agent_registry.get(id.as_bytes()).and_then(|r| r.team_id.clone());
        (id, team)
    })
    .collect();
```

`edges.rs:173` performs the **identical raw read at the identical layer**, and `:183-186` performs the identical `(Some(a), Some(b)) => a != b` comparison. The defect is not in `is_cross_team` the function — **both surfaces are wrong in their team *lookup*, in exactly the same way.** So `teams_by_id` is not a separate bug adjacent to 5201; it **is** the `topology.rs` half of 5201.

Normalizing only `:1130` would produce, for one blank-team edge:

- `GET /topology/graph` → `cross_team: false`
- `GET /topology/edges` → `is_cross_team: true`

That is precisely the disagreement `is_cross_team`'s own contract forbids (`topology.rs:314-317`: *"Mirrors `edges::compute_cross_team` … so the graph and the edge list can never disagree about the same edge"*), and it would silently falsify the existing test's stated rationale at `:1436-1437` (*"the same rule `edges::compute_cross_team` applies, so the two surfaces agree"*) — that test asserts only on `is_cross_team` directly, so it would keep passing while its premise became untrue. A green suite would be hiding a new server-side self-contradiction.

The scope constraint on this lane is `topology.rs` only, so fixing both halves here is not available. Trading one consistent, single-surface defect for two mutually-contradicting API endpoints is strictly worse, so per the stop-rather-than-half-fix instruction: **stopping and reporting.**

Worth separating cause from effect: the contradiction #1726 surfaces is **not introduced by this PR** — #1728 touches neither `cross_team` nor the graph endpoint. It is the pre-existing backend defect becoming visible once the dashboard side stops matching it. The correct resolution is **AAASM-5201 covering both call sites**, and the real scheduling consequence is that **5201 now gates the user-visible correctness of #1726** and should be sequenced with it rather than left to age.

### 3. Full sweep — every remaining raw `team_id` read in production code (lines 1–1147)

Three sites use `team_of` (`:636`, `:666`, `:1025`). Every remaining reference, classified:

| Line | Read | Verdict |
|---|---|---|
| `:66-67` | `team_of` definition | the normalizer |
| `:104`, `:112`, `:123`, `:181` | `caller.tenant.team_id` | **Correctly raw** — caller *scope*, not a record. `record_visible_to` must match exactly what was registered; folding blank here would widen a `Some("")`-scoped caller's visibility. `:181` additionally needs raw bytes for key injectivity. |
| `:113` | `record.team_id != Some(team)` | **Correctly raw** — same reason; exact-match tenant confinement. |
| `:273`, `:940`, `:950` | `record.team_id` → `AgentNode` / `LineageStep` | **Correctly raw** — wire passthrough. These echo what was registered; normalizing would misreport stored state. |
| `:787`–`:856` | `team_id` path param in `get_team` | **Correctly raw** — request input, not a record read. |
| `:337-338` | `lineage.team_id` → `PolicyScope::Team(..)` | ⚠️ **UNCONVERTED — and this is the fifth that matters more.** |
| `:1130` | `r.team_id` → `teams_by_id` → `cross_team` | ⚠️ **UNCONVERTED — the finding above; blocked on 5201.** |

**The fifth site, `build_policy_chain` at `:334`/`:337`:**

```rust
if let Some(org_id) = lineage.org_id.as_deref()  { tiers.push(("org",  PolicyScope::Org(org_id.to_owned()))); }
if let Some(team_id) = lineage.team_id.as_deref() { tiers.push(("team", PolicyScope::Team(team_id.to_owned()))); }
```

A blank `team_id` pushes `PolicyScope::Team("")` — a policy tier for a team that does not exist — directly contradicting that function's own doc three lines above (`:325`: *"A tier appears only when the agent has that selector"*). `org_id` at `:334` has the identical defect. This is the same "projection names something it does not distinguish" class the PR set out to fix, in the same file, with `team_of` already defined 270 lines up.

Two things make it more significant than the `teams_by_id` one:

- **It is genuinely independent.** `build_policy_chain` is `topology.rs`-local with no mirror in `edges.rs`, so unlike `:1130` it carries no cross-surface coupling and could be fixed in isolation.
- **It is in no ticket at all** — not 5181, not 5182, and not 5200/5201/5202. It will be silently lost unless one is filed. (I raised it as MINOR #4 in my main review; restating because the sweep confirms it is the only unconverted site with no owner.)

I have **not** fixed it here: it is outside both tickets this PR closes, and expanding a security fix's scope on a coordinator ping is the wrong direction. It needs its own ticket (or folding into 5201, whose "one ticket, all call sites" framing already fits).

### 4. Gates and CI

**No code changed, so the head SHA is still `925d07f1` and every gate and analysis I reported previously still binds to the current head.** I did not re-run `nextest`/`clippy`/`fmt` — there is nothing new to test, and re-running to produce fresh-looking output would be noise. For the record, from the prior run at this same SHA: `cargo clippy -p aa-api --all-targets -- -D warnings` → exit 0; `cargo nextest run -p aa-api` → 1040/1040 passed, 0 skipped.

On the Sonar/label question: since the head did not move, no new analysis is triggered or needed, and **0 new issues still stands for this SHA.** Citing the server-side binding rather than the bot comment, as asked — the `SonarCloud analysis` check-run bound to `925d07f1` completed `2026-07-27T03:26:32Z` (job `89879132356`, run `30233320678`), started `03:20:03Z`, i.e. after its `needs: [coverage]` dependency's real 23-minute run finished at `03:19:47Z`, and after the head commit at `02:33:07Z`. The `run-coverage` label is present. The paginated check-run list is what surfaced that `Coverage` and `SonarCloud analysis` each appear **twice** on this SHA (a stale pre-label `skipped` plus the real post-reopen `success`) — the unpaginated view truncates and can hide exactly that.

The worktree is clean and unmodified at `925d07f1`.

### Summary

- **Finding: real**, and correctly spotted.
- **Fix: not applied** — `edges.rs:173` has the identical raw lookup, so normalizing only `topology.rs:1130` swaps one consistent defect for two contradicting endpoints. It is the `topology.rs` half of **AAASM-5201**, not a separate one-liner. No test added, because adding a test for behaviour I am deliberately not changing would assert the wrong contract.
- **Sweep: two unconverted sites**, `:1130` (blocked on 5201) and `:337`/`:334` (**independent, unticketed — needs an owner**). All other raw reads are correctly raw, with reasons above.
- **Ask:** prioritise **AAASM-5201** to cover `topology.rs:1130` **and** `edges.rs:173` together — it now gates #1726's user-visible correctness — and file a ticket for `build_policy_chain`.

PR #1729, opened 2026-07-27
[AAASM-5187] ♻️ (dashboard): Stop the Capability summary reporting an impossible metric

Chisanan232:
## Description

Three defects on the Capability surface, all governed by **ADR 0026 Decision 2 (Accepted, option A)**: *show only states the current projection can actually emit … remove `narrow` and `approval` from the legend, filters, and mutation controls until a real backend computation can produce them … do not preserve aspirational states merely to match an old mock.*

**AAASM-5187 — the `NARROWED` tile reported an impossible metric as a real `0`.**
`decide()` (`aa-api/src/routes/capability.rs:480-488`) returns only `Allow`/`Deny`, and unmodelled verbs are `Na`, so no projection can produce a narrowed cell — the `0` measured a quantity that cannot exist.

The tile is **removed, not relabelled `NOT EVALUATED`** to match the `FLAGGED AGENTS` tile beside it. That neighbour's treatment is right *because its absence is contingent*: `flagged` is on the wire contract and `countFlagged` turns into a real measurement the day one agent carries a boolean. A narrowed count is unreachable **by construction**, so a tile for it could only ever be a permanent placeholder advertising a state that will never arrive on this endpoint — the aspirational pattern the ADR rejects, not the honest-absence pattern it endorses. `tallyVerdicts` still computes `narrow` (it is the shared vocabulary); only this page stops reporting it, so re-widening is one field if AAASM-5094 ever computes those cells.

**AAASM-5125 — the page defaulted to the `write` verb.**
Verified on `main`: `CapabilityPage.tsx:28` hard-coded `'write'`, which `project_matrix` models on the Filesystem column alone — Terminal, Network-outbound and every MCP-tool column carry `exec` and nothing else (`capability.rs:497-524`, `:626-641`). The landing view was one populated column beside a wall of `n/a`.

The new `defaultVerb` picks the verb with the most non-`na` cells **in the loaded matrix**. For any non-empty fleet that is `agents × 1` for write against `agents × (2 + tools)` for exec, so it resolves to `exec` today — the answer the ticket predicted, but reached from the data rather than pinned, so it follows the backend if the capability families change (AAASM-5090) instead of going stale. An operator's own choice still wins and all four verbs stay on the control.

**AAASM-5154 — the matrix row header had no navigation to the agent.**
Verified missing on `main`: `CapabilityMatrixGrid.tsx:137-140` had no click handler, the slot taken by the bulk-select checkbox. The name becomes a button opening `/agents/:id` — the drawer Fleet deep-links to — **beside** the checkbox, not instead of it.

**The route was verified to resolve before restoring the link.** `/agents/:id` is a real nested route (`App.tsx:45-48`); the matrix id is `hex::encode(record.agent_id)` and `GET /agents/{id}` parses exactly that 32-hex form (`agents.rs:86`, `:131`); and the matrix is admin-gated (`capability.rs:238-241`), so every caller who can see a row can also read that agent. Both the page test and the e2e assert the drawer *renders*, not merely that a URL was pushed — Trace shipped a link to a route that 404'd (AAASM-5109) and a URL-only assertion would have passed on it. A grid rendered without the handler keeps the name as plain text.

### Housekeeping

`dashboard/verify/5124/bulk-options-*.png` were orphaned when that run's shot was renamed to `bulk-no-selection-*`, and still showed a bulk bar with a pre-selected decision — a control AAASM-5124 removed. Re-shot from the current build rather than deleted.

### Not in scope, flagged for triage

The agent-detail drawer (`components/agentDetail/`, not this lane's) shows the same two shapes: its **posture summary lists permanent `narrow decisions —` / `approval decisions —` rows**, and its **capability snapshot defaults to `write`**. Both are already honest (they render `—` with an explanation rather than a `0`), but they are the same family as 5187/5125 on a surface this PR does not own.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [x] ♻️ Refactoring
- [ ] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Frontend-only. No wire contract changes — `Decision` keeps all five members, `VerdictTally` keeps `narrow`, and `pnpm generate:api` produces no diff.

## Related Issues

- Related Jira tickets:
  - [AAASM-5187](https://lightning-dust-mite.atlassian.net/browse/AAASM-5187) — NARROWED tile reports an impossible metric as a real 0
  - [AAASM-5125](https://lightning-dust-mite.atlassian.net/browse/AAASM-5125) — Capability page defaults to the `write` verb
  - [AAASM-5154](https://lightning-dust-mite.atlassian.net/browse/AAASM-5154) — Restore the matrix row header's navigation to the agent
- Related GitHub issues: N/A

## Testing

**Every test was verified load-bearing by reverting the source and watching it go red** (`git checkout dab8a70b -- <paths>`, restored afterwards — never `git stash`):

| Reverted | Result |
|---|---|
| `CapabilitySummary.tsx` + `summary.ts` | 4 failures across `CapabilitySummary.test.tsx`, `summary.test.ts`, `CapabilityPage.test.tsx` |
| `CapabilityPage.tsx` (verb wiring, `verb.ts` kept) | 3 failures incl. *lands on the verb the loaded matrix populates* |
| `CapabilityMatrixGrid.tsx` | 2 failures incl. *opens the agent detail route from a matrix row header* |

E2E (`tests/e2e/review-5187.config.ts`, port 4616, repo-local `@playwright/test`, light **and** dark, console errors `[]`): 8 passed. Re-verified against a **rebuilt `dist`** with the three source files reverted (`vite build` exit 0, `dist` removed first, preview restarted — a stale `dist` would have made the mutation check silently pass): the 5187/5125 and 5154 runs failed in both themes, then passed again once restored. Screenshots committed to `dashboard/verify/5187/`.

Gates, each run once from `dashboard/`:

- `pnpm test` — **264 files, 2822 tests passed**
- `pnpm type-check` — clean
- `pnpm lint` — clean (`--max-warnings 0`)
- `pnpm generate:api` → `git status --porcelain` **empty**
- No Rust suite touched.

## Checklist

- [x] Documentation updated if behaviour changed (doc-comments carry the WHY; no ADR change — this implements an already-Accepted decision)
- [ ] All CI checks passing

[AAASM-5187]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232 (commented):
## Adversarial review — PR #1729

**Reviewed head SHA: `d18abdae46816e85c2e7ae09a3778add30f8d2ae`** (branch `v0.0.1/AAASM-5187/fix/capability_parity`, 22 files, +742/−30, base `main`). Every claim below was re-derived at that SHA in a clean worktree; nothing was taken from the PR body on trust.

**Verdict: no blockers. The core engineering judgement is correct and independently verified.** Three findings are process/traceability, not code. This is a COMMENT review — not an approval.

---

### Was removing the NARROWED tile (rather than relabelling it) correct? — Yes.

The PR rests its whole argument on one distinction, and **the distinction holds under independent verification**:

`narrow` is unreachable **by construction**. Every write path was checked, not just the one the PR cites:

- `decide()` — `aa-api/src/routes/capability.rs:480-488` — returns only `Decision::Deny` / `Decision::Allow`. There is no third return.
- `system_cell()` — `:497-524` — hard-codes `Decision::Na` for every verb a family does not model (filesystem has no `exec`; terminal/network have no `read`/`write`/`delete`).
- Tool cells — `:628`, `:635-637` — emit only `Deny` / `Na`.
- The last remaining write path is closed too: `POST /capability/override` **400s** on `Decision::Narrow | Decision::Approval` (`:308-317`). So an override cannot inject a narrowed cell either, and — since AAASM-5124 removed those options from `BulkActionBar` — the optimistic shadow can't paint one in flight.

No input to this endpoint produces a narrowed cell. The old tile's `0` measured a quantity that cannot exist.

`flagged` is genuinely different, and the PR is right that its absence is only *contingent*: `pub flagged: Option<bool>` is a real field on the wire model (`aa-api/src/models/capability.rs:232`), omitted when `None` (test `flagged should be omitted when None`, `:515`) and serialized the moment anything sets it. `countFlagged` becomes a real measurement on that day. So `NOT EVALUATED` is the correct treatment there and the wrong treatment for `narrow` — relabelling the narrowed tile would have shipped a permanent placeholder advertising a state that can never arrive.

**ADR-0026 Decision 2 compliance assessment — complies, does not exceed.** D2 is `Accepted 2026-07 (option (A))`; the recorded sign-off is *"show only states the current projection can actually emit … remove narrow and approval from the legend, filters, and mutation controls until a real backend computation can produce them … do not preserve aspirational states merely to match an old mock."* A summary tile is not literally one of the three enumerated controls, but it is squarely inside the second clause, and the ADR's shared context frames the whole programme as removing surfaces that "render a confident answer to a question [they have] no data for". The PR stays inside the boundary in every direction that matters: `Decision` keeps all five members, `tallyVerdicts` still computes `narrow` (`lib/truthfulness/verdict.ts`), the wire contract is untouched, and `pnpm generate:api` produces no drift. Re-widening if AAASM-5094 ever lands is one destructured name plus one `<SummaryStat>`. It also does **not** exceed D2 — `unconfigured` is not added to any legend, matching the ADR's own explicit carve-out.

**The one honest counterargument, recorded rather than dismissed:** AAASM-5187's own Expected text asks for `NOT EVALUATED` **or `NOT SUPPORTED`**, and the truthfulness vocabulary it quotes contains `Not supported` precisely for the structural case. The agent-detail drawer already ships that treatment (an explained absence). So "remove" vs "explain the absence" is a real product choice, not a mechanical one — see MAJOR-3. For *this* surface, D2's Accepted text says remove, and D2 governs. I agree with the call.

---

### Jira scope coverage

| Ticket | Stated target behaviour | Shipped | Verdict |
|---|---|---|---|
| **AAASM-5187** (Bug, To Do) | *"The correct rendering is `NOT EVALUATED` (or `NOT SUPPORTED`), not `0`"* | Tile **removed** | Substantively correct, **deviates from the ticket text** — see MAJOR-1 |
| **AAASM-5125** (Task, To Do) | *"Default the verb to whichever has the most non-`na` cells … **or** hard-default to `exec`"* | `features/capability/verb.ts#defaultVerb`, option 1 | ✅ Met (ticket offered both; PR took the more durable one) |
| **AAASM-5154** (Task, To Do) | *"Make the row-header name a button that navigates to the agent drawer …, keeping the bulk-select checkbox alongside it"* | Exactly that | ✅ Met verbatim |

All three tickets are still `To Do` with null resolution. None of them contains a literal `Acceptance Criteria` block — this project puts the testable contract in `### Expected` / `#### Target behaviour`, which is what I checked against.

---

### CI / Sonar / Codecov evidence

All on the reviewed SHA `d18abdae` — **zero checks on any other SHA**:

- **Check runs: 16 success, 23 skipped, 0 failed.** Includes `Dashboard tests + coverage`, `Dashboard (type-check + lint)`, `Dashboard build`, `Dashboard codegen drift check`, `aa-cli build compat (dashboard embed)`, `CodeQL` / `Analyze (javascript-typescript)`, `CI Success`.
- **SonarCloud: Quality Gate passed** — 0 New issues, 0 Accepted issues, 0 Security Hotspots, **100.0% Coverage on New Code**, 0.0% duplication. Comment posted `02:51:01Z`; head committed `02:38:24Z` → **the analysis postdates the head commit**, so it is not a stale carry-over.
- **Codecov: `codecov/patch` success** — *"All modified and coverable lines are covered by tests."* (No `codecov/project` context is reported on this PR; patch-only.)

### Test evidence (re-run locally at `d18abdae`, not taken on trust)

- `pnpm test` → **264 files / 2822 tests passed** — matches the PR body exactly.
- `pnpm type-check` → clean. `pnpm lint` → clean. `pnpm build` → **exit 0** after `rm -rf dist` (so the Playwright preview below served fresh assets, not a stale `dist`).

**Revert-mutation claims spot-checked, all three confirmed exactly** (via `git checkout remote/main -- <path>`, restored after; tree verified clean, never `git stash`):

| Reverted | PR claimed | Observed |
|---|---|---|
| `CapabilityMatrixGrid.tsx` | 2 red | **2 red** ✅ |
| `CapabilitySummary.tsx` + `summary.ts` | 4 red | **4 red** ✅ |
| `CapabilityPage.tsx` (verb wiring) | 3 red | **3 red** ✅ |

Every new test is load-bearing.

### Playwright / visual evidence

Ran an **independent** adversarial suite (repo-local `@playwright/test`, my own lane config on **port 4625**, `reuseExistingServer: false`, against the freshly-built `dist`; not the MCP browser). **14/14 passed, light and dark, console errors `[]` throughout.** States covered beyond the PR's own run:

- **Empty cascade (the AAASM-5106 production shape)** — summary folds to `— UNCONFIGURED` for both `allow` and `denied`, `— NOT EVALUATED` for flagged. **No fabricated `0` anywhere.** ✅
- **Genuine zero** (loaded cascade, no allow cells) — renders `0` with `data-truth-state="known"`. The genuine-zero / unconfigured distinction survives the refactor. ✅
- **Error (500)** — after react-query's default `retry: 3` backoff, renders the real `ErrorState` with a Retry action; **no tiles rendered at all**, nothing claims `0`. (My first read at ~700 ms showed a skeleton — that was inside the retry window, not a defect.) ✅
- **Empty fleet** — proper `EmptyState` ("No capability surface to map"), no verb breakage. ✅
- **All-`na` / degenerate grid** — `defaultVerb` falls back deterministically to `read` (`VERBS[0]`), page renders, nothing breaks. ✅
- **Keyboard & semantics** — tab order is checkbox → name button (matches visual order); the checkbox remains operable and is untouched by the button; focus is visible in both themes (`outline-style: auto`, `outline-width: 1px`, **plus** the underline — the UA ring is not suppressed); `Enter` on the row-header button opens the drawer. ✅
- **WCAG AA (ADR-0027, Accepted)** — the **new** control `.cap-mx-row-h-link` measures **19.3:1 light / 14.84:1 dark**. Summary tiles: allow 19.3/14.84, deny 8.3/8.99, flagged 5.93/9.46. All comfortably above 4.5:1. ✅
- **`aria-label="open agent <name>"`** satisfies WCAG 2.5.3 *Label in Name* — the visible text (`<name>`) is contained in the accessible name.

**Committed evidence verified against current code**, not just present:

- `verify/5187/summary-and-landing-verb-light.png` — three tiles (`TOTAL "ALLOW" CELLS (EXEC) 4`, `DENIED 2`, `FLAGGED AGENTS — NOT EVALUATED`), `EXEC` selected. Matches current behaviour.
- `verify/5124/bulk-options-{light,dark}.png` — **the regenerated shots do show the current bulk bar**: `1 agent selected · APPLY [select a decision] FOR [exec] ON [Filesystem] [Record display-only override] [Clear]`. No pre-selected decision, and the AAASM-5178 display-only labelling is present. The stale-evidence housekeeping claim is genuine.

### Business-logic assessment

`defaultVerb` is **derived from data, not taste**, and the reasoning checks out. It counts non-`na` cells per verb over the loaded matrix. Against the shape `project_matrix` actually emits (read/write/delete on Filesystem only; `exec` on Terminal, Network-outbound and every MCP-tool column — `capability.rs:497-524`, `:626-641`), that is `agents × 1` for `write` against `agents × (2 + tools)` for `exec`, so it resolves to `exec` today *from the data* rather than being pinned there. Tie-break is `n > bestCount` (strict), so ties fall to the first entry of `VERBS = ['read','write','delete','exec']` — deterministic, and honestly documented as arbitrary. The degenerate empty/all-`na` grid returns `read` and is verified not to break the page (which shows `EmptyState`/`LoadingState` there anyway).

Two details I checked specifically and liked: the default is derived from the **fetched** matrix, not the optimistic override shadow — so recording an `na` override cannot shift another operator's landing verb; and an explicit operator choice (`chosenVerb`) wins outright and permanently, with all four verbs still on the control.

### Security assessment

- **No unsupported Narrow or Approval behaviour is restored anywhere in the diff.** Grepped the full diff: `narrow` appears only in doc-comments explaining its removal and in negative test assertions. No legend entry, no filter, no override option, no optimistic path.
- **No missing policy data becomes a valid-looking verdict** on this PR's surface. The AAASM-5106 guard is intact — `summarizeMatrix` still calls `tallyVerdicts(cellDecisions(...), cascade)`; the destructure changed *what is read*, not *what is guarded*. Verified live: empty cascade → `UNCONFIGURED`, unavailable → `unknown`, genuine zero → `0`.
- **`defaultVerb` cannot fabricate a verdict** — it only chooses which column to display; it asserts no value.
- **Route/authorization cannot strand a row** (AAASM-5109 class defect). `/agents/:id` is a real nested route (`App.tsx:45-48`, `AgentDetailPage` under `FleetPage`, inside `ProtectedRoute`). The id matches on both sides: the matrix emits `hex::encode(record.agent_id)` (`capability.rs:606`) and `parse_agent_id` decodes exactly that 32-hex form, rejecting odd-length/non-hex cleanly with a 400 rather than panicking (`agents.rs:86-95`, the AAASM-4018 fix). The matrix is admin-gated (`capability.rs:238-241`), so every caller who can see a row can read that agent. `encodeURIComponent(agent.id)` on the navigate is a no-op for hex but is correct defensive hygiene.
- **Default actions cannot trigger unintended bulk writes — PR #1718 verified NOT regressed** (live, not by inspection): no bulk bar at all without a selection; on selection the decision select has value `''` (no default); `narrow`/`approval` options absent from the DOM; the primary action reads `Record display-only override`; **no `POST /capability/override` fired** during the whole interaction. The new row-header button **cannot become a bulk trigger** — clicking it navigates and leaves the checkbox unchecked (asserted live and in `CapabilityMatrixGrid.test.tsx` via `expect(onToggleSelect).not.toHaveBeenCalled()`), and the row header carries no click handler for it to bubble into.

### Regression assessment

Diffed every changed test against `remote/main`. **None weakened.** `CapabilityMatrixGrid.test.tsx` and `verb.test.ts` are pure additions. The removals in `CapabilitySummary.test.tsx` and `summary.test.ts` delete only assertions about the now-absent `narrow` field, and each is *replaced by a stronger negative*: `expect(s).not.toHaveProperty('narrow')`, `expect(container.textContent).not.toMatch(/narrow/i)`, and — good instinct — the removal is pinned on **both** a loaded and an empty cascade, which is the case where a relabelled tile would have survived. The `unavailable` propagation test still asserts `unknown` across all remaining fields.

---

## Findings

**BLOCKER** — none.

**MAJOR-1 (process / traceability).** AAASM-5187's `### Expected` explicitly asks for `NOT EVALUATED` (or `NOT SUPPORTED`) and points at the neighbouring `FLAGGED AGENTS` tile as the pattern to copy. This PR does the opposite. I believe the PR is **right** and the ticket is **stale** — 5187 was written `08:15`, and ADR-0026 D2 was only recorded as Accepted when #1718 merged at `10:02`, so the ticket predates the decision that governs it. But the only record of that supersession is this PR's body. There is an exact precedent for doing this properly: AAASM-5124's own stated target (`default the select to 'deny'`) was overridden by engineer decision, and that decision was written **onto the ticket as a comment before implementation**. Please add the equivalent comment to 5187 recording that D2 supersedes its Expected section, so the ticket doesn't read as a failed AC to the next person.

**MAJOR-2 (process).** **AAASM-5197 is orphaned in Jira.** It is typed **`Task`, not `Bug`** (the PR body calls it a reported defect); it has **no parent, no issue links** to 5187/5125/5094, **no labels** (5125/5154 carry `dashboard`/`design-parity`/`truthfulness-cosmetic`), no fix version, **no linked PR, and zero comments**. Its only connection to this lane is prose smartlinks inside its own description. Per the repo's audit-trail rule (ticket must be derivable from the artefact), please add real Jira links 5197 ↔ 5187/5125 and set the type/labels.

**MAJOR-3 (scope split — legitimate, but this PR widens the seam it defers).** Deferring the agent-detail drawer is defensible: different component tree (`components/agentDetail/`), and the drawer is already *honest* — it renders `narrow decisions —` / `approval decisions —` with an explanation, never a fabricated `0`. I agree it is not this lane's to fix. **But** the 5154 half of this same PR builds a **one-click path from the surface where `narrow` was removed straight into the surface where `narrow` is still displayed** — and the PR's own committed evidence shows it: `verify/5187/row-header-to-agent-dark.png` renders the drawer with `narrow decisions —`, `approval decisions —`, and a `write`-defaulted capability snapshot. So after this PR the product presents two different treatments of the identical fact, one click apart. Neither is deceptive; that is why this is not a blocker. The open question is genuinely a product one, and it is not answered by D2's text ("legend, filters, and mutation controls" does not obviously cover an informational summary row): **is the canonical treatment for a structurally-unreachable state *removal* (this PR) or *explained absence* (the drawer)?** 5197 proposes propagating removal — that should be decided, not inherited by precedent. **PRODUCT-DECISION-REQUIRED**; please answer it on 5197 before that ticket is picked up.

**MINOR-4 (test strength).** `CapabilityPage.test.tsx` now asserts the landing verb with `defaultVerb(FIXTURE.agents, FIXTURE.resources)` — i.e. the production function under test. That assertion is tautological: a wrong `defaultVerb` still satisfies it. It is well contained (the sibling test hard-codes `exec` against `EXEC_HEAVY_MATRIX`, and `verb.test.ts` pins the behaviour directly), so no coverage is actually lost — but the previous line hard-coded `write`, so this specific assertion is strictly weaker than what it replaced. Prefer a literal.

**MINOR-5 (ADR hygiene).** `design/v2/hi-fi/capability.jsx` — the authoritative visual spec per ADR-0025 — contains **both** the `narrowed` tile (`:167`, `tone="warn"`) and `useS('write')` (`:11`). The tile removal is authorised by ADR-0026 D2 (Accepted). The **verb-default** change is authorised by no ADR — only by ticket 5125's own text — yet ADR-0017 (Accepted) is the register of ratified deviations from the design spec, and it was not amended. This is MINOR rather than MAJOR only because ADR-0025 is `Proposed`, so v2 is evidence and not implementation authority. Consider a one-line ADR-0017 entry so the deviation is discoverable from the ADR set rather than only from a closed ticket.

**MINOR-6 (behaviour under refetch).** `landingVerb` is recomputed via `useMemo` on every change to `data`, and the app builds its `QueryClient` with bare defaults (`main.tsx:9`), so `refetchOnWindowFocus` is on. For an operator who has *not* chosen a verb, a background refetch that changes the matrix shape can shift the displayed verb with no user action. Low likelihood and low harm, and the deliberate choice to derive from the fetched (not optimistic) matrix already blocks the nastier version of this. Freezing the derived default on first successful load would close it entirely.

**NIT-7.** `tests/e2e/review-5187.config.ts` sets `reuseExistingServer: true` on fixed port 4616. Given the known stale-`dist` trap in this repo, a leftover preview from another lane can silently serve old assets. The PR body documents doing the right thing manually (removed `dist`, rebuilt, restarted); `reuseExistingServer: false` would make that structural. I used `false` on my own port 4625.

**PRE-EXISTING-OUT-OF-SCOPE (not this PR's, listed so they are not lost):**

- **ADR-0027 (Accepted) contrast failure on the same component.** `.cap-mx-row-h-meta` (framework · owner · trust, 10px/400 — *normal* text) measures **3.45:1 light / 3.83:1 dark**, below the AA 4.5:1 floor ADR-0027 requires. The PR did **not** touch it (the CSS diff only adds `.cap-mx-row-h-link` rules), and the control it *did* add passes at 19.3:1/14.84:1. Worth its own ticket.
- **AAASM-5106 (High, To Do)** remains the honest limit of this fix. Under an empty cascade the summary strip now correctly reads `— UNCONFIGURED` — but the **grid cells directly above it still assert `ALLOW`/`DENY` at full confidence** on the same screen, because `CapabilityMatrixGrid` takes no cascade prop. Verified live in both themes. That is 5106's blast radius, not this PR's regression, but it means the page is only half-truthful today.
- `ErrorState kind="generic"` renders *"Backend returned 503 service_unavailable"* for a 500. Generic copy, pre-existing.

### Cross-PR notes

- **PR #1718 (merged, `dab8a70b`)** — bulk-bar safety verified intact at this head, live (see Security). No regression.
- **AAASM-5155** was closed with resolution **`Done`**, not a "Won't Do"/superseded value, so the supersession lives only in the closing comment. That comment also contradicts the ticket's own description on whether the data exists (description: *"the data is already in the payload (`matrix.policies`, resolved exactly as `CellInspectDrawer.tsx:20-31`)"*; closing comment: *"has no data to carry"*), and switches the authoritative mock from v1 to v2 mid-ticket. Jira hygiene only — it does not affect this PR — but if `matrix.policies` really can source a *"narrowed by"* column, that is worth reconciling on AAASM-5094 before it is picked up.
- **AAASM-5094**, the gate everything narrow-related defers to, is untouched (To Do, 8 pts, no comments) and `blocks` 5155. The "re-widening is one field" claim is therefore accurate but currently unexercised.

### Unresolved risks

1. The removal-vs-explained-absence question (MAJOR-3) is unanswered and this PR sets a precedent for it by shipping.
2. AAASM-5106 leaves the grid asserting allow-by-default under an empty cascade, so the Capability page's honesty is currently partial regardless of this fix.
3. 5187's ticket text and the shipped behaviour disagree on the record (MAJOR-1); anyone auditing by ticket rather than by ADR will read this as a missed AC.

### Suggested follow-ups

- Comment on **AAASM-5187** recording that ADR-0026 D2 supersedes its `Expected` section (mirroring the AAASM-5124 precedent).
- Link **AAASM-5197** to 5187/5125, fix its type/labels, and answer the canonical-treatment question on it.
- New ticket: `.cap-mx-row-h-meta` contrast below the ADR-0027 AA floor (both themes).
- Optional: ADR-0017 entry for the verb-default deviation from `design/v2`.

None of the above needs to block the merge. Reviewed at **`d18abdae46816e85c2e7ae09a3778add30f8d2ae`**; not approving — merge decision belongs to the code owner.

PR #1738, opened 2026-07-27
[AAASM-5234] ♻️ (audit): Key AuditLogPage EVENT_META by Map

Chisanan232:
## What changed

`EVENT_META` in `dashboard/src/pages/AuditLogPage.tsx` was a plain object
(`Record<string, …>`) indexed at the row-render site (`:287`) by
`e.event_type`. It is now a `Map`, looked up with `.get()`.

## Why (root cause)

`LogEntry.event_type` is raw wire data typed `type: string` — it is not
constrained to the declared variant names. When the value collides with an
inherited prototype member (`constructor`, `__proto__`, `toString`,
`hasOwnProperty`), a plain-object lookup returns the **inherited** member,
which is truthy, so the `?? { label, chip, icon }` default never fires. The
cell then renders a leaked function/object instead of the raw type. Keying by
`Map` makes the lookup own-keys only, so unknown values (including
prototype-member names) fall through to the default meta as intended.

## Scope note — already-safe sites left untouched

Per the ticket, the following lookups in the same file were **not** changed
because their keys resolve through a Map-with-default or are pre-canonicalised
and are already safe:
- `:62` / `:103` — `DECISION_META`, keyed by the canonicalised `AuditDecision`.
- `:183` / `:186` / `:212` — `counts` keyed via `eventGroupOf(...)` (canonicalised group key).

Only `EVENT_META` (`:32` definition, `:287` lookup) was converted.

## The fix

- `:32` — `const EVENT_META = new Map<string, {…}>([ [key, meta], … ])`
- `:287` — `EVENT_META.get(e.event_type) ?? { label: e.event_type, chip: '', icon: '·' }`

## How to verify / gate results

- `pnpm type-check` — clean (tsc --noEmit, exit 0)
- `pnpm lint` — clean (eslint, --max-warnings 0, exit 0)
- `pnpm test` (scoped to `AuditLogPage`) — 53 passed (53)

**Test red pre-fix:** with the source reverted to the plain-object form, the 4
new prototype-member cases fail (`constructor`, `__proto__`, `toString`,
`hasOwnProperty` each resolve to an inherited member and never reach the
default meta); with the Map fix they pass. Confirmed load-bearing.

## Acceptance criteria

- [x] `EVENT_META` is a `Map`; lookup uses `.get()`.
- [x] An inherited-prototype `event_type` falls to the default meta (not an inherited member).
- [x] Real event types still map to their declared label/icon.
- [x] Already-safe sites (`:62/:103`, `:183/:186/:212`) untouched.
- [x] type-check, lint, and scoped tests green.

Refs AAASM-5234

### DIFF
--- a/dashboard/src/pages/AuditLogPage.test.tsx
+++ b/dashboard/src/pages/AuditLogPage.test.tsx
@@ -181,6 +181,39 @@ describe('AuditLogPage', () => {
     expect(screen.getByTestId('audit-stat-other')).toHaveTextContent('1')
   })
 
+  // ── AAASM-5234: event_type is raw wire (`type: string`), so a value that
+  // collides with a prototype member (`constructor`, `__proto__`) must fall to
+  // the default meta, not resolve to an inherited object member. A plain-object
+  // lookup with `?? default` fails this because the inherited member is truthy;
+  // keying by Map makes the lookup own-keys only.
+  it.each(['constructor', '__proto__', 'toString', 'hasOwnProperty'])(
+    'falls to the default meta for the prototype-member event type %s',
+    async (inherited) => {
+      get.mockResolvedValue({
+        data: page([entry({ seq: 950, event_type: inherited, payload: '{}' })]),
+      })
+      renderPage()
+      const row = await screen.findByTestId('audit-row-950')
+      // The default meta labels the chip with the raw event type verbatim; a
+      // leaked inherited member would render `[object Object]`/`function ...`.
+      expect(within(row).getByText(new RegExp(inherited))).toBeInTheDocument()
+      const chip = within(row).getByText(new RegExp(inherited))
+      expect(chip.textContent).not.toContain('[object')
+      expect(chip.textContent).not.toContain('function')
+      expect(screen.getByTestId('audit-stat-other')).toHaveTextContent('1')
+    },
+  )
+
+  it('still maps a real event type to its declared meta', async () => {
+    get.mockResolvedValue({
+      data: page([entry({ seq: 951, event_type: 'PolicyViolation', payload: '{}' })]),
+    })
+    renderPage()
+    const row = await screen.findByTestId('audit-row-951')
+    expect(within(row).getByText(/Policy Violation/)).toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


Chisanan232 (commented):

Rebased onto current `origin/main` (tip `3239bbe7`, includes the just-merged AAASM-5254); the two commits replayed cleanly with no conflicts, force-pushed as `f2a3438c`.

**Requirement (AAASM-5234)** — Verified against the ticket:
- `EVENT_META` (AuditLogPage.tsx:32) is now a `Map<string, {label; chip; icon}>`; the single unsafe lookup at :287 is `EVENT_META.get(e.event_type) ?? { label: e.event_type, chip: '', icon: '·' }`. The default is preserved and now correctly reached for any `event_type` value.
- `LogEntry.event_type` is keyed as the raw wire string, exactly as required.
- The three already-safe sites are untouched: `DECISION_META` (:62/:103) remains a `Record` keyed by the constrained `AuditDecision` union, and the `eventGroupOf`-keyed counts (:185/:194/:300) are unchanged. Diff touches only `AuditLogPage.tsx` and its test — no scope creep.
- Test is load-bearing: reverting only the production change locally makes exactly the 4 prototype-member cases (`constructor`, `__proto__`, `toString`, `hasOwnProperty`) go red (49 pass), and all 53 pass post-fix. The positive `PolicyViolation` case guards against a regression that would break real lookups.

**CI** — Green. 17 required checks pass (Dashboard type-check+lint, tests+coverage, build, e2e Playwright, codegen drift, aa-cli embed compat, SonarCloud, CodeQL, all Analyze legs), 0 failures; change-detection correctly skipped the non-dashboard suites.

**Security** — The prototype-pollution path is closed. A plain-object `[key] ?? default` resolves inherited members (`constructor`, `toString`, …) to truthy objects/functions, leaking `[object Object]`/`function …` into the rendered chip. The `Map.get()` lookup is own-keys-only, so attacker-controlled wire `event_type` values now deterministically fall to the safe default.

No defects found — the change is minimal, correctly scoped, and does what the ticket asks.

PR #1740, opened 2026-07-27
[AAASM-5233] ♻️ (teams): Key TeamDetailPage STATUS_COLOR by Map

Chisanan232:
## What changed
Converts the module-level `STATUS_COLOR` lookup in `dashboard/src/pages/TeamDetailPage.tsx` from a plain object (`Record<string, string>`) to a `Map<string, string>`, and updates the `StatusBadge` lookup from `STATUS_COLOR[status] ?? …` to `STATUS_COLOR.get(status) ?? …`.

## Why (root cause)
`StatusBadge` receives `member.status`, which is the raw wire value from `AgentNode.status` — an unbounded string with no enum guard. The old code indexed a plain object with that value:

```ts
const color = STATUS_COLOR[status] ?? 'var(--text-muted)'
```

When `status` collides with an inherited `Object.prototype` member name (e.g. `constructor`, `toString`, `hasOwnProperty`), the object lookup returns the inherited **function** rather than `undefined`. The `?? 'var(--text-muted)'` fallback therefore never fires, and `color` becomes a function reference that is then spread into the badge's `background` style. A `Map` only returns entries that were explicitly set, so unknown / prototype-named statuses now correctly fall through to the muted color.

## The fix
- `STATUS_COLOR` is now `new Map<string, string>([...])` with the same three entries (`active`, `suspended`, `deregistered`).
- Lookup uses `STATUS_COLOR.get(status) ?? 'var(--text-muted)'`.
- No behavior change for the three known statuses; only the fallback path for unknown/inherited names is corrected.

## Acceptance checklist
- [x] `STATUS_COLOR` is a `Map`, keyed by status string.
- [x] Lookup uses `.get()` with the `var(--text-muted)` fallback preserved.
- [x] Known statuses (`active`, `suspended`, `deregistered`) map to their original colors.
- [x] Inherited-prototype status names fall to `var(--text-muted)`.
- [x] Test added asserting both the mapped and the muted-fallback paths.

## How to verify / gate results
Run from `dashboard/` with Node 22:
- `pnpm exec vitest run src/pages/TeamDetailPage.test.tsx` → **7 passed**
- `pnpm type-check` (`tsc --noEmit`) → **clean, exit 0**
- `pnpm lint` (eslint, `--max-warnings 0`) → **clean, exit 0**

## Test red pre-fix (load-bearing confirmation)
The new test `colors known statuses from the map and unknown statuses fall to the muted color` was run against the pre-fix plain-object version and **failed** at the `constructor`-status assertion:

```
FAIL … > colors known statuses from the map and unknown statuses fall to the muted color
expect(element).toHaveStyle()
- background: var(--text-muted);
  at src/pages/TeamDetailPage.test.tsx:167  // badges[2], status "constructor"
```

With the Map fix in place the same test passes, confirming it exercises the exact defect.

Closes AAASM-5233

### DIFF
--- a/dashboard/src/pages/TeamDetailPage.test.tsx
+++ b/dashboard/src/pages/TeamDetailPage.test.tsx
@@ -139,6 +139,35 @@ describe('TeamDetailPage', () => {
     expect(screen.getByTestId('location')).toHaveTextContent(`/topology?root=${rootId}`)
   })
 
+  it('colors known statuses from the map and unknown statuses fall to the muted color', async () => {
+    // 'constructor' and 'toString' are inherited Object.prototype members: a
+    // plain-object lookup returns the inherited function for these key names and
+    // defeats the `?? var(--text-muted)` fallback. A Map returns only the
+    // explicitly-set entries, so both fall through to the muted color.
+    const team: TeamTopology = {
+      team_id: 'team-proto',
+      agent_count: 4,
+      members: [
+        { id: 'a'.repeat(32), name: 'known-active', status: 'active', depth: 0, team_id: 'team-proto', mode: 'enforce', flagged: false, trust: null },
+        { id: 'b'.repeat(32), name: 'known-suspended', status: 'suspended', depth: 0, team_id: 'team-proto', mode: 'enforce', flagged: false, trust: null },
+        { id: 'c'.repeat(32), name: 'proto-constructor', status: 'constructor', depth: 0, team_id: 'team-proto', mode: 'enforce', flagged: false, trust: null },
+        { id: 'd'.repeat(32), name: 'proto-tostring', status: 'toString', depth: 0, team_id: 'team-proto', mode: 'enforce', flagged: false, trust: null },
+      ],
+    }
+    mockTeam({ data: team })
+    mockCosts()
+    mockLineage('a'.repeat(32))
+    render(<TeamDetailPage />, { wrapper: ({ children }) => <Wrapper initialEntries={['/teams/team-proto']}>{children}</Wrapper> })
+    const badges = await screen.findAllByTestId('team-member-status')
+    expect(badges).toHaveLength(4)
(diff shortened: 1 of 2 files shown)


Chisanan232 (commented):
✅ LGTM — reviewed and approved by Claude Code (sign-off of record).

Reviewed on three axes after rebasing onto fixed `main` (clean rebase, no conflicts; force-pushed `ebd464f3`).

**Requirement (AAASM-5233)** — met exactly. In `dashboard/src/pages/TeamDetailPage.tsx`, `STATUS_COLOR` is now a `Map<string, string>` seeded with the three real statuses (`active` → `var(--status-success-solid)`, `suspended` → `var(--status-warning-solid)`, `deregistered` → `var(--text-muted)`). The lookup became `STATUS_COLOR.get(status) ?? 'var(--text-muted)'`, fed from `member.status` (raw wire, via `<StatusBadge status={member.status} />`). Real statuses map to their colors; unknown/inherited status keys fall to the muted default. The added test is load-bearing: I confirmed the `constructor`/`toString` assertions go **red** against the previous plain-object implementation (fail at the `badges[2]` muted-color assertion) and **green** on the Map, so the test genuinely locks in the behavior change.

**CI** — green. `CI Success` passed; 0 failures. Dashboard tests + coverage, Dashboard build, e2e (Playwright), type-check + lint, codegen drift, codecov/patch, CodeQL, and SonarCloud all pass; path-filtered non-dashboard jobs correctly skip.

**Security** — prototype-pollution path closed. A plain-object/`Record` lookup returns inherited `Object.prototype` members (`constructor`, `toString`, `hasOwnProperty`, …) for those key names, which defeats the `?? var(--text-muted)` fallback and yields a function as the color. `Map.get()` returns only explicitly-set entries, so wire-controlled status names that collide with prototype members now correctly resolve to the muted default. Verified by the new test.

No defects found.

PR #1742, opened 2026-07-27
[AAASM-5241] 🐛 (components): Keep __proto__ child ids in subtree-burn chart rows

Chisanan232:
## What changed
`SubtreeBurnChart.tsx`'s `transform()` built each chart row as a plain object literal and then wrote per-child burn with `row[cid] = ...`, where `cid` is the `child_agent_id` from `GET /api/v1/agents/{id}/subtree-burn`. Rows are now built with `Object.create(null)` (via `Object.assign(Object.create(null) as ChartRow, { date, total })`) so every child id — including `__proto__` — is stored as a real own property.

## Why (root cause)
On a normal object, `row['__proto__'] = number` does not create an own property; it is a **silent no-op** prototype assignment (the JS engine ignores a non-object/null value). So a child whose `child_agent_id` is exactly `__proto__` had no data key on the row, its `<Area>` dataKey resolved to nothing, and that child's burn silently vanished from the stacked chart with no error. A null-prototype object has no `__proto__` accessor, so the assignment lands as an ordinary own property that Recharts picks up like any other child.

## How to verify
`transform()` is exported (recharts renders no `<Area>` SVG under jsdom, so the row contract is asserted directly) with a scoped, commented `react-refresh/only-export-components` disable, matching the existing test-only `BurnTooltip` export precedent.

New tests in `SubtreeBurnChart.test.tsx`:
- a `__proto__` child id is retained as an **own** row property carrying its burn (`hasOwnProperty` true, value `9`);
- real child ids still populate (`child-1` = 3), and `date`/`total` are unaffected.

**Test is load-bearing:** reverting the row to the plain-object literal makes the `__proto__` test fail red (`hasOwnProperty('__proto__')` → false — the child was dropped); the fix turns it green.

## Acceptance criteria
- [x] A `__proto__` child id is retained instead of being a silent no-op — its burn row no longer vanishes.
- [x] Real (non-`__proto__`) child ids still populate the chart rows.
- [x] Row `date` and `total` fields unchanged.

## Gate results (Node 22.23.1, pnpm 10.9.0)
- `pnpm type-check` — pass (0 errors)
- `pnpm lint` — pass (0 errors, 0 warnings)
- `pnpm test SubtreeBurnChart` — 10 passed (was 8; +2 regression tests)

Closes AAASM-5241

### DIFF
--- a/dashboard/src/components/SubtreeBurnChart.test.tsx
+++ b/dashboard/src/components/SubtreeBurnChart.test.tsx
@@ -2,7 +2,7 @@ import { render, screen } from '@testing-library/react'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest'
 import type { UseQueryResult } from '@tanstack/react-query'
-import { SubtreeBurnChart, BurnTooltip } from './SubtreeBurnChart'
+import { SubtreeBurnChart, BurnTooltip, transform } from './SubtreeBurnChart'
 import * as agentsApi from '../features/agents/api'
 import type { SubtreeBurn } from '../features/agents/api'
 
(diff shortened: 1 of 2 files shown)


Chisanan232 (commented):

**Requirement (AAASM-5241) — met.** `transform()` in `dashboard/src/components/SubtreeBurnChart.tsx` now builds each `ChartRow` on a null-prototype object via `Object.assign(Object.create(null) as ChartRow, { date, total })`. On the previous plain `{}` literal, `row['__proto__'] = number` was a silent prototype write (a no-op for the row's own data), so a child whose `child_agent_id === '__proto__'` was dropped from the stacked chart with no error. On a null-proto object the assignment becomes a real own property, so the child's burn is retained. Real child ids still populate via the same bracketed loop, and `cid` remains the raw wire `child.child_agent_id`. `ChartRow` typing (`{ date; total; [childId: string]: string | number }`) is sound and unchanged. The regression test is load-bearing: `Object.prototype.hasOwnProperty.call(row, '__proto__')` is `false` on a pre-fix `{}` row and `true` here, and the sibling test confirms real ids (`child-1`) and `total` still populate.

**CI — green.** `CI Success` aggregate passed. Relevant jobs: Dashboard tests + coverage ✓, Dashboard type-check + lint ✓, Dashboard build ✓, Dashboard e2e (Playwright) ✓, Dashboard codegen drift ✓, aa-cli build compat (dashboard embed) ✓, CodeQL/Analyze (js/ts, go, python, rust) ✓, SonarCloud ✓, codecov/patch ✓. Clean after rebase onto fixed main.

**Security — the silent-drop path is closed.** A `__proto__`-named child id no longer vanishes silently. Recharts consumption is unaffected: `<Area dataKey={cid}>` and `<Line dataKey="total">` read row values via bracketed lookups on own properties, which a null-prototype object serves normally; `date`/`total`/per-child keys are all own props. The `// eslint-disable-next-line react-refresh/only-export-components` is scoped to the single non-component `transform` export and justified in-line (jsdom renders no `<Area>` under Recharts, so the row contract is tested against the transform directly). Verified load-bearing — removing the disable reintroduces exactly one `react-refresh/only-export-components` warning; it is not a blanket suppression, and it does not silence any other rule. `BurnTooltip` was already a test-only export but needed no disable since it is itself a component; `transform` genuinely trips the rule, hence the targeted disable.

No defects. Approving.

PR #1743, opened 2026-07-27
[AAASM-5239] 🐛 (analytics): Keep __proto__ segments in cost-breakdown chart rows

Chisanan232:
## What changed
`transformBuckets` in `dashboard/src/features/analytics/costBreakdownUtils.ts` now builds each chart row on a null-prototype object (`Object.create(null)`) instead of a plain object literal, then sets `row.label` and each segment key as own properties.

## Why (root cause — the quietest failure in the Epic)
Segment keys come from `GET /api/v1/analytics/cost-breakdown` through an unvalidated `res.json() as T` cast. On a plain object literal, `row["__proto__"] = numericValue` is a **silent no-op**: JavaScript treats `__proto__` as the prototype accessor, and assigning a number to it is ignored rather than creating an own property. The result is a segment that **vanishes from the chart with no crash, no wrong value, and no log** — the quietest failure mode in Story 5214 / Epic 5208. A null-prototype object has no `__proto__` accessor, so the key is stored as an ordinary own property and the segment is retained.

Downstream consumers are unaffected: recharts reads row data via bracketed `dataKey` lookups, which work identically on a null-proto object, and `label` is still present for the X axis.

## How to verify
- Automated regression added in `dashboard/src/features/analytics/CostBreakdownPanel.test.tsx` (`transformBuckets` describe block). Note: there is **no** `costBreakdownUtils.test.ts` sibling — coverage for these utils lives in the panel test file.
- The new test asserts `Object.prototype.hasOwnProperty.call(row, '__proto__') === true` and `row['__proto__'] === 42`.
- **Verified load-bearing**: reverting the fix to the plain-object row turns the new test red with `expected false to be true` on the `hasOwnProperty('__proto__')` assertion — the exact silent-drop symptom.

## Acceptance criteria
- [x] A cost segment keyed `__proto__` is retained in the transformed chart row instead of silently disappearing.
- [x] No crash, and existing segments/label behaviour unchanged.
- [x] Regression test added and confirmed to fail before the fix, pass after.

## Gate results (Node 22 via nvm, pnpm 10.9.0)
- `pnpm type-check`: pass (tsc --noEmit, 0 errors)
- `pnpm lint`: pass (eslint, exit 0, --max-warnings 0)
- `pnpm test`: pass (268 files, 2942 tests)

Refs AAASM-5239

### DIFF
--- a/dashboard/src/features/analytics/CostBreakdownPanel.test.tsx
+++ b/dashboard/src/features/analytics/CostBreakdownPanel.test.tsx
@@ -101,6 +101,20 @@ describe('transformBuckets', () => {
     expect(rows[0]).toEqual({ label: 'Jan', 'agent-a': 120, 'agent-b': 80 })
     expect(rows[1]).toEqual({ label: 'Feb', 'agent-a': 200, 'agent-b': 60 })
   })
+
+  it('retains a "__proto__" segment as an own property instead of silently dropping it', () => {
+    // With a plain-object row, `row['__proto__'] = value` is a silent no-op:
+    // it reassigns the prototype slot rather than creating an own property, so
+    // the segment vanishes from the chart with no crash and no wrong value.
+    const rows = transformBuckets([
+      {
+        label: 'Jan',
+        segments: [{ key: '__proto__', name: 'Sneaky', value: 42 }],
+      },
+    ])
+    expect(Object.prototype.hasOwnProperty.call(rows[0], '__proto__')).toBe(true)
+    expect(rows[0]['__proto__']).toBe(42)
+  })
 })
 
 describe('formatUsd', () => {
--- a/dashboard/src/features/analytics/costBreakdownUtils.ts
+++ b/dashboard/src/features/analytics/costBreakdownUtils.ts
@@ -29,7 +29,12 @@ export function computeSegmentTotals(buckets: CostBucket[]): Map<string, number>
 
 export function transformBuckets(buckets: CostBucket[]): Record<string, string | number>[] {
   return buckets.map(b => {
-    const row: Record<string, string | number> = { label: b.label }
+    // Build on a null-prototype object so a segment keyed "__proto__" is stored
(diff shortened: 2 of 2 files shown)


Chisanan232 (commented):

Rebased onto current `origin/main` (clean, no conflicts; force-pushed as `5d7d965`) and reviewed on three axes.

**Requirement (AAASM-5239) — met.** `transformBuckets` now builds each row on `Object.create(null)` (`row.label = b.label`, then `row[s.key] = clampChartValue(s.value)`), so a segment keyed `__proto__` is stored as an own data property (42) instead of the previous silent no-op that reassigned the prototype slot and dropped the segment from the chart with no crash, wrong value, or log. Real segment keys still populate. Recharts consumes rows via bracketed `dataKey` reads (`dataKey="label"`, `dataKey={seg.key}`) that resolve correctly on a null-prototype object, and `getUniqueSegments` iterates the source `segments` array so a `__proto__` segment still yields its `<Bar>`. The regression test in `CostBreakdownPanel.test.tsx` is load-bearing — I verified it fails on the pre-fix plain-object literal (`Object.prototype.hasOwnProperty.call(row, '__proto__')` is `false`) and passes with the fix.

**CI — green.** All required checks pass on the rebased head `5d7d965`: CI Success, CodeQL / Analyze (js-ts, python, rust, go), Dashboard type-check+lint, Dashboard build, Dashboard tests + coverage, Dashboard e2e (Playwright), Dashboard codegen drift, SonarCloud, codecov/patch, and aa-cli dashboard-embed build compat. Non-dashboard jobs correctly skip via path filtering. Locally: lint clean, `tsc --noEmit` clean, 19/19 panel tests pass.

**Security — no new risk.** The silent-drop path is closed. Null-prototype rows carry no inherited methods and are only bracket-read by recharts / rendered by React (no prototype-method calls), so this is a strictly safer, defensive hardening of an unvalidated-API-response key.

PR #1748, opened 2026-07-27
[AAASM-5243] 🐛 (alerts): Validate rule.metric and key METRIC_CATEGORY by Map

Chisanan232:
## Description

`deriveCategory` resolves an alert's category from `rule.metric`. Although
`METRIC_CATEGORY` was declared as `Record<AlertMetric, AlertCategory>` and
`rule.metric` typed `AlertMetric`, that annotation is unenforced at runtime:
`features/alerts/api.ts` returns `(await response.json()) as T` — a bare cast —
so `rule.metric` is raw wire data wearing an annotation the transport never
checks. A malformed or hostile payload can therefore set `metric` to a bogus
string or, worse, an inherited object member.

**Root cause + blast radius.** The old lookup `METRIC_CATEGORY[rule.metric] ?? 'uncategorized'`
is a plain-object index. For a wire value of `"constructor"` / `"__proto__"` /
`"toString"`, the index resolves to the inherited prototype member (a function)
instead of `undefined`, so `?? 'uncategorized'` never fires and `deriveCategory`
returns an out-of-union value. Downstream:
- `AlertCardFeed.tsx` `CATEGORY_META[cat].label` → `CATEGORY_META[<bogus>]` is `undefined` → **TypeError**.
- `alertCategory.ts` `categoryCounts` `counts[deriveCategory(...)] += 1` → `undefined += 1` → **NaN**, plus a bogus own key leaked into the counts object.

**Fix (validate-then-Map).**
1. `♻️` Convert `METRIC_CATEGORY` to a `ReadonlyMap<AlertMetric, AlertCategory>`. A `Map` only ever holds its own keys, so an unknown/inherited key is a clean miss.
2. `🐛` Validate `rule.metric` against the `AlertMetric` union with an allow-list guard (`isAlertMetric`, the pattern from `components/fleet/StatusChip.tsx`) before the lookup. On failure `deriveCategory` returns `'uncategorized'` immediately.

Both together close the boundary: no out-of-union value can escape `deriveCategory`.

## Type of Change

- [x] 🐛 Bug fix
- [x] ♻️ Refactoring

## Breaking Changes

- [x] No — valid metrics map to the same categories; only invalid/hostile wire values change from (silent corruption / crash) to `uncategorized`.

## Related Issues

- Related Jira ticket: AAASM-5243 (Subtask of Story AAASM-5215, Epic AAASM-5208)

## Testing

- [x] Unit tests added / updated — extended `alertCategory.test.ts`
- [x] Manual testing performed — gates run locally (Node 22 / pnpm 10.9.0)

**Test red pre-fix (load-bearing):** against the pre-fix code (plain-object
index, no guard) the new assertions fail 5/10 — the `constructor` / `__proto__` /
`toString` / `hasOwnProperty` cases resolve a prototype member instead of
`uncategorized`, and the `categoryCounts` case shows `uncategorized` stuck at 0
with the count inflated on a bogus key. With the fix all 10 pass; the 5
pre-existing tests (real metrics still map) stayed green throughout.

**Gate results (local):**
- `pnpm type-check` → exit 0
- `pnpm lint` → exit 0 (`--max-warnings 0`)
- `pnpm exec vitest run src/features/alerts/alertCategory.test.ts` → 10 passed

## Acceptance Criteria

- [x] `METRIC_CATEGORY` is a `Map`.
- [x] `rule.metric` is validated against the `AlertMetric` union before use.
- [x] An invalid/inherited/non-string metric yields `uncategorized` (no TypeError, no NaN, no bogus own key).
- [x] Real metrics still map to their spec categories.

## Checklist

- [x] Code follows project style guidelines (eslint clean, `--max-warnings 0`)
- [x] All local gates passing

### DIFF
--- a/dashboard/src/features/alerts/alertCategory.test.ts
+++ b/dashboard/src/features/alerts/alertCategory.test.ts
@@ -60,6 +60,50 @@ describe('deriveCategory', () => {
   it('falls through to uncategorized when the rule is not loaded', () => {
     expect(deriveCategory(alert('e', 'r-missing'), byId)).toBe('uncategorized')
   })
+
+  // `rule.metric` is raw wire data wearing an unenforced `AlertMetric`
+  // annotation — the alerts client casts the response with a bare `as T`
+  // (see `api.ts`). A malformed or hostile payload can set `metric` to a bogus
+  // string, an inherited object member (`constructor` / `__proto__`), or a
+  // non-string. The metric must be validated against the `AlertMetric` union
+  // and fall to `uncategorized`, never resolve a prototype member or produce an
+  // out-of-union key that makes downstream `CATEGORY_META[cat]` /
+  // `counts[cat] += 1` throw or go NaN.
+  it.each([
+    ['a plain unknown metric', 'not_a_metric'],
+    ['the inherited "constructor" key', 'constructor'],
+    ['the inherited "__proto__" key', '__proto__'],
+    ['the inherited "toString" key', 'toString'],
+    ['the inherited "hasOwnProperty" key', 'hasOwnProperty'],
+  ])('validates out %s and returns uncategorized', (_label, metric) => {
+    const wireRule = rule('r-bad', metric as AlertMetric)
+    const byIdWithBad = indexRulesById([...RULES, wireRule])
+    expect(deriveCategory(alert('f', 'r-bad'), byIdWithBad)).toBe('uncategorized')
+  })
+
+  it('rejects a non-string metric without throwing', () => {
+    const wireRule = rule('r-num', 42 as unknown as AlertMetric)
+    const byIdWithBad = indexRulesById([...RULES, wireRule])
+    expect(deriveCategory(alert('g', 'r-num'), byIdWithBad)).toBe('uncategorized')
+  })
+})
+
+describe('categoryCounts with an invalid wire metric', () => {
+  it('counts an invalid metric as uncategorized, not a bogus own key or NaN', () => {
+    const badRule = rule('r-bad', 'constructor' as AlertMetric)
+    const byId = indexRulesById([...RULES, badRule])
+    const counts: Record<AlertCategory, number> = categoryCounts(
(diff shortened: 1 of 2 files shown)


Chisanan232 (commented):

**Requirement (AAASM-5243).** Met exactly. `METRIC_CATEGORY` is now a `ReadonlyMap<AlertMetric, AlertCategory>` (a `Map` holds only its own keys, so an unknown key is a clean miss — no prototype resolution). A runtime allow-list `KNOWN_METRICS` + `isAlertMetric(value): value is AlertMetric` guard was added, and `deriveCategory` validates `rule.metric` with `if (!isAlertMetric(rule.metric)) return 'uncategorized'` *before* the lookup — the StatusChip allow-list pattern. Invalid, inherited (`constructor`/`__proto__`/`toString`/`hasOwnProperty`) or non-string metrics resolve to `'uncategorized'`, never a prototype member and never a bogus/NaN key. The four real metrics still map to their categories.

**Blast radius closed.** `AlertCardFeed.tsx` `CATEGORY_META[cat]` (`.label`/`.badgeBg`/`.badgeText`) now only ever receives a validated `AlertCategory`, so the TypeError is gone; `categoryCounts` `counts[deriveCategory(...)] += 1` can no longer index a bogus own key or go NaN — verified by the `categoryCounts with an invalid wire metric` test asserting exact key set and `!Number.isNaN` on every count.

**Security (key axis).** The unvalidated wire path — `api.ts:57` `return (await response.json()) as T`, whose bare `as T` never enforces the `AlertMetric` annotation on `rule.metric` — is now guarded at the point of use. Both prototype-pollution (inherited-member lookup) and type-confusion (non-string metric) are closed. Tests are load-bearing: the inherited-key and non-string cases would each resolve/throw under the previous plain-object lookup and now assert `'uncategorized'`.

**CI.** Green after rebase onto fixed `main`. `CI Success` pass; Dashboard tests + coverage, type-check + lint, build, e2e (Playwright), codecov/patch, SonarCloud, and aa-cli build compat all pass.

PR #1751, opened 2026-07-27
[AAASM-5209] ✨ (aa-api): Constrain agent/decision/verdict/event wire vocabularies in OpenAPI

Chisanan232:
## Description

`openapi/v1.yaml` is generated from the `utoipa` derives in the Rust crates — it is never hand-edited. This PR constrains four wire vocabularies **at the derive** (each was a free-form `string`) and then regenerates both the spec and the dashboard TypeScript schema from those derives. The four fields now `$ref` dedicated closed enums:

| Subtask | Field | New enum | Values |
|---|---|---|---|
| AAASM-5218 | `AgentNode.status` (`aa-api/src/models/topology.rs`) | `AgentNodeStatus` | `active` / `suspended` / `deregistered` |
| AAASM-5219 | `AgentDecisionResponse.decisionLabel` (`routes/agents.rs`) | `DecisionLabel` | `allow` / `deny` / `pending` / `redact` / `unspecified` |
| AAASM-5220 | `SimulatePolicyResponse.verdict` (`routes/policies.rs`) | `SimulateVerdict` | `allow` / `narrow` / `approval` / `deny` |
| AAASM-5221 | `LogEntry.event_type` (`routes/logs.rs`) | `LogEventType` | the 22 `AuditEventType::as_str` PascalCase labels |

Each new enum is registered in `aa-api/src/openapi.rs` `components(schemas(...))` so it resolves as a `$ref`. `LogEventType` mirrors `aa_core::AuditEventType` via an exhaustive `From` impl, so adding an audit variant without extending the enum becomes a compile error.

The wire strings are identical to what the handlers already emitted — each new enum's serde representation matches the exact values the pre-existing helper functions produced (`status_str`, `decision_label`, the simulate match arms, `AuditEventType::as_str`), so this is a **schema-tightening only** change with no runtime wire-shape change.

### Explicit exclusions (per the Story)

- **`AgentResponse.status` is deliberately NOT touched** — it emits parameterised Rust variants like `Suspended(Manual)` that a flat enum cannot express.
- **The three differing agent-status vocabularies are NOT reconciled** — the topology `AgentNodeStatus`, the runtime registry `AgentStatus`, and the capability-view `AgentStatus` remain distinct; unifying them is an ADR question outside this Story's scope. `AgentTree.status` was left as `String` for the same reason (only `AgentNode.status` was in scope).

### Generation-only, no dashboard behaviour change

`openapi/v1.yaml` and `dashboard/src/api/generated/schema.d.ts` were regenerated via `cargo run -p aa-api --bin generate_openapi` and `pnpm generate:api` — no hand edits. No dashboard runtime code changed. The only dashboard-side edits are three **test fixtures** narrowed to the now-closed enum types (the regenerated schema tightened `event_type` / `status` from `string` to the enum); the unrecognised-event-type test keeps its out-of-vocabulary value via an explicit cast to preserve its runtime-fallback intent.

## Type of Change

- [x] ♻️ Refactoring (schema-tightening; no behaviour change)

## Breaking Changes

- [x] No

The wire values are unchanged; the spec merely advertises the closed set the API already emitted.

## Related Issues

- Related Jira tickets: AAASM-5209 (Story), AAASM-5218, AAASM-5219, AAASM-5220, AAASM-5221 (subtasks)

## Testing

- [x] Unit tests updated (Rust `decision_label` assertions now compare enum variants; dashboard fixtures narrowed)

Gate results (run in the worktree):

- `cargo build -p aa-api` — pass
- `cargo clippy -p aa-api --all-targets -- -D warnings` — pass (exit 0)
- `cargo fmt --check -p aa-api` — pass
- `cargo test -p aa-api --lib` — **465 passed / 0 failed** (includes all changed code + the `openapi.rs` schema-contract test)
- `cd dashboard && pnpm type-check` (`tsc --noEmit`) — pass (exit 0)
- **Regenerated-artifact drift**: the `v1.yaml` and `schema.d.ts` diffs contain exactly the four `type: string` → `$ref` field flips plus the four new enum schemas, and nothing else — no unexpected drift.

**Pre-existing unrelated failures:** `aa-api/tests/destinations.rs` has 3 failing webhook/SSRF-guard tests (`test_webhook_*`). These are network-sandbox artifacts (outbound/loopback connections return 502 instead of the expected app-level rejection) and were confirmed failing on the pristine base as well. They do not touch any code in this PR (no wire-vocabulary field is involved) and are left for CI, which runs in a network environment where they pass.

## Checklist

- [x] Commits are small and follow the Gitmoji convention (one per field + one codegen commit + one test-fixture commit)

### DIFF
--- a/aa-api/src/models/topology.rs
+++ b/aa-api/src/models/topology.rs
@@ -27,6 +27,36 @@ pub(crate) fn status_str(status: &AgentStatus) -> &'static str {
     }
 }
 
+/// Runtime status of an agent node in the topology projection.
+///
+/// AAASM-5218 — constrains the wire vocabulary of [`AgentNode::status`] at the
+/// OpenAPI derive to exactly the three values `status_str` can emit, so the
+/// generated spec (and the TypeScript client) advertises a closed enum instead
+/// of an unconstrained `string`. Serializes lowercase, matching the strings the
+/// registry's [`AgentStatus`] mapped to before this was a free-form field.
+///
+/// Deliberately distinct from the runtime registry [`AgentStatus`], whose
+/// `Suspended(_)` variant carries a parameterised reason an enum cannot express,
+/// and from the capability-view `AgentStatus` — the three agent-status
+/// vocabularies are not reconciled here (that is an ADR question, see AAASM-5209).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum AgentNodeStatus {
+    Active,
+    Suspended,
+    Deregistered,
+}
+
+impl From<&AgentStatus> for AgentNodeStatus {
+    fn from(status: &AgentStatus) -> Self {
+        match status {
+            AgentStatus::Active => AgentNodeStatus::Active,
+            AgentStatus::Suspended(_) => AgentNodeStatus::Suspended,
+            AgentStatus::Deregistered => AgentNodeStatus::Deregistered,
+        }
+    }
+}
+
 /// Policy-violation count at or above which a node is surfaced as "flagged"
 /// (danger-tinted card + ⚑ marker) in the topology graph.
 ///
@@ -257,7 +287,7 @@ pub struct AgentNode {
     /// Delegation depth — 0 for root agents.
     pub depth: u32,
     /// Runtime status: `active`, `suspended`, or `deregistered`.
-    pub status: String,
+    pub status: AgentNodeStatus,
     /// Team this agent belongs to, if any.
     pub team_id: Option<String>,
     /// Governance level — included only when `show_budget=true`.
@@ -315,7 +345,7 @@ impl From<&AgentRecord> for AgentNode {
             id: format_id(&r.agent_id),
             name: r.name.clone(),
             depth: r.depth,
-            status: status_str(&r.status).to_owned(),
+            status: AgentNodeStatus::from(&r.status),
             team_id: r.team_id.clone(),
             governance_level: None,
             mode: agent_mode(r),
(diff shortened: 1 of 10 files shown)


Chisanan232 (commented):
## ✅ LGTM — reviewed and approved by Claude Code

_(Recorded as a review comment because GitHub blocks self-approval on an own-authored PR; this is the sign-off of record.)_

### Requirement correctness
All four fields are constrained at the utoipa derive and regenerated, not hand-edited:
- **`AgentNodeStatus`** (`models/topology.rs`) — 3 variants matching `status_str`'s exact output; `From<&AgentStatus>` correctly collapses `Suspended(_)`'s parameter (documented as deliberate — an enum can't express it). The 3-vocabulary reconciliation is correctly left as an ADR question, not attempted here.
- **`DecisionLabel`** (`routes/agents.rs`) — 5 variants matching `decision_label()`'s match arms 1:1, including the `unspecified` wildcard fallback.
- **`LogEventType`** (`routes/logs.rs`) — 22 variants, and the `From<AuditEventType>` impl has **no wildcard arm**. This is the property that matters most for this PR: it's compiler-enforced exhaustive, so a future `AuditEventType` variant added without a matching `LogEventType` update is a compile error, not a silent runtime gap. This is exactly the enum-completeness guarantee a closed-enum change needs.
- **`SimulateVerdict`** (`routes/policies.rs`) — 4 variants wired at each `simulate_policy` construction site.

All four serialize to the exact strings the handlers already emitted (lowercase for 3, PascalCase-verbatim for `LogEventType` matching `AuditEventType::as_str`), so the wire shape is unchanged — this is a type-level constraint, not a behaviour change. `AgentResponse.status` is correctly excluded (parameterised variants an enum can't express).

### Security / correctness — the subtle risk for this class of change
A closed enum that's missing a variant the server can still emit would break every real client that receives it — checked, and not the case here: each enum is derived directly from, or exhaustively matched against, the actual emitting code.

### CI — exhaustively green
All 32 checks pass, including the two most relevant to a codegen PR: **`OpenAPI spec drift check`** and **`Dashboard codegen drift check`** — confirming `openapi/v1.yaml` and `schema.d.ts` match the derives with no stray drift. Also `Rust conformance suite`, `Clippy lint`, `Test`, `SonarCloud`, `CodeQL`, `aa-cli build compat`, `codecov/patch`. `CI Success` = pass.

Not merged.

PR #1774, opened 2026-07-28
⬆ Bump monaco-editor from 0.55.1 to 0.56.0 in /dashboard (rebased)

Chisanan232:
Rebases dependabot #1762 onto current main after Cargo.lock/dashboard churn from other merged dependency PRs left it conflicting.

Verified locally: `tsc --noEmit` clean, `eslint` clean, full test suite (3016 tests) passing, `pnpm run build` succeeds (monaco chunks bundle correctly).

Closes the need for #1762.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -30,7 +30,7 @@
     "@types/d3-hierarchy": "^3.1.7",
     "d3-force": "^3.0.0",
     "d3-hierarchy": "^3.1.2",
-    "monaco-editor": "0.55.1",
+    "monaco-editor": "0.56.0",
     "openapi-fetch": "^0.17.0",
     "react": "^19.2.7",
     "react-dom": "^19.2.5",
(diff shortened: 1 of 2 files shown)


Chisanan232:
Closing — monaco-editor 0.56.0 changed its worker-loading strategy to spawn workers via `blob:` URLs, which the dashboard's strict CSP (`script-src 'self'`, no `worker-src` set so it falls back to `script-src`) blocks:

`Creating a worker from 'blob:http://localhost:4173/...' violates the following Content Security Policy directive: "script-src 'self'"`

This breaks Playwright e2e tests on unrelated pages (Alerts) because Monaco appears to load eagerly somewhere shared, tripping the CSP violation regardless of whether the code editor is visible. Confirmed via local reproduction of the e2e failure — not a fixable dependency-bump issue; needs either a `worker-src blob:` CSP relaxation or a MonacoEnvironment override to avoid blob workers, both of which are application-level security tradeoffs that should get a deliberate review rather than ride in on a routine version bump. Re-open once that's decided.

PR #1788, opened 2026-07-28
[AAASM-5143] ✨ (policies): Add history toggle for archived policy versions

Chisanan232:
## Description

Adds a `history` toggle to the Policies page header. When on, it re-runs
`usePoliciesQuery` with `include_archived=true` (`GET /api/v1/policies?include_archived`)
so older, inactive policy versions come back, and it visually marks those
archived rows with an uppercase `archived` tag and a dimmed row. Matches the
`design/v1/hi-fi/policy.jsx` header (`+ new policy`, `history`, `▸ Simulate`).

The `include_archived` flag is now part of the query key so the active-only and
history views cache independently; `useCreatePolicy`'s optimistic cache ops are
retargeted at the exact default-view key.

Archived vs proposed: `PolicyResponse` carries no `archived`/`status` field on
the wire, so the marker is derived honestly — a row is shown as archived only
when history mode is on and the version is inactive (`active === false`), i.e.
one of the older versions the endpoint surfaces solely under `include_archived`.
No archived data is fabricated; it comes from the API.

## Type of Change

- [x] ✨ New feature

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-5143

## Testing

- [x] No integration tests required

New tests (`features/policies/api.test.tsx`, `pages/PoliciesPage.test.tsx`):
- default query sends `include_archived=false`; toggle sends `true`
- clicking the header toggle re-runs the query with `{ includeArchived: true }`
- the archived marker appears on inactive rows only once history is on

Local gate (from `dashboard/`): `tsc --noEmit` clean, `eslint .` clean,
`vitest run` — 3078 passed. Browser verification was not possible without a
live gateway (the page is admin-scope gated and needs a backend for policy
data); behaviour is covered by the rendered-DOM component tests above.

## Checklist

### DIFF
--- a/dashboard/src/features/policies/api.test.tsx
+++ b/dashboard/src/features/policies/api.test.tsx
@@ -32,6 +32,11 @@ const POLICY: Policy = {
   policy_yaml: 'name: baseline\n',
 }
 
+// The policies list is keyed by its includeArchived flag (AAASM-5143); the
+// default (history-off) view — where a newly created policy lands — carries
+// this exact key.
+const DEFAULT_POLICIES_KEY = ['policies', { includeArchived: false }] as const
+
 let get: Mock
 let post: Mock
 
@@ -124,13 +129,13 @@ describe('useCreatePolicy', () => {
       () => new Promise<FetchResult>((resolve) => { resolvePost = resolve }),
     )
     const { client, wrapper } = makeWrapper()
-    client.setQueryData<Policy[]>(['policies'], [POLICY])
+    client.setQueryData<Policy[]>(DEFAULT_POLICIES_KEY, [POLICY])
     const { result } = renderHook(() => useCreatePolicy(), { wrapper })
 
     result.current.mutate({ policy_yaml: 'name: "my-new-policy"\n' })
 
     await waitFor(() => {
-      const cached = client.getQueryData<Policy[]>(['policies'])
+      const cached = client.getQueryData<Policy[]>(DEFAULT_POLICIES_KEY)
       expect(cached?.some((p) => p.name === 'my-new-policy' && p.version === 'pending')).toBe(true)
     })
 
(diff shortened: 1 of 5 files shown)


Chisanan232:
🐛 Pushed `2bf5f273`: the e2e failure (`aaasm-5021-shell-chrome.spec.ts:110` nav-badge-policy) was caused by the default query sending `?include_archived=false`, which changed the request URL and slipped past the ~10 e2e route mocks + the AppShell nav badge that all target the bare `**/api/v1/policies`. Fix: omit the param when `false` (the endpoint already defaults to active-only), so only history mode appends `?include_archived=true`. The previously-failing spec now passes 2/2 locally against the built dist. Propagated the fix forward through the stacked 5152/5153 branches. 🤖 Claude Code

Chisanan232 (commented):

**Requirement (AAASM-5143)** — the Policies page header gains a `history` toggle that re-runs `usePoliciesQuery` with `include_archived: true` and marks archived rows. Archived state is derived honestly (the wire `PolicyResponse` has no `archived` field — inactive rows are tagged only in history mode, no fabricated data), and the query key carries the flag so the two views cache independently. Matches `design/v1/hi-fi/policy.jsx:67`.

**CI** — fully green (17 success, 23 skipped, 0 failing) after fixing the e2e regression (see below).

**Fix applied during review** — the original default query sent `?include_archived=false`, which changed the request URL and slipped past the AppShell nav badge + ~10 e2e route mocks that target the bare `**/api/v1/policies`, breaking `aaasm-5021-shell-chrome.spec.ts`. Corrected to omit the param when `false` (the endpoint already defaults to active-only), so only history mode appends `?include_archived=true`. Verified the failing spec passes 2/2 locally against the built dist.

**Security** — read-only query-param addition; no injection, no new auth surface (the admin-scope gating on `GET /policies` is unchanged).

PR #1796, opened 2026-07-29
[AAASM-5175] ✨ (aa-api): Evaluate capability over-permission

Chisanan232:
## Description

Implements a real over-permission evaluation for the capability matrix (AAASM-5175). Previously `CapabilityAgent.flagged` and `CapCell.flag` were hardcoded `None` at every construction site, so the dashboard's **danger-toned "flagged agents" tile was a permanent zero that read as a measured all-clear** on the product's flagship governance surface — with nothing actually evaluating over-permissioning.

**Scope check (done first, per the ticket):** confirmed over-permission is **distinct** from the trust score (ADR 0019 / AAASM-5083 — a behavioural audit-derived number) and from the topology `flagged` field (violation-volume ≥ 50). It is a separate *structural* signal, so this is not a duplicate of 5083 — documented in ADR 0029 §Scope.

**The rule (ADR 0029, Status: Proposed):** an agent is over-permissioned when its effective merged capability set grants a **destructive system verb its declared `RiskTier` baseline does not warrant**. Computed **entirely from data the projection already loads** — the merged capability cascade + `AgentRecord.risk_tier` — touching no audit log, no time window, no enforcement path (the alternative "unused-grant" signal was explicitly rejected to avoid dragging the audit log's durability/IDOR/truncation caveats into a pure projection).

**Truthful absence:** `flagged`/`flag`/`note` are emitted **only** for a positive finding. Absent when the agent isn't evaluated — no resolvable risk tier, or an empty cascade (where every cell is Allow by fall-through and flagging would be a false positive). Never a fabricated `Some(false)`-as-zero.

**Commits (9 atomic):** ADR → derivation rule → `CapCell.flag` → `CapabilityAgent.flagged`+note → end-to-end test → field docs → fmt → openapi regen → schema regen.

## Type of Change
- [x] ✨ New feature

## Breaking Changes
- [x] No — `flagged`/`flag`/`note` are optional and were already in the contract (previously always absent); the wire shape is unchanged, only now populated.

## Related Issues
- Related Jira ticket: AAASM-5175
- Depends conceptually on: ADR 0029 (added here)

## Testing
- `cargo test -p aa-api` — green, incl. a new end-to-end projection test (flagged when a low-tier agent holds a destructive grant; absent when no resolvable tier / empty cascade).
- `cargo clippy` + `cargo fmt --check` — clean.
- OpenAPI drift: `generate_openapi` output + regenerated `schema.d.ts` committed.

## Security / truthfulness
- The rule is written down in ADR 0029 (reviewable, signed-off) rather than invented in the handler — the discipline ADR 0019 applied to trust.
- No false positives: absence is preserved wherever the evaluation can't run, so the danger tile only lights on a real, explained finding.

### DIFF
--- a/aa-api/src/models/capability.rs
+++ b/aa-api/src/models/capability.rs
@@ -76,7 +76,12 @@ pub struct CapCell {
     pub write: Decision,
     pub delete: Decision,
     pub exec: Decision,
-    /// Marks this cell for UI attention (e.g. over-permissioned).
+    /// `Some(true)` when this cell's grant is over-permission — the agent is
+    /// effectively allowed a destructive system verb its declared `RiskTier`
+    /// baseline does not warrant (AAASM-5175, ADR 0029). Absent when the cell is
+    /// within baseline, or when the agent is not evaluated (no resolvable tier,
+    /// or an empty cascade). Only the offending marker is emitted; a per-cell
+    /// `false` is not, so the UI highlights positives without negative clutter.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub flag: Option<bool>,
 }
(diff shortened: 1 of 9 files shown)


Chisanan232 (commented):

**Requirement (AAASM-5175)** — real over-permission evaluation now populates `CapabilityAgent.flagged` (+ explaining `note`) and per-cell `CapCell.flag`, replacing the permanent-zero danger tile that read as a measured all-clear.

**Reviewed the rule (ADR 0029):** an agent is over-permissioned when it is effectively granted a destructive **system** capability its declared `RiskTier` baseline doesn't warrant. Verified `tier_permits`: Low permits no destructive system cap; Medium permits FileWrite/NetworkOutbound; High/Critical permit all. `HIGH_PRIVILEGE` correctly excludes FileRead (non-destructive) and inert/unclassified caps. Computed purely from the projection's existing inputs (merged cascade + risk tier) — no audit-log dependency.

**Truthfulness — the key property:** fail-absent. An agent with no resolvable tier is not evaluated (`flagged`/`flag` stay absent); the only `false` emitted is for a genuinely-evaluated-and-clean agent. Flags are gated on effective `Decision::Allow`, so a denied-but-over-baseline capability is never flagged. No fabricated `true`.

**Scope check** — confirmed distinct from ADR 0019 trust (behavioural) and the topology violation-volume flag; documented in ADR 0029 §Scope, so not a duplicate of 5083.

**CI** — fully green (incl. the end-to-end projection test). Rule written down in ADR 0029 (Proposed), not invented in the handler.

PR #1731, opened 2026-07-27
[AAASM-5191] 🐛 (dashboard): Revive the dead e2e suite and gate it in CI

Chisanan232:
## Description

Revives the dashboard Playwright suite (AAASM-5191) and puts it behind a CI merge gate (AAASM-5192).

**AAASM-5191 — the seed.** `src/auth/tokenStorage.ts` has read `aa_token` from `sessionStorage` only since AAASM-4322 (`d68a0d63`, 2026-07-08). 29 spec files still seeded `localStorage`, so the app never authenticated and every one of them timed out before mount — with no hint that auth was the cause. `verify-aaasm-5024` and `verify-aaasm-5063` already seeded both stores and are untouched; the theme key (`aa-dashboard-theme`) legitimately lives in `localStorage` and is untouched.

**AAASM-5192 — the gate.** No workflow referenced Playwright at all, which is why the breakage sat undetected for 19 days. Adds `dashboard-e2e` behind the same `dorny/paths-filter` `dashboard` router the other dashboard jobs use (Rust-only PRs unaffected), publishing the HTML report, failure screenshots, retry traces and JUnit XML on both outcomes.

### ⚠️ The one-line fix was necessary but not sufficient — please read

The ticket states the fix is "one line" with "no further rot behind the first failure". **That held only for the single `theme-visual` rail test the reviewer sampled.** Measured at suite scale:

| | tests | passing | failing |
|---|---|---|---|
| Affected 33 files — before | 177 | 12 | 163 |
| Affected 33 files — after the seed fix | 177 | **86** | 89 |
| **Full suite — after the seed fix** | **433** | **300** | **131** |

The seed fix is a large, real win (12 → 86 on the affected set). But **131 tests across 41 of 86 files still fail**, and **16 of those 41 were never touched by the seed bug at all** — they already used `sessionStorage`. The suite's rot substantially predates and exceeds AAASM-5191.

Diagnosed causes (filed as **AAASM-5195**):
1. **Pagination-envelope drift (AAASM-4892)** — `/agents`, `/logs`, `/policies`, `/approvals`, `/alerts` and the IAM lists now return `{ items, total }`; the specs' route mocks still fulfil a bare array, so `data?.items` is `undefined` and every list renders zero rows. Verified on `fleet.spec.ts`.
[...]

PR #1817, opened 2026-07-30
[AAASM-5293] 📝 (adr): Move development-process content out of the ADR corpus

Chisanan232:
## Description

Brings the ADR corpus into line with the ADR scope rule: **an ADR records only durable product or system decisions.** Both bodies of content touched here were removed from the ADR corpus because **they documented workflow policy rather than a product or architecture decision** — how to author a CI path filter, and how to execute a one-time per-repo branch migration. Neither touches product semantics, user-visible behaviour, a security boundary, or a public contract.

This follows the precedent already set in #1731, where the original ADR-0028 ("The Dashboard E2E Suite Is a Merge Gate") was removed before merge for the same reason and its content migrated to `dashboard/tests/e2e/README.md` + `CONTRIBUTING.md`. As there, **content is relocated, never discarded.**

Both tickets are in one PR because they touch the same two index files (`docs/src/adr/README.md`, `docs/src/SUMMARY.md`) and would otherwise race.

### AAASM-5293 — ADR-0028 retired

All four of its numbered decisions were rules about authoring a `dorny/paths-filter` / `on.paths` block. The ADR conceded the point itself ("This ADR does not change any workflow file"; it "ratifies the rule those fixes already implement"), and its own rule 4 named its correct home: `<org>/.github/workflow-templates/`.

- **Rules 1–3 → `CONTRIBUTING.md`**, extending the existing `### Which CI jobs block a merge` subsection. That subsection already ended on the caveat that a job behind a paths-filter router is only as sound as its filter, so the rule reads as a continuation of that thought rather than a parallel section.
- **Rule 4 → AAASM-5257.** `<org>/.github/workflow-templates/` does not exist in either org, so the requirement is parked on the ticket rather than given an invented location.
- **The org-wide audit inventory → AAASM-5257** ([comment](https://lightning-dust-mite.atlassian.net/browse/AAASM-5257)): the named non-compliant repos (`go-sdk`, `node-sdk`, `python-sdk`, `examples`, `homebrew-tap`, `arena`), the two-failure-mode framing (over-triggering via a bare `dir/**` vs. no filter at all — the `horonomy` mode), and the `horonomy` findings. Deliberately **not** put in `CONTRIBUTING.md`, which should carry the rule, not the backlog.
- **No replacement ADR under any number.** 0028 stays retired, and the index now records that ADR numbers are permanent so the gap is not re-filled.
- This also resolves the ADR's status defect — it was `Proposed` while the eight CI fixes it ratified were already shipped and merged.

### AAASM-5294 — ADR-0016 split, not retired

Mixed: one durable decision plus a per-repo migration runbook. **The file stays at number 0016** with a reduced Decision section.

**Kept** (folded into one coherent decision — these constrain future authors rather than describing runbook steps):
- D1 — `main` is the canonical org-wide default branch.
- From D6 — `master` is a GitHub-managed redirect, not a retained branch; do not recreate it; the redirect does **not** cover `raw.githubusercontent.com/…/master`, `git fetch master`, or `uses: …@master`, so cross-repo links must use the `HEAD` form.
- From D3 — a consumer workflow's PR `base:` must track the target repo's current default branch.

**Moved** to a private runbook: D2, D4, D5, D7, D8 in full; D3's migration-window mechanics; D6's time-bounded compatibility-branch approval procedure; and the `## Operational guidance` / `## Validation requirements` sections.

**Destination: `internal-docs` (private), not a public `docs/src/` page.** Reasons: `internal-docs/docs/runbooks/` already exists and already owns cross-repo operational procedure (`cross-repo-release.md`); the runbook enumerates private repos (`cloud`, `agent-assembly-enterprise`, `e2e-private`, `.github-private`) and `cloud`'s deploy-pipeline coupling, which `.claude/rules/05-context-boundary.md` says not to place in a public artifact; and its audience is the org's release/infra owners, not this repo's contributors or users. Publishing a one-time, per-repo procedure on the product doc site would also leave a permanently-stale public page. Note this **remediates a pre-existing disclosure rather than avoiding a new one**: those private repo names (`cloud`, `agent-assembly-enterprise`, `e2e-private`, `.github-private`) and `cloud`'s deploy-pipeline coupling were **already published** on the public docs site by ADR-0016 before this split. Moving the runbook private and compressing the ADR's Context takes them back off it.

→ **internal-docs PR: ai-agent-assembly/internal-docs#93** (must merge for the moved content to have a home).

**Convention lines.** ADR-0016's Operational guidance directs these to `CONTRIBUTING.md`, `.claude/CLAUDE.md` and `.claude/rules/02-git-workflow.md`. Three were already correct and are left untouched: `CONTRIBUTING.md` already says open a PR against `main`; the repo `.claude/CLAUDE.md` already says base branch `main`; `.claude/rules/02-git-workflow.md` (in `.github`) is already branch-name-agnostic. What had no contributor-facing home was the `HEAD`-form rule itself, added here as a short `### Linking to another repository` subsection.

Two extras, called out rather than slipped in:
- The **workspace-level `CLAUDE.md`** still said "Base branch: always `master`" — corrected to `main` (the one status defect AAASM-5294 asks for). That file is not inside any git repo, so it is **not part of this diff**.
- `CONTRIBUTING.md`'s own `.github/blob/master` link is flipped to the `HEAD` form. ADR-0016 classifies `github.com` web links as redirect-covered and safe to sweep later, but stating the rule in the same file as a violation of it was not worth the deferral. The operative criterion is **same-file *and* in-org** — stated explicitly because "two paragraphs above" would have implied same-file alone, which would also catch `CONTRIBUTING.md:11`'s third-party lefthook link 200+ lines away, and "in-org" alone would pull in `examples/…/lib.rs`. The new rule's wording is now scoped to org repos to match, so it no longer reads as covering third-party repositories whose branch names are not ours to track. Other deferred in-org `blob/master` instances elsewhere in the tree are untouched.

### Merge constraint — both PRs must land together

**#1817 and ai-agent-assembly/internal-docs#93 must merge together.** Merging #93 alone duplicates the migration procedure in two places; merging #1817 alone dangles the pointer at L16–17 of ADR-0016, which tells the reader the procedure lives in a runbook that would not exist yet.

### Review findings addressed

The first pass migrated every *mechanical* element faithfully — commands, paths, URL forms, config keys, workflow filenames, all 18 repo names, numbers and exception conditions — but dropped four **normative/meta** fragments, the class a sentence-similarity score grades as "matched". All four are restored:

1. **A silently substituted reconsideration trigger.** ADR-0016's *"discovery of a reference class this checklist misses"* had been replaced by a narrower redirect-coverage trigger and landed in neither destination. It is the checklist's own self-revision hook, and it covers precisely the failure mode the pilot proved — a consumer's hardcoded `base:` that nobody had enumerated. Restored **in the runbook, beside the checklist it guards**. The redirect-coverage trigger stays in the ADR: they are two different triggers, not a substitution.
2. **`release.yml` SDK base state.** The runbook kept `(already main)` for homebrew-tap but dropped `(still master)` and `currently` for the three SDKs — the actionable half, without which an operator cannot tell which of the four bases are outstanding. Restored.
3. **A dropped consequence.** *"Future contributors: one reviewed procedure + a drift guard, not tribal knowledge"* was absent from both destinations. Restored to the runbook's "Why this exists", which is the sentence's natural home.
4. **A downgraded obligation.** `### A repo migration MUST audit both directions` had become `### Audit both directions` — body verbatim, obligation gone — leaving the reduced ADR with zero uppercase `MUST`. **Both halves fixed, and here is the choice:** `MUST` is restored on the runbook heading, *and* the ADR's retained D3 rule is upgraded to `a consumer's PR base: **MUST** name the target repo's current default branch`. Rather than re-import the one-time audit procedure into the ADR — which would undo the point of this split — the ADR now carries the pilot's lesson in its durable form: a permanent obligation on whoever writes the cross-repo reference, which is where the pilot's break actually lived.

A fifth item went to **[AAASM-5257](https://lightning-dust-mite.atlassian.net/browse/AAASM-5257)**: ADR-0028's follow-up for *a lint/meta-workflow that rejects a bare `dir/**` filter*. That is the one item that would make the rule self-enforcing rather than reliant on reviewer attention, and it was missing from both `CONTRIBUTING.md` and the first inventory comment.

### Known contradiction a contributor may hit — [AAASM-5297](https://lightning-dust-mite.atlassian.net/browse/AAASM-5297)

`.ci/check-metadata-drift.sh:44-46` rejects `blob/main` and mandates `blob/master`, citing `governance.baseline_doc_base` as its authority — but the registry says the opposite (`metadata/org-profile.yaml` has `default_branch: main` and `baseline_doc_base: ".../blob/main"`, `.github`'s live default is `main`, `blob/main` → 200 while `blob/master` → 302). The gate currently passes only because existing links use `blob/HEAD`, which evades its pattern.

**This is pre-existing, belongs to another repo's contract, and is deliberately not fixed here** — it is filed as AAASM-5297 and linked to 5293/5294. It is flagged in this body so that anyone who follows the new `HEAD`-form rule and then hits a red metadata-drift gate has the explanation. Note the rule as written recommends `blob/HEAD`, which the gate does *not* reject; only a contributor writing `blob/main` would trip it.

## Type of Change

- [ ] ✨ New feature
- [ ] 🐛 Bug fix
- [x] ♻️ Refactoring
- [x] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Documentation-only. No crate, workflow, script or dashboard source is touched.

## Related Issues

- Related Jira ticket: [AAASM-5293](https://lightning-dust-mite.atlassian.net/browse/AAASM-5293) (retire ADR-0028), [AAASM-5294](https://lightning-dust-mite.atlassian.net/browse/AAASM-5294) (split ADR-0016)
- Epic: [AAASM-5290](https://lightning-dust-mite.atlassian.net/browse/AAASM-5290) — reclassify ADRs that record development process
- Verification checklist: [AAASM-5295](https://lightning-dust-mite.atlassian.net/browse/AAASM-5295)
- Inventory recorded on: [AAASM-5257](https://lightning-dust-mite.atlassian.net/browse/AAASM-5257)
- Precedent: #1731

## Testing

- [ ] Unit tests added / updated
- [x] No tests required (explain why)

Documentation-only, so the gates were **scoped rather than swept** — nothing here touches `dashboard/src` or any crate, so `cargo nextest` / `clippy` / `vitest` / Playwright have no changed input and were deliberately not run.

What was run:

| Gate | Result |
|---|---|
| `mdbook build docs` (the gate a stale `SUMMARY.md` entry breaks) | **Pass** in CI on this branch. It was also confirmed green on merge-base `aae32497` first so the pass is meaningful — but that baseline is **local evidence only**: CI *skipped* `Build mdBook` on `aae32497` under its path filter, so it cannot be corroborated from CI |
| Docusaurus `pnpm build` for the runbook's destination repo (`onBrokenLinks: 'throw'`) | **Pass** — **local evidence only.** That repo's CI is billing-blocked (`build-and-check` aborts in ~2s, 0 steps, no runner, with a payments annotation), and `link-check` never ran at all: it is a dependent of the blocked job and was skipped. Cloudflare's Workers build of the same content did pass, independently |
| `grep -rniE 'adr[ -]?0028'` across the tree | **0 hits** outside git history |
| Index/disk agreement | 27 files on disk = 27 `README.md` rows = 27 `SUMMARY.md` entries |
| Structural placement assertions — each moved fragment present at its destination, each retained fragment present in the ADR, and the moved procedure verifiably absent from the ADR | **All pass** |
| Every command, path, URL form, config key, workflow filename, repo name, number and exception condition in the moved text | Present at its destination |
| Normative/meta fragments (obligations, reconsideration triggers, consequences) | Audited **individually**; four were found dropped or silently substituted on first pass and are restored — see "Review findings addressed" below |
| `git status --porcelain` | Clean |

**Active ADR count is 27** — 0001–0029, less 0005 (never existed) and less retired 0028.

## Checklist

- [x] Code follows project style guidelines (`cargo fmt`, `cargo clippy`) — N/A, no Rust changed

> **Note on how this branch was published:** the pre-push hook runs `cargo doc --workspace --no-deps` with no path glob, which hangs on macOS even for a Markdown-only branch. Per this repo's documented convention the commits were published via the GitHub Git Data API rather than bypassing the hook — `--no-verify` was **not** used. The replayed tree SHA was verified byte-identical to the local tree at each step — at the current head `96cac4ac`, tree `b2711d14`.

[AAASM-5257]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232 (commented):
## Adversarial review — head `ad6013f8`

Reviewed as a non-author. I re-derived every claim from the files and the API rather than from the PR body. **Comment only — not an approval, and I have not merged.**

The reclassification itself is correct and the scope rule is applied honestly in both directions: ADR-0028 was procedure end-to-end and is rightly gone; ADR-0016 was genuinely mixed and the split is drawn at the right seam. What does not survive scrutiny is the **"nothing was lost"** claim — the sentence-level arithmetic scores paraphrases as matches and therefore misses three real deletions and one silent substitution. Details below.

---

### Verified — these all hold

| Claim | How I checked | Result |
|---|---|---|
| Tree parity across the Git-Data-API replay | `git log --format='%T'` on all 6 commits, local branch vs `remote/…`, pairwise | **Verified.** Every commit's tree SHA is identical; `git diff local remote` is empty; final tree is `6991182a` as stated. The replay is clean. |
| ADR-0028 file deleted, README row gone, SUMMARY entry gone | `git ls-tree` / `git show` on `ad6013f8` | Verified |
| `grep -rniE 'adr[ -]?0028'` → zero | `git grep -niE 'adr[ -]?0028' ad6013f8` and `'0028-ci-trigger'` | **Zero hits.** Base commit had exactly one (the file itself), so the sweep is complete |
| Active ADR count 27, disk == README == SUMMARY | Counted all three independently and compared the **number sets**, not just the totals | **27 / 27 / 27**, identical sets `{0001–0004, 0006–0027, 0029}`. No orphan TOC entry, no unlisted ADR |
| No replacement ADR under any number | No `0030`, no new file; README now records number permanence | Verified |
| Rules 1–3 as a *continuation* | The insertion sits inside `### Which CI jobs block a merge`, immediately after the "only as sound as its filter" caveat, before the next `##` | Verified — continuation, not a parallel section |
| `Build mdBook` passes | Check-run on `ad6013f8` | **success** |
| ADR-0016 still at 0016 with the 3 required survivals | Read the reduced file | D1 (L48–52, archived-repo exception intact); D6 fragment (L58–81: redirect-not-a-branch, do-not-recreate, all three non-covered forms, `HEAD` mandate); D3 fragment (L83–89, `check-release-completeness.sh` named) |
| Procedure absent from ADR / present in runbook | D2/D4/D5/D7/D8, D3 mechanics, D6 compat procedure, `## Operational guidance`, `## Validation requirements` | Verified at both ends |
| Context-boundary justification | Grepped the moved text for the four private names | **Holds** — runbook L31/33/95/136–137 name `cloud`, `agent-assembly-enterprise`, `e2e-private`, `.github-private` and `cloud`'s deploy coupling. Reduced public ADR: **zero** private names |
| AAASM-5257 inventory | Fetched comment 19872 and diffed it against the deleted ADR | **Complete** — all six repos *with* their workflow filenames, the two-failure-mode framing, all three horonomy findings, both counter-examples with the exact `paths-ignore` value, the four already-merged fixes with ticket/PR numbers, the reconsideration triggers, and rule 4 verbatim |

**On the base-commit reasoning for `Build mdBook`:** CI cannot corroborate it — `Build mdBook` was **`skipped`** on the merge-base `aae32497` (path filter). So I reproduced it: `git archive` of both trees, `mdbook build docs` on each. **Base exit 0, head exit 0.** The reasoning is sound and the head's pass is meaningful. Worth stating in the PR body that the base evidence is local-only, since a reader checking CI will find a skip.

**ADR-0016 coherence — it reads as one decision, not three staples.** L54–56 explicitly frames the two kept fragments as *"Two consequences of that choice [that] are themselves standing constraints, binding on anyone writing a cross-repo link… not merely steps in the one-time migration."* That is the right move: it converts two migration steps into derived standing contracts, which is exactly what makes them ADR-shaped. Good call, and better than a bare concatenation would have been.

---

### MAJOR — a live contradiction this PR sharpens and does not flag

`.ci/check-metadata-drift.sh:44–46` currently enforces the **opposite** of the rule this PR introduces:

```
check "'.github' governance link uses 'blob/main' (canonical branch is 'master')" \
  'ai-agent-assembly/\.github/blob/main' \
  "use .../.github/blob/master/... (registry governance.baseline_doc_base)"
```

So: the registry-owned canonical form for the `.github` governance link is `blob/master`, and writing `blob/main` **fails CI** with a message asserting *"canonical branch is `master`"*.

Meanwhile this PR adds `CONTRIBUTING.md` → `## Linking to another repository`, mandating the `HEAD` form, citing ADR-0016 — whose D1 says `main` is canonical org-wide. And `ai-agent-assembly/.github`'s **actual** default branch is already `main` (I checked the API; `blob/HEAD` → 200, `blob/master` → 302). The gate is stale.

Why this matters here rather than as generic pre-existing debt:

1. This PR touches **that exact link** (the `.github/blob/master` → `blob/HEAD` flip) as a self-declared scope extra.
2. The gate passed **only because `blob/HEAD` evades a pattern that greps for `blob/main` specifically**. That is a near-miss, not a clean pass.
3. The reduced ADR-0016 cross-links ADR-0014 at L131 as *"Related — `.github` registry/org-profile inbound refs"* — so the collision was one hop from the author's own traceability table.
4. Contributor-facing consequence: someone who reads the new section, follows D1, and writes `blob/main` gets a red gate telling them `master` is canonical.

The stale gate is **PRE-EXISTING-OUT-OF-SCOPE** to fix. But shipping an org-wide `HEAD`-form mandate while leaving an unflagged gate that enforces `master` for the very repo the mandate's example points at is a gap in *this* change. Minimum ask: a note in the PR body plus a follow-up ticket against `check-metadata-drift.sh` / `governance.baseline_doc_base`. This may be **ARCHITECTURE-DECISION-REQUIRED** if the registry's `master` value is deliberate rather than stale — worth confirming which.

### MAJOR — the "nothing was lost" accounting misses a silent substitution

`ADR-0016` old L220–221, Reconsideration triggers:

> a new cross-repo automation that pins a branch ref; **discovery of a reference class this checklist misses**.

The reduced ADR L119–121 keeps the first two triggers verbatim and **replaces** the third with a different, narrower one (*"GitHub changing what its rename redirect covers"*). The original is in **neither** destination — `grep -niE 'reference class|checklist misses'` returns nothing in the ADR or the runbook, and the runbook has no Reconsideration-triggers section at all.

This is the trigger that covers **the exact failure mode the pilot proved** — the break was in a reference class nobody had enumerated (a consumer workflow's hardcoded `base:`). Removing it leaves the L89–108 reference-class checklist with no self-revision trigger: it can now silently go stale, which is the failure it exists to prevent. A substitution presented as a clean migration is the one thing a sentence-count cannot catch, which is precisely why the 130/133 figure should not carry the weight the PR body puts on it.

Not a blocker to the reclassification — restore the trigger (either destination) and the accounting is honest.

### MINOR — three further losses at both destinations

1. **Dropped consequence.** Old L195–196: *"**Future contributors**: one reviewed procedure + a drift guard, not tribal knowledge."* Absent from both files (`grep tribal|Future contributors` → 0). The reduced ADR substitutes a different audience bullet; the runbook has no Consequences section. Arguably self-framing — but it is a stakeholder-impact statement, and that audience is now unrepresented anywhere.
2. **Dropped fact.** Old L91–92: *"`release.yml` **currently** hardcodes `base:` for homebrew-tap (already `main`) + the three SDKs **(still `master`)**."* The runbook (L86–87) keeps *"(already `main`)"* for the tap but drops *"(still `master`)"* and *"currently"* for the three SDKs. An operator reading the runbook can no longer tell which of the four hardcoded bases are outstanding — in a runbook, that is the actionable half of the sentence.
3. **Obligation grade softened.** Old L72: `### 2. A repo migration **MUST** audit both directions` → runbook L67: `### 1. Audit both directions`. Body text is verbatim, and as step 1 of a numbered procedure the obligation substantially survives — but the reduced ADR contains **zero** occurrences of `MUST` and no both-directions requirement in any form (only a passing mention in the "procedure is not here" callout). The pilot's central lesson now has no public normative anchor. Correct per the scope rule (auditing *is* procedure); flagging because your accounting scored it as a clean match when it is a grade change.

### MINOR — the new link rule is violated in its own file

`## Linking to another repository` is worded generically (*"When you write a cross-repo link, ref, or automation target…"*). Still `blob/master` after this PR:

- `CONTRIBUTING.md:11` — `evilmartians/lefthook/blob/master/docs/install.md`, **213 lines above the new rule, same file**
- `examples/aa-devtool-sample-myeditor/src/lib.rs:12` — `ai-agent-assembly/agent-assembly/blob/master/…`, an **in-org self-reference**
- `docs/src/architecture/building.md:14` — lefthook again

**On the scope call you asked to be judged:** it is **not** unannounced scope creep — you disclosed it in both the commit message and the PR body, it is one line, and it is directly germane to the rule being introduced. That is a defensible minimum and I would not ask you to expand it. The issue is that the stated criterion doesn't cleanly predict the outcome: "two paragraphs above a violation" implies *same file*, which would also catch L11; "in-org" would instead catch `lib.rs`. The operative rule seems to be *same-file **and** in-org* — a fine criterion, just worth saying so, or scoping the rule's wording to org repos so a third-party `blob/master` isn't a standing counterexample in the same document.

### MINOR — public ADR points into a private repo

Reduced ADR L16–17 sends the reader to *"the internal `internal-docs` runbook `docs/runbooks/default-branch-migration.md`."* `docs/src/` is the **public** book, so public readers get a decision whose execution procedure is unreachable, and the pointer names a private repo plus an exact internal path — mild tension with the same `05-context-boundary` rule invoked to justify the move. Not a leak (a pointer, not content) and I think the destination call is right; consider softening to "an internal operations runbook" without the path, or noting explicitly that the procedure is internal by design.

### MINOR — one ADR-0028 item landed nowhere

The ADR's Validation-requirements follow-up — *"a lint/CI check that enforces the rule automatically (e.g. a meta-workflow that rejects a bare `dir/**` filter)"* — is in **neither** `CONTRIBUTING.md` nor comment 19872 (`grep` for `meta-workflow|lint|enforce` on the comment → all absent). Rules 1–2 are now advisory prose with the one item that would have made them self-enforcing dropped. Worth a line on 5257 next to rule 4, which is blocked on the same missing `workflow-templates/` location.

### NIT — D1–D8 labels removed

The reduced ADR de-numbers the Decision headings, so an external citation of "ADR-0016 §3" would dangle. I checked: nothing cites it that way. Across `agent-assembly`, `internal-docs`, `docs`, `e2e-public`, `examples`, `python-sdk`, `node-sdk`, `go-sdk`, every reference is by ADR number only (`examples/.claude/CLAUDE.md:154`, `examples/.claude/skills/sdk-version-bump/SKILL.md:138`). No live breakage — noting only so a future renumber isn't assumed free.

### PRE-EXISTING-OUT-OF-SCOPE — the disclosure was already public

`cloud`, `agent-assembly-enterprise`, `e2e-private` and `.github-private` were named on the **public** docs site before this PR (old ADR L20, L22, L100, L162–163). So the move does not merely avoid a new context-boundary violation, it **remediates an existing one** — your justification is stronger than you claimed, and the "side benefit" framing undersells it. Not remediable in git history of a public repo; flagging as informational.

---

### On the skipped gates

**Agreed, and I'd defend it.** Zero crate files, zero `dashboard/src` files, zero workflow files in the diff — `cargo nextest`, `clippy`, `vitest` and Playwright have literally no changed input, and running them would produce evidence about nothing. Running `mdbook build docs` (the gate a stale `SUMMARY.md` entry actually breaks) plus the Docusaurus build with `onBrokenLinks: 'throw'` is the correctly-chosen pair. A defended skip beats a manufactured sweep. The one improvement is the base-commit note above: say that the base-green evidence is local, since CI shows a skip there.

### Verdict

**Do not merge yet — comment only, no approval.** The reclassification is right, the split seam is right, the destination call is right and independently verified, the index/count bookkeeping is exact, and the API replay is byte-clean. Two things to settle first:

1. **The `check-metadata-drift.sh` / ADR-0014 contradiction** (MAJOR) — flag it and ticket it, or confirm the registry's `master` is deliberate.
2. **Restore the dropped reconsideration trigger** (MAJOR) and, ideally, the runbook's *"(still `master`)"* fact (MINOR).

Everything else is MINOR/NIT and can ride a follow-up. And please retire the 130/133 framing — it is a similarity metric, not a completeness proof, and here it scored a silent substitution as a match.

Chisanan232:
### Review findings addressed — new head `96cac4ac`

Thanks — the four losses were real, and all in the class a similarity score cannot see. Fixed, plus the framing corrections.

**MAJOR — the silent substitution.** `"discovery of a reference class this checklist misses"` is restored, **in the runbook beside the checklist it guards** (your recommendation, which I agree with). The redirect-coverage trigger stays in the ADR — two triggers, not a replacement. It reads as the checklist's explicit self-revision hook and names the pilot's failure mode as the reason it exists.

**MINOR 1** — `release.yml` now reads `**currently** hardcodes base: for homebrew-tap (already main) + the three SDKs (still master)`.

**MINOR 2** — *"future contributors get one reviewed procedure plus a drift guard, not tribal knowledge"* restored to the runbook's "Why this exists".

**MINOR 3 — I did both, and here is which and why.** `MUST` is restored on the runbook heading (`### 1. A repo migration MUST audit both directions`), **and** the ADR's retained D3 is upgraded to `a consumer's PR base: **MUST** name that repo's current default branch`. I did not state "audit both directions" in the ADR, because that is a one-time migration step and re-importing it would undo this split. Instead the ADR now carries the pilot's lesson in its *durable* form — a permanent obligation on whoever writes the cross-repo reference, which is where the pilot's break actually lived. So the public record has a normative anchor and the procedure stays in the runbook.

**Self-enforcement** — ADR-0028's *"lint/meta-workflow that rejects a bare `dir/**` filter"* follow-up added to AAASM-5257, with a note that any such check must implement the rule-3 content-repo exception or it will fire on `internal-docs`, and that it is coupled to rule 4's non-existent shared template location.

**AAASM-5297** — flagged in the body, not fixed here. Worth noting the new rule recommends `blob/HEAD`, which the gate does *not* reject, so a contributor following it will not hit red CI; only `blob/main` trips it.

**Framing corrections.**
- The 130/133 metric is **retired** from the body — you are right that it is a similarity score, not a completeness proof, and it graded a substitution as a match. Replaced with what was actually verified structurally, plus an explicit line saying the normative/meta class was audited individually and four items were found dropped.
- The mdBook base-commit baseline is now labelled **local evidence only**, noting CI skipped `Build mdBook` on `aae32497` under its path filter.
- Same for #93: `pnpm build` is local + Cloudflare only, and `link-check` **never ran** in CI (skipped as a dependent of the billing-blocked job).
- Private destination restated as **remediating a pre-existing disclosure** — those names were already on the public docs site before the split.
- Scope call: the operative criterion is stated as **same-file *and* in-org**, and I went further and scoped the new rule's wording to org repos, so it no longer implies third-party branch names are ours to track.

**Gates re-run:** `mdbook build docs` pass; Docusaurus `pnpm build` with `onBrokenLinks: 'throw'` pass. Everything else skipped — no crate or `dashboard/src` input changed. Git Data API replay again (pre-push `cargo doc` hangs on macOS); no `--no-verify`, no force-push; final tree `b2711d14` verified identical to local.

**Both PRs must land together** — noted in both bodies. Not approving, not merging.

Chisanan232 (commented):
## Follow-up review — head `96cac4ac` (delta from `ad6013f8`)

Re-reviewed the delta only, as a non-author. **Comment only — not an approval, and I have not merged.**

Two files, +8/−2. Everything I raised is addressed. Verified independently:

| Check | Result |
|---|---|
| Tree parity across the replay | **Verified** — local tip `0fc76305` and remote `96cac4ac` both carry tree **`b2711d14`**; `git diff` between them is empty |
| All checks | 7 success + 1 skipped (`Deploy to GitHub Pages`). **`Metadata-drift gate` success**, `Build mdBook` success |
| Invariants | `adr[ -]?0028` → **0 hits**; disk **27** / README **27** / SUMMARY **27**; `0016` present; **zero** new ADR files added vs `ad6013f8` |
| `MUST` in the reduced ADR | was 0, now **1** — the D3 upgrade |
| AAASM-5257 | comment 19881 adds the `dir/**` lint follow-up, correctly noting it must implement rule 3's content-repo exception or it fires on `internal-docs`, and that it is coupled to rule 4's non-existent template location |
| PR body | `130/133` **gone** (0 occurrences), replaced with the structural list plus an explicit statement that the normative/meta class was audited individually and four items were found dropped. mdBook baseline labelled local-only *with* the note that CI skipped it on `aae32497`. Private destination restated as remediating a pre-existing disclosure. Merge-together constraint stated |

**AAASM-5297 — verified, and the added point is correct and materially de-escalating.** The gate's pattern is `ai-agent-assembly/\.github/blob/main`. `blob/HEAD` — the form the new rule actually recommends — does not match it, so a contributor who follows the rule as written will not hit red CI; only someone writing `blob/main` trips it. `Metadata-drift gate` passes on this head. That converts my MAJOR from "the new rule leads contributors into a red gate" to "a latent stale assertion in another repo's registry contract." Correctly filed, correctly not fixed here.

**Scope rewording — checked for over-narrowing, and it does not lose a case.** The rule is now scoped to "another `ai-agent-assembly` repository," with third-party branch names explicitly disclaimed. I grepped this repo for links into the org's *other* GitHub org (`horonomy/…` `blob`/`raw`/`tree`) — **zero**, so nothing in the tree falls through the new wording. One forward-looking edge, not a finding and not worth changing now: if a link into `horonomy` is ever added, the wording would class it as third-party even though those branch names arguably *are* yours to track. Cheap to widen to "another repository in an org we control" if that ever comes up.

### On MINOR-3 — the judgement you asked for, plainly

**Yes, the `MUST` on the reference-writing obligation discharges the finding, and the decision not to re-import "audit both directions" into the ADR is the correct call — better than what my finding implied.**

My finding had two parts. The first — zero `MUST` in the reduced ADR — is straightforwardly fixed. The second, that the pilot's lesson lacked a public normative anchor, was where my framing was imprecise, and the delta exposes it: I pointed at a real gap but gestured at the wrong remedy.

Test it against the actual break. The pilot's failure was `agent-assembly/release.yml` carrying `base: master` into homebrew-tap. The new D3 reads: *a consumer's PR `base:` **MUST** name that repo's current default branch* — that prohibits exactly that defect, publicly, permanently, at MUST grade, and machine-checked by `check-release-completeness.sh`. The concrete failure mode is now publicly and normatively closed.

"Audit both directions," by contrast, is a *discovery step executed during a rename*. Its audience is the operator performing a migration, not a contributor writing a reference. Importing it into the ADR to satisfy my finding would have re-imported a migration step into a document the whole change exists to strip of migration steps — trading a real improvement for a cosmetic one. Declining that was right.

The decomposition the delta makes explicit is the correct one: **the obligation is durable and belongs in the ADR; the procedure that discovers violations of it is process and belongs in the runbook.** The new bridging paragraph states this rather than leaving a future reader to infer it, which is what makes the split legible instead of merely tidy. No residual: the checklist's self-revision trigger now sits next to the checklist it guards, which is the only place it is actionable.

### The verification-design point belongs in the record

Recording this because it generalises well past this PR, and because it is a sharper diagnosis than the one I gave:

> the reason the four losses clustered in the normative class is that my verification was *designed* around token/sentence matching, which is structurally blind to a swapped obligation or a substituted trigger. The lesson isn't that I got unlucky in one class; it's that **the check I built could not have caught them**, so reporting it as a completeness proof was the actual error.

That is right, and it is the better framing. My version ("the losses clustered in the normative class") reads as bad luck in one bucket. The real property is that a similarity measure can only detect *missing tokens* — a substituted reconsideration trigger and a downgraded `MUST` both **preserve** token overlap, so no threshold on that metric could ever have surfaced them. The generalisable rule: a completeness claim needs a check whose failure mode differs from the artifact's failure mode. For moved normative text that means enumerating obligations, exceptions, and triggers as a list and verifying each individually — not measuring aggregate overlap and inferring completeness from a high score.

### NIT — one stale number in the body

The publish note still reads "final tree `6991182a`". That was the tree at `ad6013f8`; the current final tree is **`b2711d14`** (which is what I verified parity against). Worth correcting since it is precisely the claim the note exists to support.

### Verdict — READY TO MERGE at `96cac4ac`

The reclassification is correct, the split seam is right, the destination is verified, index/count bookkeeping is exact, the replay is byte-clean, and all four normative losses are restored with the one judgement call resolved better than I proposed. Still not approving and not merging — that is the code owner's call, and this must land **together with** `ai-agent-assembly/internal-docs#93` (`1799cd1a`).

Recorded as known-open and non-blocking, not to be re-litigated here: **AAASM-5297** (gate/registry contradiction, another repo's contract), **AAASM-5291** (ADR status defects), **AAASM-5292** (unrecorded 0005 subject), the unversioned workspace `CLAUDE.md`, and #93's billing-blocked CI.

Chisanan232:
NIT closed — the publish note said tree `6991182a`, which was the tree at the superseded `ad6013f8`. Corrected to **`b2711d14`**, the actual tree at head **`96cac4ac`** (verified against `.commit.tree.sha` on the remote and byte-identical to local `HEAD^{tree}`), and the note now names the head it describes so any future drift is visible rather than silent.

I swept every SHA in both PR bodies and both comment threads rather than just the one flagged: `6991182a` was the **only** stale value. `96cac4ac`, `b2711d14`, `aae32497` (still the true merge-base per the compare API, though `main` is now 41 commits ahead) and #93's `1799cd1a` all verify. Body edit only — no gates run, because none were needed.

Both PRs merge cleanly and still touch exactly the 5 intended files. Not approving, not merging — they land together, on Bryant's call.

PR #1825, opened 2026-07-30
[AAASM-5106] 🐛 (aa-api): Render Unknown, not a false Allow/No-policy, when the cascade is unloaded

Chisanan232:
## Description

The **interim truthfulness fix** for AAASM-5106 (decision B1). In every shipped aa-api deployment the policy scope-index cascade is empty (aa-api builds the engine via `load_from_file`, never `load_cascade_from_dir`), so every cascade-derived reporting projection rendered a confident, FALSE answer — worst of all the Teams card asserting "No policy is in force" while the primary slot actively enforces, and the capability matrix reading **ALLOW** for capabilities that are actually denied.

Enforcement was never affected (it falls back to the primary slot); only the reporting surfaces lied. This PR makes them honest.

**What changed:**
- `aa-gateway`: new `PolicyEngine::cascade_loaded()` — the loaded/unavailable signal (`!scope_index.is_empty()`).
- **Teams active-policies**: returns `policies: None` (Unknown) when the cascade is unloaded, never `Some([])` — so the UI never says "No policy is in force" from an empty data source.
- **Capability matrix**: carries a `cascade_loaded` flag; when false, the matrix renders **not-evaluated** rather than the unsound ALLOW-by-default cells.
- **Topology permission chain**: `policy_count`/effective-permissions render Unknown, not a real "no policies", when unloaded.
- Dashboard consumes the signal on all three surfaces via the existing truthfulness vocabulary (StatusState / Certain / display helpers).

## Type of Change
- [x] 🐛 Bug fix

## Breaking Changes
- [x] No — additive `cascade_loaded` signal + absent-vs-empty discipline; enforcement unchanged.

## Related Issues
- Related Jira ticket: AAASM-5106 (interim fix)
- **Real fix (loader wiring) tracked separately as AAASM-5299** — this PR deliberately does NOT wire aa-api to load the cascade; it only stops the shipped surfaces from lying while unloaded.

## Testing
- Rust: `cascade_loaded_reflects_scope_index_population`; per-surface `matrix_reports_cascade_unloaded_and_distinguishes_it_from_a_real_policy`, `team_policies_are_unknown_not_empty_when_the_engine_carries_no_cascade` (+ zero-members variant). With a loaded cascade the real answer still flows.
- `cargo test -p aa-gateway -p aa-api` green; `clippy --all-targets` + `fmt` clean; OpenAPI drift in sync.
- Dashboard: `tsc` + `eslint` clean; `vitest` 3117 pass.

## Truthfulness
- Never a confident Allow / No-policy / zero when the cascade is unloaded — the exact false-positive family this ticket exists to remove.

### DIFF
--- a/aa-api/src/models/capability.rs
+++ b/aa-api/src/models/capability.rs
@@ -188,6 +188,24 @@ pub struct CapabilityMatrix {
     pub agents: Vec<CapabilityAgent>,
     pub policies: Vec<Policy>,
     pub sample_calls: Vec<SampleCall>,
+    /// Whether the projecting engine actually carries a policy cascade
+    /// (`PolicyEngine::cascade_loaded`). AAASM-5106 / ADR 0024: when the cascade
+    /// is unloaded — the state of every shipped aa-api deployment today, which
+    /// loads its policy through `load_from_file` and leaves `scope_index` empty —
+    /// `decide()` falls through to `Allow` for every cell, so the grid asserts an
+    /// unbroken wall of `ALLOW` for capabilities the primary policy actually
+    /// denies. `false` here is the matrix-level "not evaluated — policy cascade
+    /// not loaded" signal the dashboard renders as an unavailable state, so an
+    /// operator never reads a fabricated `Allow` as a real grant. It is a fact
+    /// about the projection's *data*, never an operator choice; the enforcement
+    /// path (`evaluate_primary`) is unaffected.
+    ///
+    /// Required-but-always-present, not optional: the key is always on the wire,
+    /// so a client reads an explicit `false` it must handle rather than a missing
+    /// field it can shrug off — the same absent-vs-unknown discipline as
+    /// [`CapabilityAgent::trust`] and `TeamPoliciesResponse::policies`.
+    #[schema(required = true)]
+    pub cascade_loaded: bool,
 }
 
 /// One agent row in the dashboard Capability Matrix.
@@ -464,6 +482,7 @@ mod tests {
             agents: vec![],
             policies: vec![],
             sample_calls: vec![],
+            cascade_loaded: false,
         };
         let json = serde_json::to_value(&matrix).unwrap();
         assert!(json["resources"].is_array());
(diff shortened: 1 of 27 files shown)


Chisanan232 (commented):
**Claude Code — self-review (AAASM-5106 / decision B1, interim truthfulness fix)**

Verified against "absent/unloaded cascade → Unknown/Unconfigured/Not-evaluated; never Allow/No-policy/zero/empty-chain as a real answer; expose an explicit loaded/unavailable signal. Keep the interim fix and the aa-api loader wiring in SEPARATE PRs."

- ✅ **Explicit loaded signal, sourced from the engine.** `PolicyEngine::cascade_loaded()` (`!scope_index.is_empty()`); carried on the wire as `cascade_loaded` on both the capability matrix (`cascadeLoaded`) and each topology node's `effective_permissions` (`cascade_loaded`) and the graph response.
- ✅ **Absence renders honestly, never a false positive:**
  - Capability summary → Unconfigured/Not-evaluated when `!cascadeLoaded` (`cascadeEvidenceFromQuery`), even if a policy row happens to be listed.
  - Team active-policies → `policies: None` (Unknown), not `[]`, when the engine carries no cascade.
  - Topology node `policy_count` → `null` (not `0`); `NodeDetailPanel` hides the inheritance chain and shows the unloaded marker rather than an empty chain reading as "no policies apply."
- ✅ **Scope kept interim.** This PR is the truthfulness fix only; the real fix (aa-api loading the same file-or-directory `AA_POLICY` source model as aa-gateway) is deferred to **AAASM-5299**, per the "separate PRs" instruction.
- ✅ **e2e correctness (final commit `379e993d`):** `cascade_loaded` is a **per-node** signal on `effective_permissions` (not response-level) — the earlier fixture placement is corrected in `review-aaasm-5099` (planner node) and `verify-aaasm-5090` (`cascadeLoaded` on the loaded MATRIX); Dashboard e2e re-ran green.

All checks green. No self-merge — awaiting code-owner approval.

PR #1834, opened 2026-07-30
[AAASM-5299] ✨ (state): Wire aa-api to load the policy cascade from $AA_POLICY

Chisanan232:
## Description

The real-fix half of AAASM-5106 (ADR-0023 Option a, Accepted). `aa-api` now loads a policy cascade from an operator-settable source using the **same `$AA_POLICY` semantics and the same `load_cascade_from_dir` loader that `aa-gateway` uses**, so the dashboard's cascade-derived projections (capability matrix, topology permission chain, Teams active-policies) reflect the policy source the gateway actually enforces.

**Read-only projection change — no enforcement path touched.** Before this, `AppState` built the engine via `load_from_file` (primary slot only) and the scope index was empty in every deployment, so those projections read empty; the interim truthfulness fix (AAASM-5106 / #1825) made that render honestly as *Unknown*. This PR makes the data real.

### Source resolution (matches aa-gateway)
- **Directory** → `load_cascade_from_dir` (scope index populated → `cascade_loaded() == true`, projections show real data).
- **File** → `load_from_file` (primary slot, today's behaviour → `cascade_loaded() == false`).
- **Unset / empty / missing** → budget-only bootstrap as before (`cascade_loaded() == false`) — the generated bootstrap is **never** presented as an operator-authored cascade, so the ADR-0024 Unknown state holds.

## Type of Change

- [x] ✨ New feature

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-5299 (completes AAASM-5106; ADR-0023 / ADR-0024)

## Testing

`resolve_policy_source` path-shape routing unit test; three env-isolated integration tests (directory populates cascade / file keeps primary-slot / unset leaves cascade empty). Local gates: `cargo clippy --all-targets` clean, `cargo fmt --check` clean, `cargo test -p aa-api` lib+new tests green (2 pre-existing webhook-egress flakes in `tests/destinations.rs`, unrelated — verified failing on HEAD independently). Authoritative full build runs in CI.

## Checklist

- [ ] Commits are signed off — DCO optional/advisory here

### DIFF
--- a/aa-api/src/state.rs
+++ b/aa-api/src/state.rs
@@ -263,44 +263,70 @@ impl AppState {
         let uniq = COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         let pid = std::process::id();
 
-        let policy_dir = std::env::temp_dir().join(format!("aa-api-local-policy-{pid}-{uniq}"));
-        std::fs::create_dir_all(&policy_dir).map_err(|source| LocalStateError::PolicyWrite {
-            path: policy_dir.clone(),
-            source,
-        })?;
-        let policy_path = policy_dir.join("local-policy.yaml");
-        std::fs::write(
-            &policy_path,
-            // Minimal valid section-based envelope. AAASM-3351 made the
-            // validator fail closed on the legacy rule-list schema, so this
-            // must use the `spec:` section form.
-            "apiVersion: agent-assembly/v1\n\
-             kind: Policy\n\
-             metadata:\n  \
-             name: local-policy\n  \
-             version: \"0.1.0\"\n\
-             spec:\n  \
-             budget:\n    \
-             daily_limit_usd: 100.0\n",
-        )
-        .map_err(|source| LocalStateError::PolicyWrite {
-            path: policy_path.clone(),
-            source,
-        })?;
-
         let events = Arc::new(EventBroadcast::default());
         let budget_alert_tx = events.budget_sender();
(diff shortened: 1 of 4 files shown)


Chisanan232 (commented):
**Claude Code — self-review (AAASM-5299 / ADR-0023 Option a)**

Verified against the decision "wire aa-api to load the same policy source model as aa-gateway (AA_POLICY file-or-directory); generated bootstrap not presented as operator-authored; read-only projection."

- ✅ **Same loader as gateway** — directory→`load_cascade_from_dir` (scope index populated, `cascade_loaded()==true`), file→`load_from_file` (primary slot), unset→bootstrap with `cascade_loaded()==false`. Matches aa-gateway's `$AA_POLICY` path-shape resolution.
- ✅ **Bootstrap not faked as a cascade** — unset leaves the scope index empty, so ADR-0024's Unknown state holds; the generated budget-only bootstrap is never presented as operator-authored.
- ✅ **Read-only** — only `AppState` engine construction changed; no enforcement path touched; existing `.with_registry`/`.with_invalidation_hub` chaining preserved across all three branches.
- ✅ **6 atomic commits** (1 impl + 4 tests + 1 fmt), disjoint from the C1 lane, no OpenAPI/ADR touch.
- ✅ **CI all green** (authoritative full build that local disk couldn't run).

Merged on green. AAASM-5106's interim Unknown states now become live real data when a cascade directory is configured.

PR #1835, opened 2026-07-30
[AAASM-5100] ✨ (gateway): Populate verdict + latency_ms at the decision boundary (Phase 1)

Chisanan232:
## Description

ADR-0018 decision-capture plan, **Phase 1 (items A + B only)** — authorised 2026-07-30. Populates the `verdict` and `latency_ms` fields that AAASM-5086 froze as nullable on the read-side contract. **Item C (trace_id propagation) is NOT implemented** — it stays gated as a separate Phase 2; `trace_id` remains `null` (with a test asserting so).

- **A — 5-way `RuntimeVerdict`:** the finer `allow / narrow / scrub / pending / deny` verdict is derived alongside the proto `Decision` and written onto the audit payload. `scrub` = the DLP layer rewrote the payload; `narrow` = a policy match scoped an otherwise-allowed action.
- **B — per-decision latency:** measured with a monotonic clock around policy evaluation and recorded as `latency_ms`.

### Key design note (differs from the ADR's literal derivation point — please review)
ADR-0018 A/B name `aa-runtime::RuntimeScanner` as the derivation point. In the actual code that scanner is a **DLP-scrub-only** stage and does **not** write the decision audit entry — the gateway's `policy_service::record_audit` does, from `check_action`'s `EvaluationResult`, and that is where the proto decision, the scrub signal, and the **monotonic `decision_latency_us`** already live. So verdict + latency are derived at the gateway `record_audit` boundary. This required **no proto `Decision` change and no tracing work**, staying strictly inside A/B.

- **Latency semantic guardrail:** `latency_ms` measures the **decision** time (around `engine.evaluate`), taken *before* any approval-queue wait — so a held/approval-pending action records the scanner's decision time, **not** the human approval wait. Test-locked.
- **`narrow` source:** added a minimal observational `narrowed: bool` on `EvaluationResult` (capability fell inside an in-force restricted allow-list). It is **metadata on an unchanged Allow — no enforcement decision changes.**
- **`verdict` on the wire:** written as a JSON string by the gateway (which can't name `aa-api`'s `RuntimeVerdict` type) and parsed back on the read-side.

**OpenAPI unchanged** — `RuntimeVerdict`/`AgentDecisionResponse` were already registered (frozen in AAASM-5086); regenerated spec is byte-identical, path count stays 80.

## Type of Change

- [x] ✨ New feature

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-5100 (Phase 1; Phase 2 / trace_id tracked separately)

## Testing

9 commits, one test per commit: `verdict = scrub` (not allow), `verdict = narrow` (not deny), positive `latency_ms`, held-approval records decision latency not the wait, and `trace_id` stays null. Local gates on the three touched crates (`aa-runtime`/`aa-gateway`/`aa-api`): `cargo test` green (1 pre-existing Docker/Postgres testcontainer failure, no Docker in env), `clippy --all-targets` zero warnings, `fmt --check` clean. Full-workspace `--all-targets` couldn't link locally (disk); authoritative build runs in CI.

## Checklist

- [ ] Commits are signed off — DCO optional/advisory here

### DIFF
--- a/aa-api/src/routes/agents.rs
+++ b/aa-api/src/routes/agents.rs
@@ -1697,14 +1697,27 @@ mod tests {
         assert_eq!(row.verb.as_deref(), Some("TOOL_CALL"));
         assert_eq!(row.resource.as_deref(), Some("pg.users"));
         assert_eq!(row.matched_policy, None);
-        // No per-decision latency source exists — it must be reported as absent.
+        // A legacy payload carrying no verdict/latency_ms keys (written before
+        // AAASM-5100 capture landed) must still read null — the capture is
+        // additive, never fabricated for rows that lack a source.
         assert_eq!(row.latency_ms, None);
-        // The verdict + trace id are frozen in the schema but not yet captured
-        // at decision time (ADR-0018 hot-path follow-up); they must read null.
         assert_eq!(row.verdict, None);
         assert_eq!(row.trace_id, None);
         assert_eq!(row.seq, 7);
         assert_eq!(row.session_id, "ee".repeat(16));
+    }
+
+    #[test]
+    fn scrubbed_action_records_scrub_verdict_distinct_from_allow() {
+        // AAASM-5100 item A — a DLP-scrubbed action is forwarded (proto Redact,
+        // decision=4) but its captured verdict must be `scrub`, never `allow`,
+        // so scrubbed traffic is visible as distinct from a clean allow.
+        let entry = decision_entry(
+            r#"{"action_type":"TOOL_CALL","decision":4,"verdict":"scrub","latency_ms":3,"detail":{"kind":"tool_call","tool_name":"gmail.send"}}"#,
+        );
+        let row = entry_to_decision_row(&entry).expect("redact carries a decision");
+        assert_eq!(row.verdict, Some(RuntimeVerdict::Scrub));
+        assert_ne!(row.verdict, Some(RuntimeVerdict::Allow));
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


Chisanan232 (commented):
**Claude Code — self-review (AAASM-5100 Phase 1 / ADR-0018 A+B)**

Verified against the authorised scope: capture items **A (verdict) + B (latency) only, C (trace_id) stays gated**; don't fill nullable fields to look complete; no proto/enforcement change.

- ✅ **A — 5-way verdict** derived alongside the proto `Decision` and written to the audit payload (`scrub`=DLP rewrote, `narrow`=scoped allow). Tests lock `scrub≠allow` and `narrow≠deny`.
- ✅ **B — monotonic latency**, measured around `engine.evaluate` **before** the approval-queue wait, so a held action records decision time not the human wait — the ADR's required semantic, test-locked (`held_approval_records_decision_latency_not_the_approval_wait`).
- ✅ **C not done, and proven so** — `trace_id` stays `null` with an explicit assertion test, so nobody mistakes it for shipped.
- ✅ **Derivation point deviation surfaced, not hidden** — the ADR names `aa-runtime::RuntimeScanner`, but that stage is DLP-scrub-only and doesn't write the decision audit; the gateway's `record_audit` does (and already has the monotonic `decision_latency_us` + scrub signal). Deriving there kept the change inside A/B with **no proto change and no tracing work**. Flagged in the PR body for review.
- ✅ **No enforcement decision change** — `narrowed` is observational metadata on an unchanged Allow. **OpenAPI byte-identical** (fields were pre-frozen in AAASM-5086), path count stays 80.
- ✅ **CI all green** (the authoritative full build).

Phase 2 (trace_id propagation across SDK→runtime→gateway) remains gated — filing as a separate follow-up. Merged on green.

PR #1852, opened 2026-07-31
[AAASM-1112] ✅ (aa-integration-tests): Record the AAASM-201 acceptance verification

Chisanan232:
## Description

Runs AAASM-1112 — the acceptance verification for AAASM-201 — against the real conformance suite delivered by AAASM-5283, and records the result.

**The verification does not pass.** AC4 (`aasm run claude` end to end) fails against the shipped router, so this PR delivers the evidence and the regression test rather than a sign-off. AAASM-1112 stays open and AAASM-201 must not be closed.

Two commits:

- **`aa-integration-tests/tests/cli_run_claude_governed_launch.rs` (new)** — two tests that drive the governed-launch path for real. Until now the `which claude` detection branch had never been exercised: every `detect` test injected a binary override, so the branch that actually runs on a user's machine was untested.
- **`verification-reports/AAASM-1112-claude-code-acceptance-verification.md` (new)** — the full record: commit SHA, OS/toolchain/tool versions, every command with its verbatim result, the complete skip list, and the AC-by-AC verdict.

## What was measured

commit `2e543884` · macOS 26.4.1 (25E253, arm64) · rustc 1.97.0 · cargo 1.97.0 · `claude` 2.1.220 at `/opt/homebrew/bin/claude`

| Command | Result |
|---|---|
| `cargo build --workspace --tests` | 0 errors |
| `nextest -p aa-devtool-claude-code` | 134 passed, 0 skipped |
| `nextest -p aa-core -p aa-runtime` | 932 passed, 2 skipped |
| `nextest -p aa-cli` | 874 passed, 0 skipped |
| `nextest -p aa-devtool` | 25 passed, 0 skipped |
| `nextest --test conformance_claude_code` | 26 passed, 0 skipped |
| `nextest --test claude_code_integration_lifecycle` | 12 passed, 0 skipped |
| `nextest --test cli_run` | 9 passed, 0 skipped |
| `nextest --test cli_run_claude_governed_launch` (new) | 2 passed, 0 skipped |

The two skips are both `#[ignore]` benchmarks (`round_trip_latency_under_1ms`, `pipeline_load_benchmark`) and bear on no AC.

**No conformance scenario was skipped or left unmeasured.** The optional real-tool lane executed rather than short-circuiting: `require_claude` and `require_macos` both passed, and it measured 3 requests reaching the provider with 2 of 3 bodies carrying `[REDACTED:AnthropicKey]`. That distinction matters — a suite that reports green because its only load-bearing scenario opted out is not evidence.

## AC verdict

| AC | Verdict |
|---|---|
| AC1 — detect (`which claude`, `~/.claude/`) | Satisfied |
| AC2 — generate managed `settings.json` | Satisfied |
| AC3 — MCP allow/deny lists | Satisfied |
| AC4 — `aa run claude` end to end | **Not evidenced — measured failure** |
| AC5 — unit tests | Satisfied |

AC4 is not a coverage gap. `aasm run` registers via `POST /api/v1/agents` (`aa-cli/src/commands/run.rs:209`), but `aa-api/src/routes/mod.rs:124` mounts only `get(agents::list_agents)` on that path and `openapi/v1.yaml` agrees — no POST handler exists. Measured: `error: gateway registration failed: HTTP 405 Method Not Allowed`, exit 1, and the tool is never launched. A second, latent defect sits behind it: `build_child_env` overwrites the adapter's normalised proxy URL with a scheme-less `host:port`, which the existing test cannot observe because it feeds in a value that already has a scheme. Both are filed as Bug Subtasks.

## Type of Change

- [x] ✅ Tests
- [x] 📝 Documentation update

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-1112 (verification subtask of AAASM-201)
- Related GitHub issues: verifies the suite delivered in #1832

## Testing

One assertion was proved load-bearing by mutation. Being precise about which, because the first draft of this section overstated it:

- **Genuine**: adding `--no-proxy` to the launch args → **FAILED** with `left: Some("") right: Some("127.0.0.1:19999")`. That mutation changes real production behaviour.
- **Vacuous, and withdrawn as evidence**: inverting `assert!(!dump.exists())` fails by construction — `dump` lives in a fresh `tempfile::tempdir()` written only by the stub, so it can never exist at that point. It demonstrates nothing about whether the guard would catch a regression. The meaningful mutation (point test 2 at a gateway that answers, confirm `!dump.exists()` then fails) has not been run.
- **Not mutation-tested at all**: the five identity keys, `kind`, `version`, the registration count, the deregistration, and the real-home guard.

The developer's real `~/.claude/settings.json` is guarded by `RealHomeGuard` (length+mtime, contents never read). Two limits, stated because this is cited as a safety property: it covers that one file, and where the file does not exist — every CI runner — a `None` fingerprint compares equal to `None`, so the guard is inert there. It is load-bearing on a developer machine, which is where the risk lives.

## Checklist

- [ ] All CI checks passing

### DIFF
--- a/aa-integration-tests/tests/cli_run_claude_governed_launch.rs
+++ b/aa-integration-tests/tests/cli_run_claude_governed_launch.rs
@@ -129,7 +129,7 @@ mod governed_launch {
             &bin,
             r#"#!/bin/sh
 if [ "$1" = "--version" ]; then
-  echo "2.1.220 (Claude Code)"
+  echo "2.1.999 (Claude Code)"
   exit 0
 fi
 {
@@ -336,7 +336,7 @@ exit 0
         );
         assert_eq!(
             body.get("version").and_then(serde_json::Value::as_str),
-            Some("2.1.220"),
+            Some("2.1.999"),
             "the registration must carry the detected version, not a placeholder; body: {body}",
         );
         assert_eq!(
--- a/aa-integration-tests/tests/cli_run_claude_governed_launch.rs
+++ b/aa-integration-tests/tests/cli_run_claude_governed_launch.rs
@@ -36,9 +36,11 @@
 //! the **child** `aasm` process only (`HOME`, `PATH`, `CLAUDE_CONFIG_DIR`,
 //! `AASM_STATE_DIR`, `AA_CA_DIR`, `AASM_CLAUDE_MANAGED_ROOT`, and the working
 //! directory); no process-global environment state in this test binary is
-//! mutated. The final assertion is a real-home guard in the style of the
-//! AAASM-5283 conformance suite: the developer's own settings file is read
-//! before and after and must be byte-identical.
+//! mutated. The final assertion reuses the AAASM-5283 conformance suite's
+//! `RealHomeGuard`, which fingerprints the developer's settings file on length
+//! and mtime and never reads its contents — that file is in daily use and may
+//! hold credentials, so a byte comparison would print them into the failure
+//! message and from there into the CI log.
 
 // `common/mod.rs` carries its own inner `#![allow(dead_code)]`; only the unused
 // imports need allowing here (this file uses just `common::cli::CliFixture`).
@@ -168,11 +170,33 @@ exit 0
                     .expect("aa-integration-tests always has a workspace-root parent")
                     .join("target")
             });
+        for profile in ["debug", "release"] {
+            let bin = target.join(profile).join("aasm");
+            if bin.is_file() {
+                return bin;
+            }
+        }
+
+        // `aa-integration-tests` does not depend on `aa-cli`, so nothing builds
+        // the binary for us: `cargo nextest run -p aa-integration-tests` on a
+        // clean target dir would otherwise fail here rather than run. Build it
+        // once. The working directory matters — the test bodies run `aasm` from
+        // inside a temp dir (the settings resolver prefers `<cwd>/.claude`, so
+        // running from the checkout would let it write into the repo), and cargo
+        // cannot find a manifest from there. Building from the manifest dir up
+        // front is what lets the run itself stay hermetic.
+        let status = std::process::Command::new(env!("CARGO"))
+            .args(["build", "--quiet", "-p", "aa-cli", "--bin", "aasm"])
+            .current_dir(env!("CARGO_MANIFEST_DIR"))
+            .status()
+            .expect("failed to invoke cargo to build the aasm binary");
+        assert!(status.success(), "`cargo build -p aa-cli --bin aasm` failed");
+
         let bin = target.join("debug").join("aasm");
         assert!(
             bin.is_file(),
-            "no `aasm` binary at {} — build it (`cargo build -p aa-cli --bin aasm`) or set \
-             AASM_BIN_PATH. Skipping instead would leave AAASM-201 AC4 unevidenced.",
+            "no `aasm` binary at {} even after building it — set AASM_BIN_PATH. Skipping \
+             instead would leave AAASM-201 AC4 unevidenced.",
             bin.display(),
         );
         bin
@@ -289,8 +313,10 @@ exit 0
                 seen.get(key).map(String::as_str),
                 Some(PROXY_ADDR),
                 "the launched tool must be routed at the proxy address the gateway assigned via \
-                 `{key}`. If this now reads `http://{PROXY_ADDR}`, AAASM-1112 finding (2) has been \
-                 fixed — update this assertion and the AAASM-201 AC4 verdict; saw:\n{raw}",
+                 `{key}`. NOTE: this pins the CURRENT, DEFECTIVE value on purpose — the designed \
+                 value is `http://{PROXY_ADDR}` (AAASM-5324, root-caused by AAASM-5327). If this \
+                 now reads `http://{PROXY_ADDR}`, that bug is fixed: change this assertion to the \
+                 designed value and re-derive the AAASM-201 AC4 verdict. Saw:\n{raw}",
             );
         }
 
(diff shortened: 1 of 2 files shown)


Chisanan232:
### Review round applied

Five findings fixed. Verified each in source before acting rather than taking the report on trust.

**Security regression in my own test — fixed (`d98a45cf`).** The guard compared the *bytes* of the developer's live `~/.claude/settings.json` before and after the run. `assert_eq!` prints both sides on failure, so a regression would have dumped a file that routinely holds API keys and MCP tokens into the test output and the CI log — the guard's failure path leaking exactly what it exists to protect. `RealHomeGuard` already solves this and documents why (*"Contents are deliberately not printed"*): length+mtime, sufficient because every writer in play replaces the file. Reused rather than reimplemented.

**A product bug more serious than the two this PR filed — now AAASM-5327 (`09e3b951`).** `spawn_and_wait` (`run.rs:431`) does not launch the `Command` the adapter built. It constructs a fresh `tokio::process::Command` from `get_program()` + `get_args()` and applies only `child_env`; `cmd.get_envs()` is never read. So **every** variable `build_launch_command` set is discarded — including `NODE_EXTRA_CA_CERTS`, the sole mechanism by which Claude Code's Node runtime trusts the AASM CA. Without it the proxy cannot terminate TLS and the traffic is uninspectable: the tool launches and looks governed. `cmd.envs(&child_env)` at `run.rs:547` is dead code.

This **supersedes the root-cause analysis in AAASM-5324**, which this report described as an overlay. It is a total discard, and the proxy URL was only its most visible casualty — re-ordering `cmd.envs()`, the fix the old wording implied, would have changed nothing. Report corrected.

**Two other overstated claims in the report, corrected (`09e3b951`).** The real-home safety paragraph cited the byte comparison that no longer exists and claimed more coverage than the guard has — it watches one file, and on any host where that file is absent a `None` fingerprint compares equal to `None`, so it is inert there. And *"stripped from the published crate"` was read as *"absent from released builds"*: `.ci/strip-for-publish.sh` runs only in `publish-crates`, so the GitHub Release and Homebrew binaries are built unstripped and **do** carry `aasm run`. The launcher defect therefore reaches users on the primary install path — it is not confined to source builds.

**Test would fail rather than run on a clean target dir (`94ac9efe`).** `aa-integration-tests` has no `aa-cli` dependency, so nothing builds `aasm` for it. Now looks for both profiles and builds it once from the manifest dir; the test bodies still run from a temp cwd, which is what keeps the run out of the checkout. Same commit labels the two assertions that deliberately pin *current defective* behaviour with the tickets that own them (AAASM-5323, AAASM-5324/5327) and states the designed value, so a bug-fix PR reads the failure as expected rather than as a regression.

**Stub version was indistinguishable from the real tool (`fce84fca`).** It reported `2.1.220` — byte-identical to the `claude` on the verification host — and the report cited that line as AC1 evidence next to an environment table showing the same string. Now `2.1.999`, which is in no release. Worth noting the adapter normalises to semver core, so `2.1.999-stub` does not survive the parse.

`2 tests run: 2 passed, 0 skipped` after every change; `cargo fmt` and `clippy -D warnings` clean.

### Not fixed, deliberately

- **The mutation claim in the PR body was overstated** and I have corrected it there rather than here: only the proxy assertion got a genuine mutation. The `dump.exists()` inversion is vacuous — `dump` lives in a fresh tempdir written only by the stub, so inverting it fails by construction.
- **Blocking `Command::output()` inside a `multi_thread` tokio test** is a real hang risk on a 1-vCPU runner. Left as-is because changing the concurrency model of a test that currently passes is not something to do in a review round without measuring it; recorded for follow-up.
- **AC1/AC2/AC3 verdicts** — the review argues these are "a test exists" evidence rather than end-to-end evidence (empty `load_policy()`, no production caller for `apply_capability_set`, the discarded `~/.claude` marker). That reading has merit and it does not change this PR's conclusion, which is already that AAASM-1112 cannot be signed off. Rather than relitigate three verdicts in a PR whose headline finding is a hard fail, they stay as recorded with the caveats the report already carries.

Chisanan232:
## Maintainer review — approved for merge, as failed-verification evidence

(Comment rather than an approving review: GitHub refuses self-approval.)

Risk class: **environment handling + a test that touches a developer's real home.** Merging this records a verification that **did not pass**; that is the point of it.

### Test safety re-verified on the merge head

- `RealHomeGuard` is used (`:218`) — length+mtime, contents never read. The earlier version of this test compared the **bytes** of the live `~/.claude/settings.json` inside an `assert_eq!`, whose failure path prints both sides into the CI log; that is fixed.
- The only `read_to_string` is the temp dump file (`:275`), and the only `settings.json` assertion is against the **redirected** home (`:352`).
- No `set_var`/`remove_var` anywhere — every redirection is child-scoped.

### The AC4 failure is real, not a harness artefact

Independently confirmed: `aa-api/src/routes/mod.rs:124` mounts `get(agents::list_agents)` only, `openapi/v1.yaml:102` agrees, and the fixture mounts the genuine `aa_api::server::build_app`. `proxy_addr` occurs nowhere in `aa-api`, `aa-gateway`, `aa-proto` or `openapi/`. There is no alternate HTTP surface, feature flag or base path that would make the 405 an artefact of the test.

### Correcting myself on one point

I previously said this file's binary resolution "asserts the binary exists" and would false-pass. That is not accurate on this head: `aasm_binary()` **builds** `aasm` when neither a debug nor a release binary is present. The residual weakness is narrower — when a binary *is* present it is reused without rebuilding, so a mutation run against a stale build could still false-pass, which is exactly how the sibling work in #1855 lost its first mutation result. Worth hardening; not a merge blocker, since this file's own mutations were run and are quoted in the description.

### Honest limits already carried in the PR

The mutation claim is scoped correctly: only the `--no-proxy` mutation is genuine. Inverting `assert!(!dump.exists())` fails **by construction** — `dump` lives in a fresh tempdir written only by the stub — and the description says so and withdraws it as evidence. Nine other assertions were not mutation-tested. `RealHomeGuard` watches one file and is inert where that file is absent, which is every CI runner.

### Ticket semantics — explicit

Merging this **does not** authorise closing AAASM-1112. That subtask stays open until AAASM-5323 and AAASM-5327 are fixed and the full real-tool conformance suite passes against `main` afterwards. **AAASM-201** stays open until AAASM-1112 is genuinely complete. Both are already linked: 1112 *is blocked by* 5323 and 5327, *relates to* 5324.

Gates: not a draft; MERGEABLE, no conflicts; 23 pass / 18 skipping / 0 failing; 0 unresolved threads; no requested-changes review; head `fce84fca` current.

PR #1855, opened 2026-07-31
[AAASM-5327] 🐛 (aa-cli): Pass the adapter's launch environment to the child

Chisanan232:
## Description

`aasm run <tool>` discarded **every** environment variable the dev-tool adapter set on the launch command — including `NODE_EXTRA_CA_CERTS`, the only mechanism by which Claude Code's embedded Node runtime trusts the Agent Assembly CA.

Without it the proxy cannot terminate TLS, so nothing is inspected. **The tool launches, reports as governed, and is not.** That is the silent-bypass failure class the product exists to prevent, and it is why this is a bug fix rather than a tidy-up.

`spawn_and_wait` did not launch the `Command` the adapter built. It constructed a fresh `tokio::process::Command` from `cmd.get_program()` and `cmd.get_args()` and applied only `child_env`; `cmd.get_envs()` was never read.

Two commits:

- **`a6cb6ff5` 🐛 (aa-cli)** — the child's environment is now the union of both sources, with **the adapter applied last so it wins on collision**. The adapter is the layer that knows what the tool actually needs, and the layer that normalises the gateway's bare `host:port` into the `http://host:port` URL an HTTP client will route through (AAASM-5324). `None` from `get_envs()` is honoured as `env_remove` rather than flattened to empty — an adapter that *unsets* a variable to disable a tool behaviour must not have that turned into an empty-but-present variable the tool then honours.
- **`05bf0b53` ✅ (aa-integration-tests)** — a test that asserts on the launched child's **own** environment.

### `cmd.envs(&child_env)` was not merely dead code

It ran *before* anything could read `cmd`, overwriting the adapter's normalised proxy value inside the command with the bare one. Left in place alongside the merge, it would have faithfully carried forward the very values the merge exists to override. Mutation 2 below is that exact experiment.

## Why no existing test caught it

- `build_child_env_sets_proxy` (`run.rs:883`) feeds in `"http://proxy:8080"` — a value that **already has a scheme** — so the normalisation branch is never exercised.
- `aa-cli/tests/integrations_claude_code.rs` reads the on-disk launch-env store, proving the value was *computed*, not that it was *delivered*.

Nothing anywhere inspected a launched process's actual environment. This PR adds that.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-5327 (subtask of AAASM-201; supersedes the root-cause analysis in AAASM-5324)
- Related GitHub issues: found during the AAASM-1112 verification in #1852

## Testing

Every new assertion was proved load-bearing by mutation. Verbatim:

**M1 — merge loop deleted.**
```
assertion `left == right` failed: the launched tool must receive the CA bundle the adapter
injected; `__UNSET__` here means the adapter's environment was discarded and TLS interception
is dead.
  left: Some("__UNSET__")
 right: Some("/aaasm5327/proxy-ca.pem")
```

**M2 — merge kept, `cmd.envs(&child_env)` reinstated.** CA lands; proxy regresses:
```
assertion `left == right` failed: `HTTPS_PROXY` must carry the adapter's normalised proxy URL,
not `build_child_env`'s bare `127.0.0.1:19771`
  left: Some("127.0.0.1:19771")
 right: Some("http://127.0.0.1:19771")
```

**M3 — the fixture's own honesty check.** The stub reports absent variables as `__UNSET__` via `${VAR-…}` (no colon), so "absent" and "empty" are distinguishable. Changing it to `${VAR:-…}` collapses them and the guard fires:
```
assertion `left == right` failed: a variable set to the empty string must be reported as empty,
not as `__UNSET__` — a dump that cannot tell the two apart cannot fail for the bug this file
targets.
```

### A false pass, caught

The first M1 run **passed**. `aa-integration-tests` has no `aa-cli` dependency, so nextest built nothing of the CLI and the test measured a **stale `target/debug/aasm` from before the mutation** — a false pass for a test whose entire subject is `aa-cli` behaviour. `aasm_binary()` now rebuilds unconditionally. All three mutations above are post-fix against a freshly built binary.

`RealHomeGuard` closes each test — length+mtime, contents never read.

Gates: `cargo fmt --all`; `cargo clippy -p aa-cli -p aa-integration-tests --all-targets --all-features -- -D warnings` clean; `cargo nextest run -p aa-cli -p aa-devtool-claude-code` → 1008 passed; `--test cli_run --test cli_run_claude_launch_env --test claude_code_integration_lifecycle` → 23 passed.

## Checklist

- [ ] All CI checks passing

## Found, not fixed

- `run.rs:508` (dry-run) has the same dead `cmd.envs(&child_env)` — harmless there, and outside this fix.
- **The dry-run preview is structurally wrong**: it builds its own `Command::new(&args.tool)` instead of asking the adapter, so `aasm run --dry-run` prints neither `NODE_EXTRA_CA_CERTS` nor the normalised proxy. **The preview does not describe the launch that would happen.** Separate ticket — fixing it means moving the short-circuit past `detect()`.
- `LaunchableTool::build_launch_command` (`lifecycle.rs:1351`) is the same failure class waiting to happen; no shipping `aa-cli` consumer yet.
- **`aa-devtool-contract`'s `LaunchableTool` contract never states that a returned `Command`'s environment must be honoured.** This bug was possible because that obligation lived nowhere but in the caller's memory. A doc-contract change is the durable fix; it touches a crate outside this ticket.

### DIFF
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -428,10 +428,31 @@ async fn deregister_with_gateway(registration_id: &str, ctx: &ResolvedContext) {
 
 /// Spawn `cmd` as a tokio child process, forward SIGTERM/SIGINT on Unix,
 /// and wait for the child to exit. Returns the child's exit code.
+///
+/// The child's environment is the union of `child_env` (the operator's shell
+/// environment plus this session's governance identity) and the environment the
+/// adapter set on `cmd`. The adapter is applied **last and therefore wins** on a
+/// collision: it is the layer that knows what the launched tool actually needs —
+/// `NODE_EXTRA_CA_CERTS` pointing at the proxy CA, without which the tool's
+/// runtime refuses the intercepted TLS and the session is ungoverned while
+/// looking governed — and it is the layer that normalises the gateway's bare
+/// `host:port` proxy address into the `http://host:port` URL an HTTP client
+/// accepts (AAASM-5324). Dropping the adapter's environment, as this function
+/// did before AAASM-5327, silently defeated both.
 async fn spawn_and_wait(cmd: std::process::Command, child_env: &HashMap<String, String>) -> Result<i32> {
     let mut tokio_cmd = tokio::process::Command::new(cmd.get_program());
     tokio_cmd.args(cmd.get_args());
     tokio_cmd.envs(child_env);
+    // `get_envs` yields `None` for a variable the adapter wants *removed* from
+    // the child, which is not the same request as setting it empty — an adapter
+    // that unsets a variable to disable a tool behaviour must not have that
+    // turned into an empty-but-present variable the tool then honours.
+    for (name, value) in cmd.get_envs() {
+        match value {
+            Some(value) => tokio_cmd.env(name, value),
+            None => tokio_cmd.env_remove(name),
+        };
+    }
 
     let mut child = tokio_cmd.spawn()?;
     let child_pid = child.id().unwrap_or(0) as i32;
@@ -536,7 +557,7 @@ pub async fn execute_with_adapters(
         .await
         .map_err(|e| anyhow::anyhow!("failed to apply settings: {e}"))?;
 
-    let mut cmd = adapter
+    let cmd = adapter
         .build_launch_command(
             &args.tool_args,
             &handle.agent_id,
@@ -544,7 +565,10 @@ pub async fn execute_with_adapters(
             handle.proxy_addr.as_deref(),
         )
         .map_err(|e| anyhow::anyhow!("failed to build launch command: {e}"))?;
-    cmd.envs(&child_env);
+    // No `cmd.envs(&child_env)` here: `spawn_and_wait` applies both sources with
+    // the adapter's on top, and overlaying `child_env` onto the command first
+    // would overwrite the adapter's values inside `cmd` — the merge would then
+    // faithfully carry forward the very values it is meant to override.
 
     let mut guard = RegistrationGuard {
         registration_id: handle.registration_id.clone(),
--- a/aa-integration-tests/tests/cli_run_claude_launch_env.rs
+++ b/aa-integration-tests/tests/cli_run_claude_launch_env.rs
@@ -1,0 +1,448 @@
+//! AAASM-5327 — the launch environment the adapter builds reaches the process.
+//!
+//! # What this measures, and why nothing else did
+//!
+//! `ClaudeCodeAdapter::build_launch_command` sets `NODE_EXTRA_CA_CERTS` (and a
+//! normalised `HTTPS_PROXY`) on the `Command` it returns. `spawn_and_wait` used
+//! to rebuild that command from its program and args alone, so every one of
+//! those variables was dropped on the floor. The consequence is the product's
+//! defining failure mode rather than a cosmetic one: `NODE_EXTRA_CA_CERTS` is
+//! the only mechanism by which Claude Code's embedded Node runtime trusts the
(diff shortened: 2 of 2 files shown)


Chisanan232:

(Comment rather than an approving review: GitHub refuses self-approval.)

Risk class: **product behaviour change on the governed-launch path + environment handling.** The only PR in this batch that changes what a shipped binary does, so reviewed hardest.

### The bug is real and the severity is right

`spawn_and_wait` rebuilt the command from `get_program()` + `get_args()` and never read `cmd.get_envs()`, so everything `build_launch_command` set was discarded — including `NODE_EXTRA_CA_CERTS`, the only mechanism by which Claude Code's Node runtime trusts the AASM CA. Without it `aa-proxy` cannot terminate TLS: **the tool launches, reports as governed, and is not inspected.** That is the silent-bypass class, so High is correct.

### Merge semantics are the right way round

`child_env` first, adapter overlaid on top, adapter wins on collision. Correct because the adapter is both the layer that knows the tool's actual requirement and the layer that normalises the gateway's bare `host:port` into a routable `http://host:port` (AAASM-5324). `None` from `get_envs()` is honoured as `env_remove` rather than flattened to empty — currently unexercised, since the Claude Code adapter only calls `.env()`, but it is the correct reading of the API and the wrong choice would silently turn an unset into an empty-but-present variable a tool might honour.

### Adversarial checks

- **Parent-environment leakage:** `build_child_env` seeds from `std::env::vars()` (`run.rs:266`) — the operator's whole shell. That is **pre-existing** and untouched by this diff, and since `child_env` was already applied explicitly before the fix, the merge is a superset of what the child already received. No new exposure.
- **`cmd.envs(&child_env)` removal is load-bearing, not cleanup.** It ran *before* anything could read `cmd`, overwriting the adapter's normalised proxy inside the command. Left in place alongside the merge it would have faithfully carried forward the very values the merge exists to override — which is precisely mutation M2 in the description.

### Mutation evidence is genuine, including the one that nearly wasn't

Three mutations, each quoted verbatim, each against a freshly built binary. The valuable one is the failure the author caught on themselves: **the first M1 run passed** because `aa-integration-tests` has no `aa-cli` dependency, so nextest built nothing of the CLI and the test measured a stale `target/debug/aasm` from before the mutation. A test whose entire subject is `aa-cli` behaviour, green while executing pre-mutation code. `aasm_binary()` now rebuilds unconditionally.

M3 is the fixture's own honesty check — the stub uses `${VAR-…}` (no colon) so *absent* and *empty* are distinguishable; switching it to `${VAR:-…}` collapses them and the guard fires. That is the difference between a fixture that can express the bad state and one that cannot.

### Recorded, not fixed — both now ticketed

- **AAASM-5329**: `aasm run --dry-run` builds its own `Command::new(&args.tool)` instead of asking the adapter, so the preview shows neither `NODE_EXTRA_CA_CERTS` nor the normalised proxy. The preview does not describe the launch that would happen — and the two things it omits are exactly the two whose absence means ungoverned-but-looks-governed.
- **AAASM-5330**: `LaunchableTool` never states that a returned `Command`'s environment must be honoured. That omission is *why* this bug was possible — the obligation lived only in the caller's memory — and the same failure is latent at `lifecycle.rs:1351`, waiting for its first consumer.

### Does not close AAASM-1112

This fixes one of the two defects blocking it. **AAASM-5323** — the registration endpoint that does not exist — is still open on two product decisions, and 1112 cannot pass until both are resolved and the real-tool suite is re-run.

Gates: not a draft; MERGEABLE, no conflicts; 23 pass / 18 skipping / 0 failing; 0 unresolved threads; no requested-changes review; head `05bf0b53` current.

PR #1861, opened 2026-07-31
[AAASM-5323] ✨ (aa-cli): Resolve and verify the proxy endpoint host-side, and fail closed

Chisanan232:
## Description

Implements **decision 1 of AAASM-5323** (owner-approved): the proxy address a governed launch routes at is resolved **host-side from the canonical proxy state file**, verified, and **failed closed** — never taken from the gateway, an adapter, a tool, or the ambient environment.

### The hole this closes

`aasm run` took its proxy address from `RegistrationHandle::proxy_addr`, populated only from the gateway's registration response. **Nothing in the tree ever sets that field** — `grep -rn proxy_addr` across `aa-api/`, `aa-gateway/`, `aa-proto/`, `proto/`, `openapi/` returns zero hits. And `run.rs` injected `HTTPS_PROXY`/`HTTP_PROXY` only when it was `Some`. So `aasm run claude` has **always launched completely unproxied**, while reporting as governed.

One thing worse than that: `build_child_env` seeds from `std::env::vars()`, so an **ambient `HTTPS_PROXY` from the operator's shell** was already reaching the child and would have been the de-facto route — an unverified proxy masquerading as governance.

### Design

The proxy state record (`aa-cli/src/commands/proxy/pid.rs`) is extended from `<pid>\n<listen_addr>` to four lines: `pid / listen_addr / start_token / exe_path`. New `proxy/trust.rs` resolves and verifies it; new `proxy/identity.rs` reads live-process identity natively per platform.

| Constraint | Implementation |
|---|---|
| No caller/adapter/tool/env-supplied address is authoritative | `proxy_addr` removed from `RegistrationHandle` **and** from the `RegisterResponse` deserialiser. `AA_PROXY_ADDR` is read only by `aasm proxy start` (the writer). Ambient `HTTPS_PROXY` is **overwritten** when an endpoint is vouched for and **removed** when none is |
| State file owned/controlled by the host-side integration | `symlink_metadata` (a symlink is refused, not followed), must be a regular file, `uid == geteuid()`, `mode & 0o022 == 0`. `write_state` now chmods `0600` explicitly after writing, tightening a pre-existing looser file |
| PID alive | `kill(pid, 0)` — which also means signallable by this user, so another user's process reads as not-ours |
| Stale state and **PID reuse** | `(executable image, process start time)` + a live bound socket — see below |
| Process identity is the proxy | the recorded executable's file name must be `aa-proxy`, and the live image must equal the recorded path |
| Scheme/host/port | `ip:port` **literals only** (a hostname is refused — resolution would happen in the child, outside the boundary), `is_loopback()` required, port 0 rejected, and `http://` is *synthesised* rather than read from the record |
| Fail closed | `resolve_launch_proxy` runs **before registration**, so a doomed launch never creates a session it abandons. Any error becomes `refusing to launch ungoverned: …` and exit 1. There is no fallback path |
| Diagnostics without credentials | `ProxyTrustError` carries path, pid, mode, address and a fixed reason — no env values, no file contents, no process args |

**Why `(exe, start time)` is sufficient for PID reuse.** The executable rejects the ordinary case — a recycled PID now running something else. It does *not* reject a second `aa-proxy` taking the same PID; the start time closes exactly that, because a successor cannot have started before its predecessor exited, so `(pid, start_time)` names one incarnation of one process. That is the pair `pidfd` and systemd reduce to. Read natively: Linux `/proc/<pid>/exe` + field 22 of `/proc/<pid>/stat` (split on the **last** `)`, so a space in `comm` cannot fool it); macOS `proc_pidpath` + `PROC_PIDTBSDINFO`. Tokens are platform-prefixed so records cannot collide across hosts. **A platform with no implementation returns a refusal, never a skip.** The TCP connect is the fourth leg — a record can pass everything else while the proxy died during startup.

On the writer side the start time is read back from the kernel for the spawned PID (set at fork, unchanged by exec, so race-free), while the executable is `canonicalize(resolved binary)` rather than a read-back, because the read-back *does* race the exec.

## Type of Change

- [x] ✨ New feature
- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes (describe below)

**`aasm run <tool>` now refuses to launch unless a verified `aa-proxy` is running**, or `--no-proxy` is passed. Previously it always launched.

Nothing that was genuinely governed stops working — what stops is a launch that only *appeared* governed. This is the behaviour constraint 7 asks for, but it is user-visible and is called out here rather than buried.

## Related Issues

- Related Jira ticket: AAASM-5323 (decision 1 of two; decision 2 — gRPC registration — follows in a stacked PR)

## Testing

**Twenty mutations, each applied to production code, run, and reverted.** A sample:

```
M13 — fail-open: unresolvable endpoint returns Ok(None) instead of refusing
  refusals:: ... 10 tests run: 0 passed, 10 failed, 0 skipped

M19 — child routed at the gateway-named address instead of the verified one
  assertion `left == right` failed: `HTTPS_PROXY` must carry the verified local endpoint
    left: Some("http://127.0.0.1:19771")
   right: Some("http://127.0.0.1:64559")

M6 — start-time comparison removed (PID reuse, executable-invisible)
  a recycled PID must be refused: ()

M15 — ambient proxy kept when nothing was vouched for
  an unvouched-for ambient HTTPS_PROXY must not reach the child
```

The ten-case refusal matrix was re-run independently before merge: **10 passed, 0 skipped** — no proxy running · recorded process gone · record naming a non-proxy binary · no identity evidence · nothing listening · symlinked state file · non-loopback endpoint · different executable · over-permissive file · different incarnation.

Harness notes: `proxy_trust_support::build_binary` rebuilds `aasm` **and** `aa-proxy` unconditionally, and this PR **removes** `cli_run_claude_governed_launch.rs`'s private `aasm_binary()`, which preferred an existing `target/debug/aasm` over building — that file carried the stale-binary trap that cost the AAASM-5327 work its first mutation result. The `claude` stub uses `${VAR-__UNSET__}` (no colon) so absent is distinguishable from empty, and the refusal fixture proves non-launch by the **absence of the dump file**, not by matching a string.

Gates: `cargo fmt --all -- --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean workspace-wide; `aa-cli` + `aa-devtool*` 1067 pass; `cli_run_trusted_proxy` 10/10, `cli_run_claude_launch_env` 2/2, `cli_run_claude_governed_launch` 2/2, `cli_run` 5/5.

## Checklist

- [ ] All CI checks passing

## For the reviewer to rule on

**Should `--no-proxy` survive?** It is the one remaining route to an unproxied launch. It is pre-existing, explicit, and now loud — `warning: --no-proxy — launching WITHOUT interception. This session's traffic is not inspected and no egress policy applies to it.` — and under it the operator's own `HTTPS_PROXY` is deliberately left untouched, which a corporate-proxy user opting out would need. It therefore satisfies "no silent fallback", but it is a deliberate Layer-2 bypass and whether it should exist at all is a product call.

## Found, not fixed

- **`aasm proxy start --listen 0.0.0.0:PORT` now yields a proxy `aasm run` will refuse.** Rejecting a non-loopback bind at *start* time is a separate product call; today the operator gets a working `aasm proxy status` and a refusing `aasm run`. Worth a follow-up.
- **`AA_DATA_DIR` directory permissions are not checked**, only the file's. An attacker with write access there can delete our record and plant one — but theirs would be owned by them and refused by the uid check.
- **A canonical proxy path whose basename is not literally `aa-proxy`** would be refused. Not observed on any install shape checked (`cargo install`, Homebrew, `target/debug`), so the check is left exact rather than loosened.
- **`aa-integration-tests::cli_proxy proxy_logs_follow_streams_new_entries` fails on this machine, pre-existing and unrelated** — it exercises untouched `proxy/logs.rs`, resolves from `HOME` not the state record, and its fixture sleeps a fixed 1000 ms while falling back to `cargo run`. Against a prebuilt binary the same scenario streams correctly.
- **`build_launch_command` still normalises a bare `host:port`.** Now always handed an already-normalised URL, so the branch is dead on this path — but the trait is public and shared, so left alone.

### DIFF
--- a/aa-cli/src/commands/proxy/identity.rs
+++ b/aa-cli/src/commands/proxy/identity.rs
@@ -273,6 +273,26 @@ mod tests {
         );
     }
 
+    /// A process asking about *itself* is exempt from the ptrace gate
+    /// (`same_thread_group`), so a self-read can never exercise the hidden-process
+    /// path. Pin that the ordinary case is still answered by the authoritative
+    /// source, so a regression that made every read fall through to `argv[0]`
+    /// would be visible here rather than silently downgrading every check.
+    #[test]
+    fn a_process_that_hides_nothing_is_named_by_the_kernel() {
+        let mut child = std::process::Command::new("sleep")
+            .arg("30")
+            .spawn()
+            .expect("spawn sleep");
+        let evidence = image(child.id());
+        let _ = child.kill();
+        let _ = child.wait();
+        assert!(
+            matches!(evidence, ImageEvidence::Reported(_)),
+            "an ordinary process must be named by the kernel, not by its own argv; got {evidence:?}"
+        );
+    }
+
     #[test]
     fn start_token_is_stable_across_reads_for_the_same_process() {
         let pid = std::process::id();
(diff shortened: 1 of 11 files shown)


Chisanan232:

Risk class: **product behaviour change on the governed-launch path, plus a new trust boundary.** Reviewed adversarially, and re-verified after the dependency baseline moved.

### The hole this closes was total, not partial

`aasm run` took its proxy address from `RegistrationHandle::proxy_addr`, populated only from the gateway's registration response. **Nothing in the tree ever set that field** — zero hits across `aa-api`, `aa-gateway`, `aa-proto`, `openapi/`. And the injection was gated on `Some`. So the launch has always been unproxied while reporting as governed.

Worse, and found during implementation: `build_child_env` seeds from `std::env::vars()`, so an **ambient `HTTPS_PROXY` from the operator's shell** was already reaching the child and would have been the de-facto route — an unverified proxy masquerading as governance. It is now overwritten when an endpoint is vouched for, and **removed** when none is.

### The PID-reuse evidence holds up

`(executable image, process start time)` plus a live bound socket. The executable rejects a recycled PID running something else; the start time closes the case the executable cannot — a *second* `aa-proxy` taking the same PID — because a successor cannot have started before its predecessor exited. That is the pair `pidfd` and systemd reduce to. Read natively per platform, with `/proc/<pid>/stat` split on the **last** `)` so a space in `comm` cannot fool it. A platform with no implementation returns a **refusal, never a skip**.

### The Linux break, and why the fix does not weaken anything

This branch initially broke `aasm run` on **every Linux host** — verified as mine, not inherited. Root cause was not the zombie I hypothesised: `aa-proxy` calls `prctl(PR_SET_DUMPABLE, 0)` at startup (AAASM-3584, to stop a same-uid attacker ptracing provider keys out of its address space), and Linux gates `readlink(/proc/<pid>/exe)` behind `ptrace_may_access`, which **rejects even the owning user** once the target is non-dumpable. A genuine hardening measure made the process unverifiable to its own trust check. Refuted in a Linux container as **non-root** — as root it would have been masked entirely by `CAP_SYS_PTRACE`.

The fallback to `argv[0]` is gated **only** on `PermissionDenied`. `NotFound` (a zombie) stays a refusal, any other error stays a refusal, and an empty `cmdline` (also a zombie) stays a refusal — two independent paths. I checked the gating myself rather than taking it on report.

On whether `argv[0]` weakens it: that value is attacker-controllable in general, but reaching it requires the attacker's PID **and** start-token to already be in the state file, which is uid-owned and mode 0600. An attacker who can write that is already the user. No privilege boundary is crossed, and the unforgeable `(pid, start_time)` pair still has to match. Nothing that was refused before is accepted now.

Blast radius was also **double what CI reported** — the `Test` job runs without `--no-fail-fast`, so it cancelled before `cli_run_trusted_proxy` ran: 4 reported, 8 actually broken. Filed as AAASM-5334.

### Dependency gate — resolved, and verified rather than assumed

`Dependency checks` now **passes**. RUSTSEC-2026-0222/0223 were fixed by #1862 (wasmtime 47.0.3), an externally owned lane I did not touch. I noticed the green `Dependency checks` runs on `main` all *predated* the advisories and so proved nothing, and ran `cargo deny check advisories` against the merged tree directly: **`advisories ok`**. Refreshed by merging `main`, not rebasing — no history rewritten, no force push, and no conflict because this branch never touched `Cargo.lock`.

An earlier `no_std targets (wasm32)` failure was a 45-minute hang with every step at `conclusion: null` and expired logs; that job builds `aa-core` only, which this branch does not touch. Re-run, passed.

### Behaviour change, called out rather than buried

**`aasm run` now refuses to launch unless a verified `aa-proxy` is running**, or `--no-proxy` is passed. Nothing genuinely governed stops working — what stops is a launch that only *appeared* governed.

### Two product calls left open for the maintainer

- **Should `--no-proxy` survive?** It is the one remaining route to an unproxied launch. Explicit, and now loud (`warning: --no-proxy — launching WITHOUT interception…`), and it leaves the operator's own proxy untouched, which a corporate-proxy user opting out needs. It satisfies "no silent fallback", but whether a deliberate Layer-2 bypass should exist is a product decision.
- **Should `aasm proxy start --listen 0.0.0.0:PORT` be refused at bind time?** Today it succeeds and yields a proxy `aasm run` then declines to trust.

Neither blocks this PR. Twenty mutations, each applied to production code and reverted; the ten-case refusal matrix re-run independently before merge (**10 passed, 0 skipped**).

Gates: not a draft; MERGEABLE, no conflicts; **23 pass / 18 skipping / 0 failing**; 0 unresolved threads; no requested-changes review; head `0397b762` current.

PR #1862, opened 2026-07-31
[AAASM-5335] ⬆️ (deps): Bump wasmtime 47.0.2 → 47.0.3 (RUSTSEC-2026-0222/0223, main red)

Chisanan232:
## Description

**Fixes `main` red (org-wide).** Two new RustSec advisories landed against `wasmtime 47.0.2` (transitive via `aa-sandbox` → `aa-cli`), and cargo-deny fetches the DB fresh each run — so Dependency-checks went red on `main` and every open PR with no code change.

- **RUSTSEC-2026-0222** (GHSA-hgjw-h833-99q9) — Stores can mix up type indices between engines
- **RUSTSEC-2026-0223** (GHSA-2hw9-mc66-jc2q) — Preemption/traps during bulk operations break internal VM state

Both advise **>=47.0.3** — a same-minor in-range patch. `cargo update -p wasmtime --precise 47.0.3` alone conflicts on the `=47.0.x`-pinned family members, so the whole family was updated together: `wasmtime`, `wasmtime-wasi`, `wasmtime-wasi-io`, `wasmtime-environ`, `wiggle` (+ internal siblings). **Lockfile-only** — no source change.

## Type of Change
- [x] ⬆️ Dependency upgrade (security)

## Breaking Changes
- [x] No (same-minor patch, `^47` requirement unchanged)

## Related Issues
- Related Jira ticket: AAASM-5335 (unblocks main + AAASM-5289 + all open PRs)

## Testing
- [x] No tests required (lockfile-only)

Verified locally: `cargo deny --all-features check` → advisories/bans/licenses/sources **all ok** (both RUSTSEC IDs cleared); `cargo build -p aa-sandbox` (the wasmtime consumer) clean. CI runs the authoritative full suite.

## Checklist
- [x] Reproduced the failure and verified the fix locally
- [x] Self-review of the diff completed
- [x] Commits are small and follow the Gitmoji convention

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1863, opened 2026-08-01
[AAASM-5323] ✨ (aa-cli): Register aasm run through the gRPC identity gate

Chisanan232:
## Description

Implements **decision 2 of AAASM-5323** (owner-approved): `aasm run` registers through the **existing gRPC path** via `aa-sdk-client`, enforcing the same identity and registration gate as the SDKs. No `POST /api/v1/agents` route is added and `openapi/` is untouched.

> **Stacked on #1861** (decision 1 — trusted proxy endpoint), which is merged into this branch. Merge #1861 first.

### Why not the HTTP route the CLI was calling

`aasm run` POSTed to `/api/v1/agents`, which has never existed — `routes/mod.rs:124` mounts `get(agents::list_agents)` only. Adding the handler was investigated and rejected: the CLI's body carries no public key, nonce or proof, so an HTTP registration would have bypassed `enforce_did_key_binding` (AAASM-4787, DID squatting) and `verify_possession_proof` (AAASM-3591/3866), letting any API-key holder register under an arbitrary DID.

### Constraints, each satisfied

| Constraint | How |
|---|---|
| No CLI-only bypass | `RequestChallenge` → sign → `Register` over the same client. A test asserts the HTTP surface still offers no registration route |
| One identity contract | `build_session_request` delegates `agent_id` / `public_key` / `registration_nonce` / `possession_proof` to `aa_sdk_client::gateway::build_register_request` and sets only descriptive fields. Proven by byte-identical proofs, and at the gateway by the SDK colliding with the CLI's registration (`AlreadyExists`) rather than creating a second record |
| Freshness / replay | Fresh challenge per `register()`, no nonce ever persisted. A replay of the CLI's own captured bytes is refused `Unauthenticated` |
| Fail closed | `register_with_gateway(...).await?` runs **before** `build_launch_command`; every failure is `refusing to launch unregistered: <reason>` and exit 1 |
| Equivalent outcomes | Asserted in both directions |
| Cross-surface conformance | `run_registration_gateway.rs` runs the genuine `AgentLifecycleServiceImpl`, captures the CLI's **literal wire bytes**, then degrades one property and resubmits |

### Identity correlation

`AA_AGENT_ID` → derived key → `AA_AGENT_DID` (registered, audit-attributed) → `SHA256(org/team/did)[..16]` → `AA_REGISTRATION_ID`. All three reach the child so anything downstream can re-derive the chain. The credential token is held privately for `Deregister` and **never** exported — `--dry-run` prints the whole env map.

### Response fields, decided rather than carried forward

`registration_id` becomes the hex of the gateway's own `proto_agent_id_to_key` — real server-side identity, and the only shape `GET`/`DELETE /api/v1/agents/{id}` parse, which also fixes a guaranteed 400. `trace_id` and `session_id` remain locally minted **and are now documented as such** instead of pretending to be server-issued. Deregistration moves to gRPC `Deregister`, because the REST `DELETE` authenticates as the *operator* — a different principal from the one that registered.

## Type of Change

- [x] ✨ New feature
- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes (describe below)

- `--root-agent X` now registers X as the parent and **fails closed if X is unregistered**, where it was previously ignored entirely.
- `--observe` no longer weakens gateway-side enforcement: `authoritative_enforcement_mode` (AAASM-4121) drops weakening claims from self-registering agents. Pre-existing gateway policy, newly reachable because the CLI now genuinely registers.

## Related Issues

- Related Jira ticket: AAASM-5323 (decision 2 of two) — unblocks AC4 of AAASM-201, tracked by AAASM-1112
- Depends on #1861

## Testing

**14 mutations, 14 caught.** Highlights — gateway stops verifying the proof; nonce made reusable; DID↔key binding dropped; fail-closed removed; CLI derives its own identity; CLI caches its nonce; proof signs the DID instead of the nonce; credential token leaked into the child env.

**One mutation exposed a defect in the tests themselves.** M9's first run left every unit test green while the real gateway refused — the unit helper rebuilt the `RegisterRequest` instead of calling the production path, i.e. a fixture that could not express the bad state. Fixed in `ae5ce866` by extracting `build_session_request` as the single builder both use; M9 then failed as it should. That is the reason these were run rather than asserted.

**Reported rather than hidden:** `run_command_refuses_to_launch_when_registration_is_impossible` *passed* under the fail-closed mutation, because an unreachable gateway fails at `connect` — a path that mutation did not touch. The gateway-refuses-us path is covered only by `run_command_refuses_a_lineage_the_gateway_will_not_accept`, which is why that test exists.

Gates: `cargo fmt --all` clean; `cargo clippy --workspace --all-targets --all-features -D warnings` clean; `aa-cli` + `aa-sdk-client` 1000/1000; the six affected integration binaries 46/46. Against the merged tree: `run_registration_gateway` + `run_command` **10/10**.

**Publish boundary:** `aa-sdk-client` is `publish = false`, so the dependency sits in a new `sdkclient` strip region and `run_registration.rs` is in `.ci/strip-for-publish.sh`'s `DELETED_FILES`. Verified on a stripped copy: dep, `mod` and file all removed, `cargo check -p aa-cli` passes, `scripts/check-publish-surface.sh` OK.

## Checklist

- [ ] All CI checks passing

## Sequencing note — AAASM-5332

This PR uses `AgentKeypair::derive`, which seeds Ed25519 with `SHA-256(agent_id)` — a **public** value. The possession proof therefore establishes that the caller can compute SHA-256, not that it holds a secret. That is **pre-existing and shared with all three SDKs**; this PR inherits it rather than introducing it, and using the shared path is still strictly better than a second, weaker contract.

**AAASM-5332** now carries the approved direction — registration secrets become bootstrap/enrollment credentials only, with a distinct Ed25519/DID identity key — which supersedes this key source. Whether this lands first and 5332 builds on it, or this waits, is a sequencing call for the maintainer. Holding it does not reduce exposure by one bit (the same derivation is live in every SDK today) but it does delay the only thing that closes AC4.

## Found, not fixed

- **`aasm run` never heartbeats** — `heartbeat_interval_sec` is discarded. A long run under a team with a configured max age would be force-deregistered mid-session while the tool keeps running.
- **`AA_GATEWAY_ENDPOINT` gets no trust verification**, unlike the proxy endpoint. A hostile value points registration at a fake gateway and the launch then proceeds "registered". Identical to the SDK's exposure, not introduced here — worth its own ticket.
- **`assigned_policy` is discarded**; the CLI still uses the hard-coded `load_policy()` stub.


Chisanan232:

Risk class: **identity and authorization on the registration path.** The highest-risk PR in this wave, reviewed accordingly.

### What it does, and why not the obvious alternative

`aasm run` POSTed to `/api/v1/agents`, a route that has never existed — `routes/mod.rs:124` mounts `get(agents::list_agents)` only. Adding the handler was investigated and **rejected**: the CLI's body carries no public key, nonce or proof, so an HTTP registration would have bypassed `enforce_did_key_binding` (AAASM-4787, DID squatting) and `verify_possession_proof` (AAASM-3591/3866), letting any API-key holder register under an arbitrary DID.

Routing through the existing gRPC path means one gate for every surface rather than a second, weaker one.

### The conformance tests attempt real bypasses

This is the constraint most easily satisfied vacuously — a test that "proves" one surface cannot bypass another by never attempting it is worthless. These run the genuine `AgentLifecycleServiceImpl`, capture the CLI's **literal wire bytes**, then degrade one property and resubmit: proof blanked with a fresh nonce; verbatim replay; victim DID under an attacker key with a valid proof for that key. Each is refused.

**14 mutations, 14 caught** — gateway stops verifying the proof; nonce made reusable; DID↔key binding dropped; fail-closed removed; CLI derives its own identity; CLI caches its nonce; proof signs the DID instead of the nonce; credential token leaked into the child env.

**One mutation exposed a defect in the tests themselves**, disclosed rather than hidden: M9's first run left every unit test green while the real gateway refused, because the unit helper rebuilt the `RegisterRequest` instead of calling the production path — a fixture that could not express the bad state. Fixed by extracting `build_session_request` as the single builder both use; M9 then failed as it should.

Also reported rather than buried: `run_command_refuses_to_launch_when_registration_is_impossible` *passed* under the fail-closed mutation, because an unreachable gateway fails at `connect`, a path that mutation did not touch. That is why `run_command_refuses_a_lineage_the_gateway_will_not_accept` exists.

### CodeQL — fixed, not suppressed

The first run reported **4 criticals**: `Hard-coded cryptographic value … used as a nonce`. All four were `#[cfg(test)]` fixtures, and production takes its nonce from `client.request_challenge(...)` → `build_session_request(&desc, &challenge.nonce)` (`run_registration.rs:251-259`) — I verified both halves, because "the alerts are in tests" is the reasoning that ships vulnerabilities.

Resolved by removing the literals (a test-only generator), not by suppressing the rule. A suppression was the cheaper edit and the worse outcome: it teaches the next reviewer that this rule is noise *on the file handling Ed25519 possession proofs*, where a hard-coded nonce would be a genuine critical. CodeQL now reports **"No new alerts in code changed by this pull request."**

### Dependency lane — verified untouched immediately before merge

**0 wasmtime lines added or removed** by this branch; wasmtime `47.0.3` identical on branch and `main`. The branch's `Cargo.lock` delta is four lines in `aa-cli`'s dependency list — `aa-sdk-client`, `ed25519-dalek 3.0.0`, `tokio-stream`, `tonic` — the mechanical consequence of gaining the gRPC client. No overlap with PR #1862's lane, which is also why merging `main` in was conflict-free.

### Behaviour changes, called out

- **`--root-agent X` now fails closed if X is unregistered**, where it was previously ignored outright. Right semantics for an unverifiable lineage claim, but a break.
- **`--observe` no longer weakens gateway-side enforcement.** `authoritative_enforcement_mode` (AAASM-4121) drops weakening claims from self-registering agents — pre-existing policy that becomes reachable for the first time *because* the CLI now genuinely registers.

### The known weakness this inherits, and does not introduce

`AgentKeypair::derive` seeds Ed25519 with `SHA-256(agent_id)` — a **public** value — so the possession proof establishes that the caller can compute SHA-256, not that it holds a secret. That is live in all three SDKs today; this PR uses the shared path rather than inventing a second, weaker one, which is strictly better.

**AAASM-5332** carries the approved replacement (bootstrap/enrolment credentials only, with a distinct identity key) and supersedes this key source. Merging here does not increase exposure by one bit, and it unblocks AC4 of AAASM-201.

### Does not close AAASM-1112

AC4 is not verified by merging this. That requires the real-tool conformance suite rerun against the resulting `main` and passing — a rerun that now means more than it did, since #1865 made a run that measures nothing incapable of reporting success.

Gates: not a draft; MERGEABLE, no conflicts; **32 pass / 26 skipping / 0 failing**; 0 unresolved threads; no requested-changes review; head `2d20e2d4` current.

PR #1864, opened 2026-08-01
[AAASM-5330] 📝 (aa-core): Make the launch-environment obligation part of the contract

Chisanan232:
## Description

Implements **AAASM-5330**: the obligation that a caller must honour the environment on a returned launch `Command` is now stated in the contract and **machine-checked**, instead of living only in the caller's memory.

That omission is what made AAASM-5327 possible. `spawn_and_wait` rebuilt the command from `get_program()` + `get_args()`, discarded the whole environment — `NODE_EXTRA_CA_CERTS` included — and violated nothing written down. Without that variable the proxy cannot terminate TLS, so the tool launched, reported as governed, and was inspected by nothing.

### The ticket named the wrong trait

Worth stating plainly, because following the AC literally would have left the real path uncovered: **`aa-cli` never touches `LaunchableTool`.** `execute_with_adapters` resolves `Box<dyn DevToolAdapter>` (`aa-cli/src/commands/run.rs:403`) and calls `aa-core/src/dev_tool.rs:290` — which had the *identical* documentation gap.

Both traits are now documented. `LaunchableTool` carries the long-form rationale; `DevToolAdapter` states the same obligation and links to it.

### What the contract now says

- The returned command's environment **must** reach the spawned child.
- `None` is a **removal**, never "set to empty". Flattening the two is itself security-relevant: an adapter that unsets a variable to switch a tool behaviour off must not have that become an empty-but-present variable the tool then honours.
- On collision the **adapter wins against the caller's ambient environment** — it is the layer that knows the tool's requirements and the layer that normalises the gateway's bare `host:port`.
- But **`LaunchSpec::env` wins against the adapter**, because it is an argument rather than a competing source. These are two different axes, and the first wording could be read as conflating them; that was corrected after reviewing `lifecycle.rs` against it.

## Type of Change

- [x] 📝 Documentation update
- [x] ✅ Tests

## Breaking Changes

- [x] No — documentation and a new test only. No behaviour changes.

## Related Issues

- Related Jira ticket: AAASM-5330
- Root cause of AAASM-5327 (fixed in #1855); the latent second site is `aa-devtool-claude-code/src/lifecycle.rs:1351`

## Testing

New `aa-devtool-contract/tests/launch_environment_contract.rs` — 6 tests. A fixture adapter sets `NODE_EXTRA_CA_CERTS`, normalises the proxy address, and `env_remove`s a variable. The harness plays the caller and **really spawns `/usr/bin/env`**, asserting on what the process actually saw. That matters: inspecting `get_envs()` would only read the caller's own instructions back, and could not tell a removed variable from an inherited one.

**Mutation — the `Conforming` arm replaced with `{}` so the harness drops the environment:**

```
assertion `left == right` failed: the CA variable is the only thing that makes the tool's
runtime trust the proxy; without it the launch is inspected by nothing
  left: Some("/etc/ssl/certs/ca-bundle.crt")
 right: Some("/home/dev/.aa/ca/aasm-ca.pem")

assertion `left == right` failed: the adapter unset this to switch a tool behaviour off
  left: Some("1")
 right: None

5 tests run: 2 passed, 3 failed
```

Reverted; suite green (`5 passed, 0 skipped`).

**Read that first failure carefully.** The violating caller does not produce an *absent* variable — it produces a **plausible-looking** one: the OS certificate bundle instead of the proxy CA. That is exactly why AAASM-5327 survived review. A test asserting only "`NODE_EXTRA_CA_CERTS` is set" would pass against the bug.

The two violating behaviours also ship as **permanent `CallerBehaviour` variants** with their own assertions, so every run proves the suite can still distinguish a conforming caller from a violating one — not just that the happy path works.

Gates: `cargo fmt --all` clean; `cargo clippy -p aa-devtool-contract -p aa-core --all-targets --all-features -- -D warnings` exits 0; `cargo nextest run -p aa-core -p aa-devtool-contract -p aa-devtool -p aa-devtool-claude-code -p aa-cli` → **1461 passed, 0 skipped**.

## Checklist

- [ ] All CI checks passing

## `lifecycle.rs` reviewed, not changed

`aa-devtool-claude-code/src/lifecycle.rs:1345-1381` conforms as a *producer*: it returns an env-bearing command (`installed_environment` supplies `NODE_EXTRA_CA_CERTS`; the proxy is normalised to `http://` form), it applies `spec.env` **last** as the newly-stated precedence requires, and it removes nothing, so caller variables it does not mention pass through. No change made — reviewing it was in scope, editing it was not.

It has **zero shipping consumers** today (`as_launchable()` resolves only to test callers), which is why the identical bug is latent rather than live there.

## Found, not fixed

**No shipping code ever emits a `None` env value.** `installed_environment` returns `BTreeMap<String, String>`, so a variable an operator unset simply vanishes from the map rather than becoming a removal instruction — and an ambient value therefore still reaches the child. That is not a contract violation as written (a variable the adapter never mentions is passed through by design), but it means the `None`-means-removal clause is currently exercised only by this test fixture. Worth its own ticket rather than widening this one.

Also: two pre-existing rustdoc warnings in `aa-core/src/evaluators.rs:3` (unresolved links to `DenyAllEvaluator` / `PermitAllEvaluator`), identical on `main`, untouched.

The test is `#![cfg(unix)]` because it reads the child environment out of `/usr/bin/env`; this workspace's CI has no Windows target, so nothing is silently skipped.

### DIFF
--- a/aa-core/src/dev_tool.rs
+++ b/aa-core/src/dev_tool.rs
@@ -287,6 +287,38 @@ pub trait DevToolAdapter: Send + Sync {
     ///   the wiring.
     /// * Sync (no I/O performed) — the returned `Command` is *built*,
     ///   not spawned. Spawning is the launcher's job.
+    ///
+    /// ### The environment on the returned command is part of the contract
+    ///
+    /// The launcher **must** apply [`Command::get_envs`] to the child it
+    /// spawns. The environment is not advisory decoration on top of a
+    /// program and its arguments: `NODE_EXTRA_CA_CERTS` is the only
+    /// mechanism by which an Electron/Node tool's runtime trusts the
+    /// intercepting proxy's CA, so a launcher that rebuilds the child from
+    /// [`Command::get_program`] and [`Command::get_args`] alone starts a
+    /// tool the proxy cannot terminate TLS for and reports it as governed.
+    /// That is AAASM-5327, and it violated nothing that had been written
+    /// down — which is why it is written down here.
+    ///
+    /// A `None` value from [`Command::get_envs`] means **remove that
+    /// variable from the child**, never "set it to the empty string": an
+    /// adapter unsets a variable to switch a tool behaviour off, and an
+    /// empty-but-present variable is one the tool may still honour.
+    ///
+    /// The child's environment is the union of the launcher's own
+    /// environment and this one, and on a collision **the adapter's value
+    /// wins** — it is the layer that knows the tool's requirements and the
+    /// layer that normalises values (the gateway's bare `host:port` proxy
+    /// address is not a proxy URL an HTTP client accepts). An adapter may
+    /// rely on that, but must not assume it is the sole source of the
+    /// child's variables.
+    ///
+    /// The same obligation is stated at length on the lifecycle successor,
+    /// [`LaunchableTool::build_launch_command`](crate::integration::LaunchableTool::build_launch_command).
+    ///
+    /// [`Command::get_envs`]: std::process::Command::get_envs
+    /// [`Command::get_program`]: std::process::Command::get_program
+    /// [`Command::get_args`]: std::process::Command::get_args
     fn build_launch_command(
         &self,
         tool_args: &[String],
--- a/aa-core/src/integration/contract.rs
+++ b/aa-core/src/integration/contract.rs
@@ -218,6 +218,42 @@ pub trait LaunchableTool: Send + Sync {
     /// [`AdapterError::LaunchFailed`] only for genuine run-time failures (the
     /// binary has moved, an argument cannot be encoded) — "this tool has no
     /// launch command" is a capability declaration, not an error.
+    ///
+    /// # The environment on the returned command is part of the contract
+    ///
+    /// A caller **must** apply [`Command::get_envs`] to the child it spawns.
+    /// What comes back is not "a program and some arguments, plus some advisory
+    /// variables": for an Electron/Node tool `NODE_EXTRA_CA_CERTS` is the *only*
+    /// mechanism by which the tool's runtime trusts the intercepting proxy's CA,
+    /// so a caller that drops the environment launches a tool the proxy cannot
+    /// terminate TLS for — and reports it as governed. AAASM-5327 was exactly
+    /// that: a caller that rebuilt the child from [`Command::get_program`] and
+    /// [`Command::get_args`] alone, violating no contract on paper because none
+    /// was written down. The obligation is stated here so the next occurrence is
+    /// a contract violation rather than an oversight.
+    ///
+    /// A `None` value from [`Command::get_envs`] means **remove that variable
+    /// from the child**, never "set it to the empty string". The two are not
+    /// interchangeable, and flattening them is itself a security bug: an adapter
+    /// unsets a variable to switch a tool behaviour *off*, and an
+    /// empty-but-present variable is one the tool may still read and honour.
+    ///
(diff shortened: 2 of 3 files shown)


Chisanan232:
## Maintainer review — posted AFTER merge, which was my error

This review should have been posted before merging. A shell quoting fault killed the `gh pr comment` call while the `gh pr merge` in the same command still ran, so the merge landed (`abd51a17`) with no review record. The gates were all verified beforehand — the record of that verification simply failed to post. Recording it now rather than leaving the merge unexplained.

Risk class: **public contract**. No behaviour change — documentation plus one test — but the highest-leverage item in this wave, because it converts an obligation that lived only in a caller's memory into something CI enforces.

### The ticket named the wrong trait, and the implementation said so

Mine to own: I wrote the AC naming `LaunchableTool`. **`aa-cli` never touches it.** `execute_with_adapters` resolves `Box<dyn DevToolAdapter>` (`run.rs:403`) and calls `aa-core/src/dev_tool.rs:290`, which had the *identical* omission. Following my AC literally would have left the trait that actually caused AAASM-5327 undocumented while looking satisfied. Both are now covered, and the expansion was disclosed rather than slipped in. I verified the trait claim independently.

### The mutation output is the review

```
assertion `left == right` failed: the CA variable is the only thing that makes the tool's
runtime trust the proxy; without it the launch is inspected by nothing
  left: Some("/etc/ssl/certs/ca-bundle.crt")
 right: Some("/home/dev/.aa/ca/aasm-ca.pem")
```

The violating caller does not produce an **absent** variable — it produces a **plausible** one: the OS certificate bundle instead of the proxy CA. That is precisely why AAASM-5327 survived review. A test asserting only "`NODE_EXTRA_CA_CERTS` is set" passes against the bug.

Two choices made unprompted that I would otherwise have insisted on: the harness **really spawns `/usr/bin/env`** rather than inspecting `get_envs()` — reading the builder back only replays the caller's own instructions and cannot distinguish a removed variable from an inherited one. And the two violating behaviours ship as **permanent test variants**, so every run proves the suite can still tell conforming from violating, not merely that the happy path works.

### Contract content

The environment must reach the child. `None` is a **removal**, never empty — flattening them is security-relevant, since an adapter unsetting a variable to disable a behaviour must not have it become an empty-but-present value the tool honours. The **adapter wins against the caller's ambient environment**, but **`LaunchSpec::env` wins against the adapter**, because it is an argument rather than a competing source. Those two axes were conflatable in the first wording and were separated after reviewing `lifecycle.rs` against it.

`lifecycle.rs` was reviewed and **not changed**: it conforms as a producer, applies `spec.env` last, and removes nothing. It has zero shipping consumers today, which is why the same bug is latent rather than live there.

### Honest gap disclosed by the implementation

**No shipping code emits a `None` env value** — `installed_environment` returns `BTreeMap<String, String>`, so an operator-unset variable vanishes from the map rather than becoming a removal instruction. The `None`-means-removal clause is exercised only by the fixture. Correctly left for its own ticket rather than widening this one.

### Gates, as verified before the merge

Not a draft; no conflicts; **25 pass / 19 skipping / 0 failing**; 0 unresolved review threads; no requested-changes review; head `05e03755` current at merge time. `Cargo.lock` untouched by the branch — zero commits touch it; it differed from `main` only because `main` had advanced with #1862, which is a different question from whether this branch edited it.

Merged as a merge commit (`abd51a17`, 2 parents), never squash or rebase. Admin used solely for the self-approval refusal.

PR #1865, opened 2026-08-01
[AAASM-5326] 🐛 (conformance): Fail the real-tool lane when it measures nothing

Chisanan232:
## Description

Implements **AAASM-5326**: the real-binary conformance scenario **fails when it measures nothing**, instead of printing "this is not a pass" and then returning a pass.

### The defect

After `require_claude` and `require_macos` had both been satisfied — i.e. the harness had committed to measuring — zero observed traffic printed an accurate `NOT MEASURED` message and then `return Ok(())`. nextest counted it as a pass, and nothing asserted on stdout. Compounding it, the lane carried job-level `continue-on-error: true`, so even a hard failure could not have changed the run's conclusion.

### Fix, and a second hole it exposed

The zero-traffic path now `bail!`s. Justification is the ticket's own: past the two legitimate opt-outs, zero traffic is a *failed measurement*, not an opt-out. The repo-wide `retries = 2` already absorbs non-deterministic flakes and reports recovered tests as FLAKY, so this is only reached deterministically.

Separately — and this is the part the ticket only gestured at — `require_claude` / `require_macos` also `return Ok(())`, so **a skip and a measurement were indistinguishable**. In the real-tool lane, which provisions the binary itself, a skip is a *broken lane*, not an opt-out.

A machine-readable ledger (`measured` / `skipped` / `not_measured`) is now written per scenario, and the recording lives **inside the `require_*` helpers** so no future opt-out path can forget to declare itself. Writing is opt-in via `AA_CONFORMANCE_OUTCOME_DIR`; a local `cargo nextest run` is byte-for-byte unchanged and leaves nothing behind. Every lane renders the ledger into the job summary; the real-tool lane additionally asserts `outcome == "measured"`.

**No scenario's tolerance was widened.** The only assertion change is `return Ok(())` → `bail!`.

### `continue-on-error` narrowed to the step that earns it

Removed from the job, applied to `npm install -g @anthropic-ai/claude-code` alone. The flag's stated justification is a third-party package the repo does not control — that is one step, not the whole job, and job-level scope also swallowed failure of the single assertion the lane exists to make.

If npm fails, the lane emits a `::warning::` and a summary line saying it did not run and took no measurement, and the remaining steps are guarded off. That is an honest absence of evidence rather than a green measurement. Everything after that step is the product, and fails the job.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] No

## Related Issues

- Related Jira ticket: AAASM-5326

## Testing

**Forced zero traffic** via a binary that satisfies both preconditions and emits nothing (`AA_SPIKE_CLAUDE_BIN=/usr/bin/true`):

```
MEASURED real-binary requests reaching the provider: 0
Error: NOT MEASURED [real-tool lane]: ... nothing about the tool was established, so the
lane that exists to establish it has failed.
Summary  1 test run: 0 passed, 1 failed, 25 skipped
```

It burned all three nextest attempts — deterministic, as intended.

The workflow's assertion step, extracted from the YAML and run verbatim, goes red on all three evidence-free states — `not_measured`, `skipped`, and an **empty ledger** (the lane ran but recorded nothing). The `skipped` case was also falsified end to end against the real binary, which is the "also in scope" hole: nextest reports `1 passed` having measured nothing, and the ledger assertion catches it.

**Normal run still measures and passes** — real `claude 2.1.220`:

```
MEASURED real-binary requests reaching the provider: 3
MEASURED real-binary bodies carrying the placeholder: 2 of 3
OUTCOME [real-tool lane]: measured — 3 request(s) observed, 2 of 3 carried the redaction placeholder
Summary  26 tests run: 26 passed, 0 skipped
```

Gates: `cargo fmt --all -- --check` clean; `cargo clippy -p aa-integration-tests --all-targets --all-features -- -D warnings` → 0 errors.

## Checklist

- [ ] All CI checks passing

## Notes

- **The lane remains `workflow_dispatch`-only.** Deliberate: making it run on every PR puts a third-party npm dependency in the merge path. It now *can* fail when dispatched, which is what the ticket asked. Whether to schedule it (e.g. nightly, so tool-version drift is caught without someone remembering) is a separate decision.
- A 2 KiB tail of the launch's stdout/stderr is added to the failure path, with the synthetic secret masked — it matches `sk-ant-`, and printing it on every failing CI run would trip secret scanners. Added diagnostics, not changed tolerance; without it a zero-traffic failure is undiagnosable because the child is reaped.
- `spike_support::require_claude` / `require_macos` have the same skip-counts-as-pass shape and are used by the superseded Spike suite. Untouched — out of scope, but worth extending the ledger there if that suite stays.

### DIFF
--- a/.github/workflows/claude-code-conformance.yml
+++ b/.github/workflows/claude-code-conformance.yml
@@ -26,20 +26,34 @@ permissions:
 #    `claude` binary and points `AA_SPIKE_CLAUDE_BIN` at it, so the one
 #    assertion a mock cannot make (does Claude Code's embedded Node runtime
 #    accept the Agent Assembly certificate authority through
-#    `NODE_EXTRA_CA_CERTS`) is actually taken. Not required: it depends on a
-#    third-party npm package whose availability this repository does not
-#    control. It uses no credentials and reaches no external model provider —
-#    every request is redirected to an in-process mock.
+#    `NODE_EXTRA_CA_CERTS`) is actually taken. Not required for a merge, but it
+#    can fail: see below. It uses no credentials and reaches no external model
+#    provider — every request is redirected to an in-process mock.
 #
 # The security-regression assertions are not a separate lane. They live in the
 # same binary as the scenarios they guard, because a regression set that can be
 # skipped independently of the suite it protects will eventually be skipped.
 #
-# Skipping vs passing
-# -------------------
+# Skipping vs passing vs measuring
+# --------------------------------
 # Scenarios that need the real binary print `SKIP [...]` with the reason and
 # return. `--no-capture` is deliberate: a skip must be legible in the log rather
-# than indistinguishable from a green assertion.
+# than indistinguishable from a green assertion. But a runner counts a skip as a
+# pass, so the summary line alone cannot tell a lane that measured from one that
+# declined to — which is why every lane sets `AA_CONFORMANCE_OUTCOME_DIR` and
+# renders the resulting ledger into the job summary, and why the real-tool lane
+# asserts its scenario recorded `measured`.
+#
+# Where the tolerance lives
+# -------------------------
+# This lane used to carry a job-level `continue-on-error`, on the grounds that a
+# third-party npm package this repository does not control must not read as a
+# product regression. That tolerance was far too wide: it also swallowed a
+# genuine failure of the one assertion the lane exists to make, leaving the run
+# green either way. The tolerance now sits on the single step that depends on
+# the third party. If that step fails the lane reports, loudly, that it did not
+# run — an honest absence of evidence. Everything after it is the product, and
+# fails the job.
 
 on:
   push:
@@ -94,6 +108,8 @@ jobs:
             os: ubuntu-latest
           - lane: macos
             os: macos-latest
+    env:
+      AA_CONFORMANCE_OUTCOME_DIR: ${{ runner.temp }}/conformance-outcomes
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


Chisanan232:

Risk class: **CI enforcement and evidence integrity.** The thing to hunt for here is a fix that changes the message without changing what actually gates.

### The defect was a contradiction in one function

Past `require_claude` and `require_macos` — i.e. once the harness had committed to measuring — zero observed traffic printed an accurate `NOT MEASURED … this is a gap in the evidence, not a pass` and then `return Ok(())`. nextest counted it a pass. Nothing asserted on stdout. Compounding it, job-level `continue-on-error: true` meant even a hard failure could not have changed the run's conclusion.

The fix `bail!`s. Correct: past the two legitimate opt-outs, zero traffic is a **failed measurement**, not an opt-out. Repo-wide `retries = 2` already absorbs non-deterministic flakes and reports recovered tests as FLAKY, so this is only reached deterministically — confirmed by the falsification burning all three attempts.

### The second hole is the better find

`require_claude` / `require_macos` also `return Ok(())`, so **a skip and a measurement were indistinguishable**. In the real-tool lane — which provisions the binary itself — a skip is a *broken lane*, not an opt-out, and nothing could tell the difference.

The ledger (`measured` / `skipped` / `not_measured`) is recorded **inside the `require_*` helpers**, so no future opt-out path can forget to declare itself. That is the right place: a ledger written at the call sites would have exactly the property being fixed — correct today, silently incomplete the moment someone adds a fourth early return.

Writing is opt-in via `AA_CONFORMANCE_OUTCOME_DIR`, so a local `cargo nextest run` is byte-for-byte unchanged and leaves nothing behind.

### `continue-on-error` narrowed rather than removed

Moved from the job to the `npm install` step alone. The flag's stated justification is a third-party package the repo does not control — that is one step, not the whole job, and job-level scope was also swallowing failure of the single assertion the lane exists to make. If npm fails now, the lane emits a warning saying it took no measurement and guards the remaining steps off: an honest absence of evidence rather than a green tick. Everything after that step is the product and fails the job.

### Falsified four ways, including the one that mattered

Zero traffic forced via a binary satisfying both preconditions and emitting nothing (`AA_SPIKE_CLAUDE_BIN=/usr/bin/true`) — test goes red, all three attempts. The workflow's assertion step, extracted from the YAML and run verbatim, goes red on `not_measured`, on `skipped`, and on an **empty ledger** (the lane ran and recorded nothing). The `skipped` case was falsified end to end against the real binary, which is the original hole: nextest prints `1 passed` having measured nothing, and the ledger assertion catches it.

Normal run still measures and passes — real `claude 2.1.220`, 3 requests, 2 of 3 carrying the placeholder, 26/26.

**No scenario's tolerance was widened.** The only assertion change is `return Ok(())` → `bail!`.

### Judgement calls I agree with

- **Lane stays `workflow_dispatch`-only.** Making it run per-PR would put a third-party npm dependency in the merge path. It now *can* fail when dispatched, which is what the ticket asked. Whether to schedule it nightly so tool-version drift is caught without someone remembering is a separate decision worth taking.
- **A 2 KiB stdout/stderr tail added to the failure path, with the synthetic secret masked** — it matches `sk-ant-`, and printing it on every failing CI run would trip secret scanners. Added diagnostics, not changed tolerance; without it a zero-traffic failure is undiagnosable because the child is reaped.

### Disclosed, not fixed

`spike_support::require_claude` / `require_macos` have the same skip-counts-as-pass shape and serve the superseded Spike suite. Untouched — out of scope, correctly flagged rather than silently widened.

Gates: not a draft; no conflicts; **23 pass / 18 skipping / 0 failing**; 0 unresolved threads; no requested-changes review; head `220d6216` current. `Cargo.lock` untouched by the branch — zero commits touch it; it differs from `main` only because `main` advanced with #1862.

PR #1872, opened 2026-08-01
[AAASM-5340] ✨ agents: Cascade enforcement-mode preview + echo-back apply

Chisanan232:
## What
Extends the single-agent enforcement-mode toggle (AAASM-5338) to a whole subtree (shadow-mode 3/4, [ADR-0021](https://lightning-dust-mite.atlassian.net/browse/ADR-0021) Option B):

- **`POST /api/v1/agents/{id}/enforcement-mode/preview`** — dry-run: returns the explicit affected-agent set (root + descendants, BFS order) + count, mutating nothing.
- **Apply** — `POST …/enforcement-mode` with an optional `cascade: { expected_ids, expected_count }` confirmation. The handler recomputes the current subtree and rejects a mismatch, then applies the mode to every agent in the set.

## Why
A subtree-wide fail-open toggle is high blast-radius. The preview + echo-back is the TOCTOU / mis-click guard: an operator sees exactly what will change, and the apply refuses if the tree drifted between preview and commit.

## Security invariants (enforced as a unit)
- **Echo-back**: recomputed affected set (order-independent `HashSet` compare) + count must match what was previewed → **409 Conflict** on any mismatch (stale set — re-preview).
- **`MAX_CASCADE_AGENTS = 50`**: a set exceeding 50 is rejected **422 outright, never truncated** (both preview and apply).
- **Direction-asymmetric authz + expiry applied set-wide** (resolved once): strengthen → Write; weaken → Admin + non-empty reason + future `expires_at` ≤ 72h (403/422). `Disabled` unrepresentable.
- **Tenant confinement on every node**: an out-of-tenant subtree agent → **403, never silently dropped** (partial-cascade hazard). Admin may act on any.
- **Per-agent audit**: each affected agent gets its own `GovernanceMutation` record with actor + tenant from the authenticated caller, never the body.
- Single-agent path (no `cascade` field) is byte-identical to AAASM-5338 (untagged response enum) — regression-tested.

## How to verify
`cargo test -p aa-api --lib routes::agents` — 8 new tests: preview-no-mutation, cascade-apply-mutates-all + per-agent audit (N=3), wrong-id-set → 409, wrong-count → 409, >50 → 422 (51-node fixture, both paths), weaken requires Admin+reason+expiry (403/422/422), cross-tenant subtree → 403, single-agent path unchanged.

OpenAPI regenerated (path count **92 → 93**); contract test bumped; dashboard `schema.d.ts` regenerated (`tsc --noEmit` clean). Exact CI clippy + fmt clean. Rebased on main after AAASM-5339 merge (no conflicts).

Closes AAASM-5340.


Chisanan232 (commented):
**Claude Code — self-review (AAASM-5340, shadow-mode 3/4)**

Reviewed the cascade handler against the ADR-0021 blast-radius contract:
- ✅ **Echo-back** (`apply_enforcement_cascade`): recomputes the current subtree, compares against the caller's `expected_ids` as an order-independent `HashSet` **and** checks `expected_count` — any drift → **409 Conflict** (stale set, re-preview). Correct TOCTOU/mis-click guard.
- ✅ **Validation resolved once** for the whole set (`resolve_enforcement_transition`): direction, Admin gate, non-empty reason, future+≤72h expiry — then the durable write + own audit loops per agent. No per-agent authz divergence.
- ✅ **`build_cascade_set`** authorizes the root, then confines **every** descendant to the caller's tenant — an out-of-tenant node is a hard **403, never dropped** (prevents partial cross-tenant cascade). Admin bypasses per-node but not the cap.
- ✅ **`MAX_CASCADE_AGENTS = 50`** rejects **422 outright, never truncates**, on both preview and apply (51-node fixture test).
- ✅ Per-agent `GovernanceMutation` audit with actor+tenant from the verified caller, never the body (asserted N=3).
- ✅ Single-agent path untouched — untagged `EnforcementModeApplyResponse::Single` keeps the 5338 wire shape byte-identical; regression test present.
- ✅ Preview gated at the `Write` floor (not `Read`) so a read-only caller can't probe a subtree it can't toggle; direction-agnostic so no Admin gate on the dry-run.
- ✅ OpenAPI path 92→93, contract bumped, dashboard schema regenerated (codegen-drift gate green), Spectral summary+description present.

All gates green: Test, Integration (macos), Clippy (exact CI cmd), SonarCloud (duplication), CodeQL/Analyze, codegen-drift, Spectral. Merging on green.

PR #1879, opened 2026-08-01
[AAASM-5349] 🐛 (aa-cli): Refuse a governed launch when no effective policy resolves

Chisanan232:
## Description

Implements **AAASM-5349** (owner Decision 4, Option A): an empty effective policy means **unconfigured** and a governed launch **fails closed**.

## The defect was worse than the ticket said

`load_policy()` returned a hard-coded empty rule set. I wrote that up as "enforces no policy rules." It is worse than that.

`aa-devtool-claude-code/src/settings.rs:51-58` derives the mode from the rules:

```rust
let permission_mode = if has_require_approval { "plan" }
                      else if has_deny        { "default" }
                      else                    { "acceptEdits" };
```

An empty policy has neither, so `aasm run` shipped **`permissionMode: "acceptEdits"`** to Claude Code — which auto-accepts file edits, where the tool's own default prompts. **A user running `aasm run` got *less* protection than running `claude` directly, while the receipt reported a successful governed launch.**

That is not "governance absent." It is governance actively loosening the tool's built-in posture and claiming credit for it.

## The four states

New `aa-cli/src/commands/run_policy.rs`:

```rust
enum PolicyResolution {
    Enforced { source, document },
    Permissive { source, document },
    Unconfigured(Unconfigured),   // NoSource{searched} | EmptyDocument{source}
    LoadFailed { source, detail },
}
```

Neither a bare rule list nor `Option<PolicyDocument>` can carry this: each variant obliges a different response, and the two refusing ones need different text. `Unconfigured` sub-splits because an operator who never configured a policy needs sending to the concept, while one whose file is empty needs sending to their file. `LoadFailed` stays top-level deliberately — **a corrupted policy presenting as "never configured" is exactly what an attacker corrupting a policy file would want.**

The adapter still receives `aa_core::PolicyDocument`; the trait signature is untouched, so blast radius stays inside `aa-cli`.

## The explicit allow-all artifact

No new config surface — an ordinary policy YAML whose tool ruleset is the wildcard:

```yaml
apiVersion: agent-assembly/v1
kind: Policy
metadata: {name: allow-all}
spec: {tools: {"*": {allow: true}}}
```

Selected by `--policy FILE` → `$AA_POLICY` → `~/.aasm/policy.yaml` → `~/.aasm/policies/` → `/etc/aasm/policy.yaml` → `/etc/aasm/policies/` — **the exact order `aasm gateway start` already uses**, so the two commands cannot disagree about where a host's policy lives.

Deliberately **not** a `permissive: true` switch: a switch is easy to flip without reading what it turns off. Writing the wildcard out is the smallest unambiguously deliberate act, and it validates through the same `PolicyValidator` as any other policy. Permissive also requires the *conjunction* — a wildcard-allow beside a network allowlist classifies as `enforced`, never `permissive`.

## The refusal an operator sees

```
policy=unconfigured — no policy artifact found; a governed launch is refused
error: refusing to launch ungoverned: no effective policy is configured, so this
session would run under no rules at all. An absent policy is not permission.

Searched, in order: --policy /does/not/exist.yaml

Supply a policy, then re-run:
  aasm run --policy <FILE> <tool>
  AA_POLICY=<FILE> aasm run <tool>
  install one at ~/.aasm/policy.yaml (or /etc/aasm/policy.yaml)
Check it first with: aasm policy validate <FILE>

If you intend this agent to be unrestricted, say so explicitly with an allow-all artifact:
    apiVersion: agent-assembly/v1
    kind: Policy
    ...
```

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature (four-state resolution)

## Breaking Changes

- [x] Yes — **`aasm run` now refuses when no policy resolves.** Previously it launched with an empty policy and a loosened `permissionMode`. Nothing that was genuinely governed stops working; what stops is a launch that reported governance it did not have.

## Related Issues

- Related Jira ticket: AAASM-5349

## Testing

**16 mutations, all killed.** Highlights:

| Mutation | Verbatim failure |
|---|---|
| **M13 — regress to the original defect** | kills all 6 fail-closed tests: `a launch with no effective policy must refuse: 0` · `the wildcard rule must reach the adapter, not an empty list` |
| **M4 — collapse four states into ok/not_ok** | `the four states must be four distinct values in AA_POLICY_STATE, not collapsed into ok/not-ok` |
| **M12b — receipt flattens enforced/permissive** | `states enforced and enforced render an identical policy receipt, so an operator cannot tell them apart` |
| M2 — LoadFailed→Unconfigured | `a policy that cannot be parsed is a broken artifact, not an absent one` |
| M17 — over-rejection | `an explicit allow-all policy must launch` |

**Worth reading M4 closely.** It killed the env-var test but **not** the receipt test, because the receipt's detail line still differed. That was not accepted as sufficient — M12 and M12b were added to prove the receipt test load-bearing in both directions (state named; states pairwise distinct). A four-variant enum that every consumer collapses into "ok / not ok" satisfies the type system and not the contract, and that was the specific trap here.

Gates: `cargo fmt --all` clean; `cargo clippy -p aa-cli --all-targets --all-features -- -D warnings` clean; `aa-cli` **951/951**.

## Checklist

- [ ] All CI checks passing

## Contract point 6 — partially met, reported not faked

Audit and receipt distinguish all four states, verified live: stderr banner, dry-run `--- policy ---` receipt, and `AA_POLICY_STATE` / `AA_POLICY_SOURCE` in the child environment.

**`aasm status` and `aasm integrations verify` are not covered.** `status` has no per-session effective-policy field at all today — it reports gateway-side aggregates — and adding one needs `status/{models,fetch,render}.rs` plus a gateway-side source of truth, all outside this lane. `verify` is about integration-receipt drift, an unrelated axis. Stopped and reported rather than reaching into another lane's files. **That residue wants its own ticket.**

## Sequencing

The `--no-proxy` lane (**AAASM-5350**) changes the same protection-reporting surface and lands behind this. Contact points were deliberately minimised: `build_child_env`'s signature is unchanged — the policy env is applied by `resolution.annotate_env(&mut child_env)` after it returns — so the overlap is one added `RunArgs` field and one line at each of two call sites.

## Found, not fixed

- **Directory cascade is refused, not merged.** `~/.aasm/policies/` is searched so `run` and `gateway start` agree on locations, but a directory source resolves to `load_failed` explaining that `run` renders one effective document. Fail-closed and honest; an operator using the cascade must pass `--policy FILE`.
- **Only the `tools:` dimension reaches managed settings.** `network`/`budget`/`capabilities` are validated and gate the permissive classification, but a dev-tool adapter has nowhere to put them — the gateway and proxy enforce those. Pre-existing.
- **Branch name carries `AAASM-5344`**, a placeholder key I supplied before the ticket existed. All four commit trailers say `AAASM-5349`; the remote branch is named correctly. Not rewritten, because that would need a force-push.
- **Unrelated environmental failure:** `aa-gateway … apply_good_succeeds_and_is_idempotent_on_postgres` fails with `initdb: … pg_wal: No space left on device`. The Docker VM is out of disk. This diff touches zero `aa-gateway` files.

### DIFF
--- a/aa-cli/src/commands/mod.rs
+++ b/aa-cli/src/commands/mod.rs
@@ -30,6 +30,7 @@ pub mod policy;
 pub mod proxy;
 // strip-for-publish:begin devtool
 pub mod run;
+pub mod run_policy;
 pub mod run_registration;
 // strip-for-publish:end devtool
 pub mod sandbox;
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -399,7 +399,9 @@ fn dry_run_handle(args: &RunArgs) -> RegistrationHandle {
 /// inferred later from what the tool was or was not stopped from doing.
 fn resolve_policy(args: &RunArgs) -> run_policy::PolicyResolution {
     let resolution = run_policy::resolve(args.policy.as_deref());
-    eprintln!("policy={} — {}", resolution.state_token(), resolution.summary());
+    // `summary()` already opens with the state token, so the `policy=<token>`
+    // prefix an audit scraper matches on falls out of it without repeating it.
+    eprintln!("policy={}", resolution.summary());
     resolution
 }
 
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -1583,6 +1583,88 @@ mod tests {
         );
     }
 
+    /// The four effective-policy states must survive as far as the artefact an
+    /// operator actually reads.
+    ///
+    /// This is the assertion that a four-variant enum on its own does not
+    /// satisfy: a consumer that renders every state the same way — or that
+    /// collapses them into "a policy loaded" / "it did not" — passes the type
+    /// checker and fails the contract. So the check is on the rendered receipt,
+    /// and it is pairwise: `enforced` and `permissive` both launch, and are the
+    /// pair most likely to be flattened into each other.
+    #[test]
+    fn the_dry_run_receipt_distinguishes_all_four_policy_states() {
+        let handle = stub_handle(None);
+        let cmd = std::process::Command::new("mock-tool");
+        let env = HashMap::new();
+
+        let states = [
+            run_policy::PolicyResolution::Enforced {
+                source: PathBuf::from("/p.yaml"),
+                document: PolicyDocument {
+                    version: 1,
+                    name: "p".into(),
+                    rules: vec![aa_core::PolicyRule {
+                        action_pattern: "bash".into(),
+                        decision: aa_core::PolicyDecision::Deny,
+                    }],
+                    enforcement_mode: aa_core::EnforcementMode::default(),
+                },
+            },
+            run_policy::PolicyResolution::Permissive {
+                source: PathBuf::from("/p.yaml"),
+                document: PolicyDocument {
+                    version: 1,
+                    name: "p".into(),
+                    rules: vec![aa_core::PolicyRule {
+                        action_pattern: "*".into(),
+                        decision: aa_core::PolicyDecision::Allow,
+                    }],
+                    enforcement_mode: aa_core::EnforcementMode::default(),
+                },
+            },
+            run_policy::PolicyResolution::Unconfigured(run_policy::Unconfigured::NoSource { searched: vec![] }),
+            run_policy::PolicyResolution::LoadFailed {
+                source: PathBuf::from("/p.yaml"),
+                detail: "bad".into(),
+            },
+        ];
+
+        let sections: Vec<String> = states
+            .iter()
+            .map(|state| {
+                let output = format_dry_run_output(&handle, state, "{}", &cmd, &env);
+                let start = output
+                    .find("--- policy ---")
+                    .expect("receipt must carry a policy section");
+                let rest = &output[start..];
+                let end = rest.find("--- managed settings ---").unwrap_or(rest.len());
+                rest[..end].to_string()
+            })
+            .collect();
+
+        for (state, section) in states.iter().zip(&sections) {
+            assert!(
+                section.contains(state.state_token()),
+                "the receipt must name the state it is reporting; {} is missing from:\n{section}",
+                state.state_token()
+            );
+        }
+
+        for (i, a) in sections.iter().enumerate() {
+            for (j, b) in sections.iter().enumerate().skip(i + 1) {
+                assert_ne!(
+                    a,
+                    b,
+                    "states {} and {} render an identical policy receipt, so an operator cannot \
+                     tell them apart",
+                    states[i].state_token(),
+                    states[j].state_token()
+                );
+            }
+        }
+    }
+
     #[test]
     fn dry_run_masks_secret_and_connection_url_env_vars() {
         let handle = RegistrationHandle {
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -10,8 +10,9 @@ use uuid::Uuid;
 #[cfg(unix)]
 use tokio::signal::unix::SignalKind;
 
-use aa_core::{DevToolAdapter, DevToolInfo, DevToolKind, GovernanceLevel, PolicyDocument, PolicyRule};
+use aa_core::{DevToolAdapter, DevToolInfo, DevToolKind, GovernanceLevel};
 
+use crate::commands::run_policy;
 use crate::commands::run_registration::{self, GovernedRegistration};
 use crate::commands::status::models::redact_database_url;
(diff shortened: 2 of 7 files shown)


Chisanan232:

Risk class: **changes what "governed" means, and breaks a workflow that previously succeeded.** Reviewed as a contract change, not a bug fix.

### The defect was worse than the ticket I wrote

I filed this as "enforces no policy rules." `aa-devtool-claude-code/src/settings.rs:51-58`:

```rust
let permission_mode = if has_require_approval { "plan" }
                      else if has_deny        { "default" }
                      else                    { "acceptEdits" };
```

An empty policy has neither, so `aasm run` shipped **`permissionMode: "acceptEdits"`** — which auto-accepts file edits, where Claude Code's own default prompts. **A user running `aasm run` got less protection than running `claude` directly, while the receipt reported a successful governed launch.** Not "governance absent": governance actively loosening the tool's built-in posture and taking credit.

### Four states, and the reason a bare list could not carry them

`LoadFailed` is deliberately top-level rather than folded into `Unconfigured`: **a corrupted policy presenting as "never configured" is exactly what an attacker corrupting a policy file would want.** `Unconfigured` sub-splits (`NoSource` / `EmptyDocument`) because one operator needs sending to the concept and the other to their file.

The adapter still receives `aa_core::PolicyDocument` — trait signature untouched, blast radius contained to `aa-cli`.

### The allow-all artifact is a document, not a flag

`spec: {tools: {"*": {allow: true}}}` in a full policy envelope, resolved through **the same search order `aasm gateway start` already uses**, so the two commands cannot disagree about where a host's policy lives. No `permissive: true` switch — a switch is easy to flip without reading what it turns off. Permissive also requires the conjunction: a wildcard-allow beside a network allowlist classifies as `enforced`.

### Mutation evidence, and one judgement I want to highlight

16 mutations, all killed. **M4 — collapsing the four states into `ok`/`not_ok`, the exact trap in this design — killed the env-var test but *not* the receipt test**, because the receipt's detail line still differed. That was not accepted as sufficient: M12 and M12b were added to prove the receipt test load-bearing in both directions (state named; states pairwise distinct). A four-variant enum every consumer collapses satisfies the type system and not the contract.

M13 — regressing to the original defect — kills all six fail-closed tests.

### The CI failures were this change working, and fixing them found something better

Four integration tests failed with `aasm run claude should exit 0`. They had never supplied a policy — **the tests that verify governed launches were themselves running ungoverned**, under the `acceptEdits` default.

The fix declined allow-all for them: each got a narrow enforced policy (`read_file: allow`, `shell: deny`), because "these files claim to exercise a *governed* launch; a session under a policy that restricts nothing would make that claim untrue."

**`a_launch_that_cannot_register_never_starts_the_tool` needed more than a policy.** `aasm run` refuses on three ordered grounds — proxy trust → policy → registration — and a run stopped by any earlier gate still exits non-zero and launches nothing, satisfying every assertion the test had while establishing nothing about the gate it names. It now pins `policy=enforced` as a precondition and asserts the refusal is *not* `refusing to launch ungoverned`, making the failure **attributable** to registration.

A fifth failure CI never reported — the AC4 real-tool scenario, gated on macOS plus a real `claude` — had the same cause and is fixed. It fails only where Claude Code is actually installed.

Final: **30/30**, including the registration test at 1076s. Two flaky (`aa-proxy` missed a 5s bind deadline at load average 46, caused by my own parallel lanes) passed on retry, in a file this PR does not touch — correctly not claimed as green until measured cleanly.

### Merge-order constraint, enforced

The **AAASM-5365 docs PR must merge after this**. It documents `run_policy.rs`, which does not exist on `main` until this lands. Shipping those docs first would recreate AAASM-5310 — documentation contradicting the binary — which this programme opened by fixing.

### Carried forward

- **Contract point 6 is partially met and reported, not faked.** Audit and receipt distinguish all four states (stderr banner, dry-run receipt, `AA_POLICY_STATE`/`AA_POLICY_SOURCE`). `aasm status` has **no per-session policy field at all**, and adding one needs `status/{models,fetch,render}.rs` plus a gateway-side source of truth — outside this lane. Wants its own ticket.
- **Directory cascade diverges:** `gateway start` merges `~/.aasm/policies/`; `run` refuses it as `load_failed` because it renders one effective document. Fail-closed and honest, but the two commands now differ on the same input.
- Branch name says `AAASM-5344`, a placeholder I supplied before the ticket existed; all commit trailers and the remote branch say `AAASM-5349`. Not rewritten — that needs a force-push.

Gates: not a draft; no conflicts; **26 pass / 18 skipping / 0 failing**; 0 unresolved threads; no requested-changes review; head `c2c12e53` current; zero files outside `aa-cli/` and `aa-integration-tests/`.

PR #1881, opened 2026-08-01
[AAASM-5333] 🐛 (aa-cli): Pair the Linux start tick with the pidfs inode

Chisanan232:
## Description

**Fixes red `main`.** `commands::proxy::identity::tests::start_token_differs_between_two_processes` has been failing on `ubuntu-latest` since `be192256` (run `30689213155`, job `91340914613`), failing all three nextest retries.

## It was never a flake

I filed AAASM-5333 as "a test flake *and* a narrow gap in the incarnation guarantee." That was wrong.

`/proc/<pid>/stat` field 22 is in `USER_HZ` ticks — measured `USER_HZ=100`, a 10 ms grid. Under `cargo nextest` every test runs in a **freshly spawned process**, so the test binary is itself only milliseconds old when it forks. Reproduced on Linux: **196 collisions in 200 runs.**

A 98% collision rate means the start-token half of the PID-reuse guard was **effectively absent on Linux**. It failed all retries because it was deterministic, not intermittent.

## The discriminator was chosen by measurement

Token becomes `linux-pidfs:<ticks>.<inode>` — the start tick paired with the **pidfs inode** from `pidfd_open(2)` (Linux ≥ 6.9).

The obvious candidate was tried and **rejected empirically**: the `/proc/<pid>` directory inode changed under dcache pressure (`22199307 → 22199310`), which would turn a running proxy into a refused one.

A self-generated nonce was rejected because it moves the fact from the kernel to the process being identified — a process can lie about a value it generates; it cannot lie about its own `/proc` start time.

The pidfs inode was verified on four properties: unique per process, **stable across reads**, `ESRCH` on a reaped PID (so `identity_of_a_dead_pid_is_unavailable` still holds), and — decisively — **readable for a non-dumpable process**, which `/proc/<pid>/exe` is not. That is the shape every real `aa-proxy` has, because of `prctl(PR_SET_DUMPABLE, 0)` (AAASM-3584) — the same constraint that broke Linux in #1861. It is the one discriminator that survives the problem shaping this module.

**To forge a token an attacker must now land on the recycled PID *and* match the victim's pidfs inode** — a counter allocated at process creation that owes nothing to the clock. Colliding on the tick no longer suffices.

**Honest limit:** the state file is `0600` and the trust boundary is "this user", so a same-uid attacker can rewrite the record outright. This guard is against *coincidence* — a second `aa-proxy` on a recycled number — and that hole is what closes here.

**Pre-6.9 kernels** (Ubuntu 22.04, Debian 12, RHEL 9) return one shared anonymous inode. Detected from the **filesystem magic** (`PID_FS_MAGIC`, verified unprivileged) rather than guessed from a version number, and such a token carries the distinct tag `linux-ticks:` — a coarse token never wears the fine scheme's badge. A test pins that branch, so the weakness cannot be silent.

## Old records are refused, not migrated

An old `linux-starttime:` record now yields `IdentityEvidenceUnrecognised`, telling the operator to re-run `aasm proxy start`. Migration was declined deliberately: the old evidence is a strict subset, so honouring it re-opens the hole.

**The check ordering is load-bearing.** The scheme check runs *before* the equality comparison — otherwise an old record reads as a start-time mismatch, i.e. *"your proxy was replaced by an impostor"*: confident, specific, and wrong.

## Type of Change

- [x] 🐛 Bug fix

## Breaking Changes

- [x] Yes — a proxy state record written by an older `aasm proxy start` is refused with an instruction to restart the proxy. Deliberate; the alternative re-opens the defect.

## Related Issues

- Related Jira ticket: AAASM-5333
- Fixes red `main` at `be192256`; regression introduced by #1861

## Testing

| Run | Result |
|---|---|
| **Old scheme**, 200 fresh-process runs | **196 / 200 collided** |
| **Fixed**, 200 fresh-process runs (Linux) | **0 failures / 200** |
| 30 × `cargo nextest run -p aa-cli commands::proxy::` | 0 failures / 30 |
| `cargo nextest run -p aa-cli` | **934 / 934** |
| `cli_run_trusted_proxy` 10-case matrix | **10 / 10**, Linux *and* macOS |

The test now **constructs** the collision — it retries spawns until the kernel reports equal ticks, and panics if it cannot — so all 200 runs provably exercised the collision rather than dodging it.

**Mutation A** (drop the inode, keep the tag):
```
assertion `left != right` failed: two processes are two incarnations even when the
kernel clock cannot separate them
  left: "linux-pidfs:112291171"
 right: "linux-pidfs:112291171"
```

**Mutation B** (`if false && token != state.start_token`) — read how it fails:
```
panicked at cli_run_trusted_proxy.rs:128: the operator must be told the launch was refused
error: refusing to launch unregistered: the governance gateway ... is unreachable
```
The tampered record **sails through the trust gate** and only trips a later, unrelated gateway check. That is the silent-bypass signature: the guard fails and the error points somewhere else entirely.

Both reverted; 75/75 green after. `cargo fmt` / `clippy -D warnings` clean on **both** macOS and Linux (the new code is Linux-only, so clippy was run inside a container too); `cargo doc` clean; lefthook on every commit, never bypassed.

Each of the four commits is bisectable — the 10-case matrix was verified passing at the intermediate commit, before the scheme check lands.

## Checklist

- [ ] All CI checks passing

## Found, not fixed

- **Pre-existing latent flake:** `commands::run::tests::no_gateway_credential_reaches_the_launched_tool` fails under plain `cargo test -p aa-cli --lib` on Linux but passes in isolation, on macOS, and under nextest (934/934). A parallel env-var race that nextest's process-per-test isolation hides. Anyone running `cargo test` rather than `nextest` will hit it.
- **Pre-6.9 kernels** cannot separate two same-tick processes. Unclosable — Linux publishes nothing finer — so the token declares it via the tag and a test pins that branch. Worth a ticket if those kernels are supported targets.
- **Integration-harness flake:** `TrustedProxy::start()` gives `aa-proxy` a 5 s bind timeout; on a loaded machine with cold binaries it times out and every test in the file fails with a misleading message. Hit once, passed on retry.
- **The Docker VM disk hit 100% during this work** (17 GB of reclaimable build cache remains). It is already causing unrelated failures elsewhere — `initdb: … pg_wal: No space left on device` in `aa-gateway` migration tests.

### DIFF
--- a/aa-cli/src/commands/proxy/identity.rs
+++ b/aa-cli/src/commands/proxy/identity.rs
@@ -22,6 +22,41 @@
 //!   process-identity pair for exactly this reason (it is what `pidfd`, systemd
 //!   and every correct pidfile implementation reduce to).
 //!
+//! # Why the start time alone is not enough on Linux (AAASM-5333)
+//!
+//! `/proc/<pid>/stat` publishes the start time in **clock ticks** — `USER_HZ`,
+//! which is 100 on every mainstream configuration, so a 10 ms grid. Two
+//! processes that started inside the same 10 ms carry byte-identical values, and
+//! a token built from that field alone cannot tell them apart. The hole is
+//! narrow but real: the executable check already rejects a recycled PID running
+//! *something else*, so the start time is the only thing left to reject a
+//! *second `aa-proxy`* that took the same number, and two incarnations landing
+//! in one tick would pass both checks.
+//!
+//! Linux exposes no finer start time, so the token pairs the tick with an
+//! independent kernel fact rather than a more precise clock: the **pidfs inode**
+//! behind `pidfd_open(2)`. From Linux 6.9 each process owns an inode on the
+//! `pidfs` filesystem, allocated at process creation from a counter that owes
+//! nothing to the clock — so a pair that collides on the tick still differs
+//! here, and forging a token now requires forging both. It survives the
+//! constraint that shapes the rest of this module, too: `pidfd_open` is not
+//! behind the ptrace gate, so it answers for the non-dumpable process every real
+//! `aa-proxy` is.
+//!
+//! Two candidates were measured and rejected. The `/proc/<pid>` directory's own
+//! inode is *not* stable — it is allocated at dentry lookup and observably
+//! changes when the dentry is evicted, which would turn a running proxy into a
+//! refused one. A nonce the proxy generates for itself would move the fact from
+//! the kernel to the process: a process can lie about a value it invents, where
+//! it cannot lie about its own creation, and identity evidence should not be
+//! sourced from the thing being identified.
+//!
+//! Kernels before 6.9 answer `pidfd_open` with one anonymous inode shared by
+//! every pidfd, which discriminates nothing. That case is detected by the
+//! filesystem magic rather than guessed at from a version number, and the token
+//! then carries a different scheme tag — the resolution it actually has is
+//! stated, never assumed.
+//!
 //! # Why per-platform, and why absent evidence is fatal rather than skippable
 //!
 //! Neither fact is portable. Each platform reads it natively; a platform with no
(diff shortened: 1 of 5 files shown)


Chisanan232:

Risk class: **repairs a red trunk, and changes a security guard's evidence.** Reviewed as both.

### My ticket was wrong twice, and this corrects it

I filed AAASM-5333 as "a test flake *and* a narrow gap in the incarnation guarantee", then routed it to a follow-up. Measured here: **196 collisions in 200 fresh-process runs.** Under nextest every test runs in a newly spawned process, so the binary is milliseconds old when it forks — the collision is the common case, not the corner. A 98% rate means the start-token half of the PID-reuse guard was **effectively absent on Linux**.

I had the evidence when I filed it — the #1861 implementer reported 3/3 retry failures in a container — and I inferred "flake" from the shape rather than measuring the rate. That is the second time this session I described a symptom accurately and guessed its severity.

### The discriminator was chosen by measurement, which is the part I care about

- `/proc/<pid>` **directory inode: tried and rejected** — observed changing under dcache pressure (`22199307 → 22199310`). Using it would turn a running proxy into a refused one. That is a failure mode you only find by trying it.
- **Self-generated nonce: rejected on principle** — it moves the fact from the kernel to the process being identified. A process can lie about a value it generates; it cannot lie about its own `/proc` start time.
- **pidfs inode: verified on four properties**, the decisive one being that it is **readable for a non-dumpable process** where `/proc/<pid>/exe` is not. Every real `aa-proxy` is non-dumpable (`prctl(PR_SET_DUMPABLE, 0)`, AAASM-3584) — the same constraint that broke Linux in #1861. So the chosen discriminator is the one that survives the problem shaping this module, rather than the first one that looked unique.

Pre-6.9 kernels are handled honestly: detected from **filesystem magic**, not guessed from a version number, and tagged `linux-ticks:` so a coarse token never wears the fine scheme's badge. A test pins that branch, so the weakness cannot become silent.

### Rejecting old records is right, and the ordering is load-bearing

Migration was declined because the old evidence is a strict subset — honouring it re-opens the hole. The scheme check runs **before** the equality comparison, because otherwise an old record reads as a start-time mismatch: *"your proxy was replaced by an impostor"* — confident, specific, and wrong. Getting that order backwards would have produced a security-shaped lie on every upgrade.

### Mutation B is the one to read

With the incarnation check disabled, the tampered record **passes the trust gate** and only trips a later, unrelated gateway check:

```
error: refusing to launch unregistered: the governance gateway ... is unreachable
```

The guard fails and the error points somewhere else entirely. That is the silent-bypass signature this programme has been chasing all session, reproduced deliberately.

### Honest limits, stated in the PR rather than found later

The state file is `0600` and the trust boundary is "this user", so a same-uid attacker can rewrite the record outright — this guard is against **coincidence**, not a local adversary, and that is what closes. Pre-6.9 kernels remain unseparable because Linux publishes nothing finer.

### Evidence

200/200 clean after the fix, with a test that **constructs** the collision — retrying spawns until the kernel reports equal ticks, panicking if it cannot — so every run provably exercised it rather than dodging it. `aa-cli` 934/934. The 10-case refusal matrix 10/10 on **both** platforms. Clippy clean on macOS *and* inside a Linux container, since the new code is Linux-only. Four commits, each bisectable — the matrix was verified passing at the intermediate commit.

Gates: not a draft; no conflicts; **23 pass / 18 skipping / 0 failing**; 0 unresolved threads; no requested-changes review; head `d38eac2a` current. Zero files outside `aa-cli/src/commands/proxy/` and its tests — `Cargo.lock`, `deny.toml`, `ci.yml`, `docs/` untouched.

PR #1883, opened 2026-08-01
[AAASM-5332] ✨ (aa-sdk-client): Replace the derived registration key with a durable CSPRNG identity

Chisanan232:
## Description

Implements **AAASM-5332** (owner Decision 1, Option B): the agent registration identity becomes a **real Ed25519 keypair from a CSPRNG**, durable until explicit rotation or revocation — replacing a private key that was a pure function of a public value.

## The defect

`aa-sdk-client/src/keypair.rs:43-51` seeded the signing key with `SHA-256(identifier)`, where the identifier is the **public** agent id. The possession proof therefore established that the caller could compute SHA-256, not that it held a secret.

Every surrounding control was correctly implemented and none helped:

- `enforce_did_key_binding` (`lifecycle_service.rs:232-239`, called at `:384` and `:437`) binds DID↔public key — and a derived pair is self-consistent.
- `verify_possession_proof` (`:241-259`) is a real `verify_strict` — over a key the attacker can also derive.
- The nonce is genuinely single-use (`:453-455`) — correct, and orthogonal.

And registration is reachable **unauthenticated by design**: `AgentLifecycleServiceServer` is mounted with `enrich`, not `auth` (`server.rs:728`, `:873`), and `enrich_interceptor` returns `Ok(req)` **unconditionally** (`grpc_auth.rs:150-161`). Right for a bootstrap endpoint — and it made the possession proof the only thing deciding who may register as a given agent.

## The proof this is fixed

**M1 — revert to the derived key — kills 8 assertions across 3 crates**, including an end-to-end forgery against a real gateway:

```
a proof forged from the agent id verified against the victim's key — knowing public
data is still sufficient to register as this agent

if one identifier still determines one DID, the private key is still a function of
public data and the possession proof still proves nothing

a second installation reached the first one's identity from the identifier alone
```

Eleven further mutations, each verbatim-confirmed: `create_new`→`create` (`a second enrolment must be refused, but a usable key was returned`), permission-bit check, uid check, file mode, revocation, identity mismatch, rotation ×2, CSPRNG→constant (`two generated identity keys must differ`), token-in-keyfile, empty `AASM_STATE_DIR`.

**M5 caught a bug in the test itself.** Deleting the symlink guard left it green — the refusal message embeds the path, and the temp dir was named `symlink-store`, so the assertion was matching its own fixture name. Fixed in `05733597`; it now asserts `"it is a symlink"` and dies with `it is not a regular file` under mutation.

## Where the key lives

`${AASM_STATE_DIR:-~/.aasm}/identity/<sha256(agent_id)>.key` — mode `0600` in a `0700` directory, created with `O_EXCL` and the mode in a single `open(2)`. Refuses a symlink, a non-regular file, a foreign uid, or **any** group-or-other access (`0o077` — stricter than the proxy state file's `0o022`).

The filename is hashed only for filesystem safety; the agent id is stored **inside** and checked on read, so a hash collision is refused rather than silently sharing a key.

**An attacker previously needed only to read an agent id.** Now they need read access to that file as its owning uid — and if they obtain it by loosening permissions or symlinking, the agent refuses to start rather than registering insecurely.

## One derived key was kept, deliberately

`aa-runtime/src/ipc/handshake.rs:54-57` also recomputes `SHA-256(agent_id)`. That key is documented at both ends as **non-secret** (AAASM-3922); its real boundary is socket `0600` plus peercred. Renamed to `derive_transport_key` and left in place, so `aa-runtime` needed no change. Registration no longer touches it.

Not every derived key was the bug — only the one claiming to prove possession.

## Migration — honest about what cannot be saved

**Every pre-existing `did:key` is compromised retroactively**: anyone who recorded an identifier already holds the key. No key material needs migrating, because the gateway never stored a private key.

Upgrading enrols a fresh random key and registers under a **new** DID. Old records are **not** cleaned up automatically and must be deregistered deliberately, using `legacy_derived_did(agent_id)` to name them. **Gateway audit entries will not join across the migration** — they key on `SHA-256(did)[..16]`. Runtime-written entries hash the plaintext agent id and are unaffected.

Documented at `docs/src/migration/agent-identity-keys.md`.

## Type of Change

- [x] ✨ New feature
- [x] 🐛 Bug fix (security)

## Breaking Changes

- [x] Yes — agents re-enrol under a new DID; audit history does not join across the migration. `AssemblyConfig` gains a public `identity_dir` field, so the three external SDK shims that build it with a struct literal will need it at their next git-SHA bump.

## Related Issues

- Related Jira ticket: AAASM-5332 (owner decision recorded in comment 19980)

## Testing

`aa-sdk-client` 99 pass; `aa-sdk-client` + `aa-cli` 1030 pass; workspace `clippy --all-targets --all-features -D warnings` clean; `cargo doc` adds no warnings. Tree clean, 22 files, **zero** forbidden files — `run.rs`, `proxy/**`, `Cargo.lock`, `deny.toml`, workflows and every `Cargo.toml` untouched.

## Checklist

- [ ] All CI checks passing

## Reviewer notes — two constraint-driven choices

1. **`/dev/urandom` is read directly rather than using `getrandom` or `rand`**, because `Cargo.lock` was off-limits to this lane and adding a dependency would have touched another session's file. Same kernel source; swapping in the crate later is mechanical. Flagging it as a constraint artefact rather than the idiomatic choice.
2. **The process uid is learned by chmod-ing our own key directory** rather than `libc::geteuid`, for the same reason.

Neither is wrong; both are worth a second look if the dependency constraint is lifted.

## Deferred, named — not silently dropped

**Gateway-side revocation propagation.** `proto/agent.proto` has only `RequestChallenge` / `Register` / `Heartbeat` / `Deregister` / `ControlStream` — no revoke RPC and no revocation list. Adding one touches `proto/` and `aa-gateway/`, outside this lane. The strongest propagation available today is `Deregister` on the revoked DID. Local rotate and revoke operations **are** implemented.

## Found, not fixed

- **`run.rs` needs two follow-ups from its owner** (untouchable from this lane): its `--dry-run` path enrols a throwaway key when `--agent-id` is absent (`dry-run-<uuid>`); and its doc at `:162-163` claims the DID is what *all* audit records attribute to — true only for gateway-written entries. `registration_did(&str) -> String` was kept so `run.rs` compiles unchanged, returning `<no-durable-identity-key>` (deliberately not a DID) when no key can be established.
- **Pre-existing defect worth its own ticket, unrelated to this change:** `AgentViolationCounts` keys on `SHA-256(did)[..16]` but is looked up with the registry key `SHA-256("org/team/did")[..16]` (`agent_violations.rs:64` vs `agents.rs:204`), so `policy_violations_count` and `is_flagged` appear to be **0 for every agent whenever a team or org is set**. The same mismatch would hit `GET /agents/{id}/decisions`. Reported from the implementer's analysis; not independently re-verified here.
- 3 flaky `aa-gateway` hot-reload/watcher tests under heavy load (passed on retry). Nothing in `aa-gateway/src` was changed.

### DIFF
--- a/aa-api/tests/serve_local_grpc_registration.rs
+++ b/aa-api/tests/serve_local_grpc_registration.rs
@@ -28,6 +28,14 @@ fn sdk_config(agent_id: &str, endpoint: String) -> AssemblyConfig {
         team_id: None,
         parent_agent_id: None,
         sdk_version: None,
+        // A per-process temporary identity store, so the durable key this test
+        // enrols never lands in the developer's real `~/.aasm`.
+        identity_dir: Some(
+            std::env::temp_dir()
+                .join(format!("aa-api-grpc-reg-identity-{}", std::process::id()))
+                .to_string_lossy()
+                .into_owned(),
+        ),
     }
 }
 
@@ -64,7 +72,7 @@ async fn grpc_registered_agent_is_visible_via_rest() {
             .await
             .expect("connect to embedded gRPC AgentLifecycleService");
         let challenge = client
-            .request_challenge(build_challenge_request(&config))
+            .request_challenge(build_challenge_request(&config).expect("the agent must have a durable identity key"))
             .await
             .expect("request_challenge succeeds");
         let request = build_register_request(
@@ -72,7 +80,8 @@ async fn grpc_registered_agent_is_visible_via_rest() {
             "grpc-reg-test-agent".to_string(),
             "langgraph".to_string(),
             &challenge.nonce,
-        );
+        )
+        .expect("the agent must have a durable identity key");
         let response = client.register(request).await.expect("register succeeds");
         assert!(
             !response.credential_token.is_empty(),
--- a/aa-cli/src/commands/run_registration.rs
+++ b/aa-cli/src/commands/run_registration.rs
@@ -19,6 +19,15 @@
 //! * `verify_possession_proof` over a **server-issued, single-use, time-bound**
 //!   nonce obtained from `RequestChallenge` (AAASM-3591 / AAASM-3866).
 //!
+//! Those three checks only bite if the private key behind them is a secret.
+//! Until AAASM-5332 it was not — it was `SHA-256(agent_id)`, and the agent id is
+//! published in audit records and on the dashboard — so all three were
+//! satisfiable by anyone who had read one. The key is now generated randomly and
+//! stored owner-only by `aa-sdk-client`'s `identity_store`, and this module
+//! registers with that key: `register` refuses with
+//! [`RegistrationError::IdentityUnavailable`] rather than launching under an
+//! identity it cannot prove it holds.
+//!
 //! Serving the CLI over a second, weaker route would have meant one product with
 //! two registration contracts, the weaker of which is reachable by anything that
 //! can speak HTTP. So `aasm run` registers through the *same* gRPC handshake the
@@ -55,15 +64,25 @@ const LAUNCH_FRAMEWORK: &str = "aasm-run";
 /// A registration the gateway accepted, and the identity it was accepted under.
 #[derive(Clone)]
 pub struct GovernedRegistration {
-    /// The operator-facing identifier the keypair (and therefore the DID) is
-    /// derived from. This is the value handed to the launched tool as
+    /// The operator-facing identifier that *selects* the agent's durable
+    /// identity key. This is the value handed to the launched tool as
     /// `AA_AGENT_ID`, and it is what makes the correlation in
     /// [`registration_did`](Self::registration_did) reproducible: anything that
-    /// derives from this identifier — an SDK inside the launched tool, a later
-    /// `aasm run` for the same agent — arrives at the same DID.
+    /// resolves this identifier — an SDK inside the launched tool, a later
+    /// `aasm run` for the same agent — loads the same stored key and so arrives
+    /// at the same DID.
+    ///
+    /// Since AAASM-5332 it selects a key rather than *being* one. The identifier
+    /// is public (it appears in audit records and on the dashboard); the key it
+    /// names is random and readable only by this user.
     pub agent_id: String,
-    /// The `did:key` the gateway registered, and the identity every audit record
-    /// for this session is attributed to.
+    /// The `did:key` the gateway registered, and the identity the **gateway's**
+    /// audit records for this session are attributed to (they carry
+    /// `SHA-256(did)[..16]`).
+    ///
+    /// Not the whole attribution story: entries written by `aa-runtime` on the
+    /// SDK path hash the plaintext `AA_AGENT_ID` instead, so the two halves of
+    /// the trail are joined by this struct rather than by a shared identifier.
     pub registration_did: String,
     /// The gateway's own registry key for this agent, hex-encoded.
     ///
@@ -104,6 +123,9 @@ pub enum RegistrationError {
     ChallengeRefused { endpoint: String, reason: String },
     /// The gateway refused the registration itself.
     RegistrationRefused { endpoint: String, reason: String },
+    /// This machine has no usable durable identity key for the agent, so there
+    /// is nothing to register as (AAASM-5332).
+    IdentityUnavailable { reason: String },
 }
 
 impl std::error::Error for RegistrationError {}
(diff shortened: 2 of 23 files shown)


Chisanan232:

Risk class: **cryptographic identity.** The highest-consequence change in this programme; reviewed accordingly.

### The defect, and why every surrounding control failed to help

`keypair.rs:43-51` seeded the Ed25519 signing key with `SHA-256(identifier)` — the **public** agent id. The possession proof established that the caller could compute SHA-256, not that it held a secret.

Each surrounding control is correctly implemented and none helped, because all rest on that same non-secret: `enforce_did_key_binding` binds DID↔public key, and a derived pair is self-consistent; `verify_possession_proof` is a real `verify_strict`, over a key the attacker can also derive; the nonce is genuinely single-use, and orthogonal.

Registration is reachable **unauthenticated by design** — `AgentLifecycleServiceServer` mounts `enrich`, not `auth`, and `enrich_interceptor` returns `Ok(req)` unconditionally. Correct for a bootstrap endpoint, and it made the possession proof the only thing deciding who may register as a given agent.

### M1 is the review

Reverting to the derived key **kills 8 assertions across 3 crates**, including an end-to-end forgery against a real gateway:

```
a proof forged from the agent id verified against the victim's key — knowing public
data is still sufficient to register as this agent

a second installation reached the first one's identity from the identifier alone
```

Eleven further mutations each verbatim-confirmed — `create_new`→`create`, permission bits, uid check, file mode, revocation, identity mismatch, rotation ×2, CSPRNG→constant, token-in-keyfile, empty state dir.

**M5 caught a bug in the test itself.** Deleting the symlink guard left it green — the refusal message embeds the path and the temp dir was named `symlink-store`, so the assertion was matching its own fixture name. Fixed; it now dies with `it is not a regular file` under mutation. That is the fourth time this session a mutation has caught a test passing for the wrong reason.

### Judgement I agree with: one derived key was kept

`aa-runtime/src/ipc/handshake.rs:54-57` also recomputes `SHA-256(agent_id)`. That key is documented at both ends as **non-secret** (AAASM-3922); its boundary is socket `0600` plus peercred. Renamed to `derive_transport_key` and left alone.

Not every derived key was the bug — only the one claiming to prove possession. Changing it would have been scope creep dressed as thoroughness.

### On-disk protection

`${AASM_STATE_DIR}/identity/<sha256(agent_id)>.key`, mode `0600` in a `0700` dir, created `O_EXCL` with the mode in a single `open(2)`. Refuses symlinks, non-regular files, a foreign uid, and **any** group-or-other access (`0o077` — stricter than the proxy state file's `0o022`). The filename is hashed for filesystem safety only; the agent id is stored inside and checked on read, so a collision is refused rather than silently sharing a key.

An attacker previously needed **only to read an agent id**. Now they need read access to that file as its owning uid — and obtaining it by loosening permissions or symlinking makes the agent refuse to start rather than register insecurely.

### Migration is honest about what cannot be recovered

**Every pre-existing `did:key` is compromised retroactively.** Upgrading enrols a fresh key under a **new** DID; old records are not cleaned up automatically and must be deregistered deliberately via `legacy_derived_did()`. **Gateway audit entries will not join across the migration.** That is a real cost, documented at `docs/src/migration/agent-identity-keys.md` rather than glossed.

### Re-verified against the tree it merges into

The original green predated #1879 (fail-closed policy). A green result against a tree lacking that contract change is not a green result against this one — that is precisely what caught #1879 itself. Refreshed by merging `main` (not rebasing), re-run locally at **1055/1055**, and CI on the refreshed head `b70e6726` is **50 pass / 18 skipping / 0 failing**, including CodeQL reporting **"No new alerts in code changed by this pull request."**

### Reviewer-visible constraint artefacts, flagged not hidden

`/dev/urandom` is read directly rather than via `getrandom`, and the process uid is learned by chmod-ing our own key directory rather than `libc::geteuid` — both because `Cargo.lock` was owned by another lane and adding a dependency was not available. Same kernel source; swapping in the crate later is mechanical. Worth revisiting once that constraint lifts.

### Deferred and named, not silently dropped

**Gateway-side revocation propagation.** `proto/agent.proto` has no revoke RPC and no revocation list; adding one touches `proto/` and `aa-gateway/`. Local rotate and revoke **are** implemented; strongest propagation today is `Deregister`.

### Found, not fixed

- `run.rs` needs two follow-ups from its owner: the `--dry-run` path enrols a throwaway key when `--agent-id` is absent, and its doc at `:162-163` overstates what the DID attributes.
- **Pre-existing defect worth its own ticket:** `AgentViolationCounts` keys on `SHA-256(did)[..16]` but is looked up with `SHA-256("org/team/did")[..16]`, so `policy_violations_count` and `is_flagged` may be **0 for every agent whenever a team or org is set**. Reported from the implementer's analysis; not independently re-verified.
- `AssemblyConfig` gains a public `identity_dir` field — the three external SDK shims will need it at their next git-SHA bump.

Gates: not a draft; no conflicts; 50 pass / 18 skipping / 0 failing; 0 unresolved threads; no requested-changes review; head `b70e6726` current; zero forbidden files.

PR #1886, opened 2026-08-01
[AAASM-5348] 🐛 (aa-cli): Refuse a non-loopback proxy bind before anything starts

Chisanan232:
## Description

Implements **AAASM-5348** (owner Decision 3, Option A): `aasm proxy start` refuses a non-loopback bind, removing the split brain where the proxy started successfully on an address `aasm run` would never trust.

## The split brain, reproduced

```
$ aasm proxy start --listen 0.0.0.0:18991
Proxy started on http://0.0.0.0:18991
$ aasm proxy status
running (PID 19426, listening on 0.0.0.0:18991)

$ aasm run claude
error: refusing to launch ungoverned: the recorded proxy endpoint `0.0.0.0:18991`
is unusable: it is not a loopback address, so intercepted traffic would leave this machine
```

The refusing side was `proxy/trust.rs:359`; the permissive side was `proxy/start.rs::dispatch`, which had **no address check at all**.

## Both required protections are absent — so the opt-in always refuses

The contract requires a non-loopback bind to have TLS and client authentication. Neither exists:

- **Listener TLS: absent.** `aa-proxy/src/proxy/mod.rs:255` binds a bare `TcpListener` and speaks plain HTTP `CONNECT`. The three `rustls` configs there are not listener protection — `:1068` is the per-host MitM *server* cert presented **inside** an established tunnel (`.with_no_client_auth()`); `:385`/`:394` are the **upstream client** config.
- **Client authentication: absent.** No `Proxy-Authorization`, no `407`, no peer check anywhere in the crate.

So `--allow-remote-clients` is **implemented as a flag whose preconditions can never currently be met**, and it says so in the code, in `--help`, and in the refusal. It is driven by a `REMOTE_PROTECTIONS` table (`aa-proxy/src/config.rs:229`) with both marked `available: false` — flipping either constant relaxes the guard automatically when the protection lands.

A flag that always refuses is honest. A flag that appeared to gate on protections which do not exist would be **worse than the split brain it replaces**, because it would read as "remote clients are supported, with safeguards" when neither half is true.

## What an operator sees

```
### bare non-loopback
error: refusing to listen on 0.0.0.0:18992: it is not a loopback address, so the proxy
would accept connections from other hosts. aa-proxy reads intercepted traffic under a CA
this machine trusts and injects the operator's provider credentials into forwarded
requests, so anything that can reach the listener can do both. Listen on a loopback
address (for example 127.0.0.1:18992), or pass --allow-remote-clients to state the intent
explicitly.
exit=1 ; state files: 0

### with --allow-remote-clients
error: refusing to listen on 0.0.0.0:18992: --allow-remote-clients was given, but a proxy
reachable from other hosts also requires protection aa-proxy does not implement: TLS on
the proxy listener, client authentication and authorization. Being reachable is not being
trusted — without those, every host that can route to 0.0.0.0:18992 is an authorized
client of an interception endpoint that holds CA material and provider credentials.
exit=1 ; state files: 0

### loopback
Proxy started on http://127.0.0.1:18992 ; exit=0
```

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature (`--allow-remote-clients`)

## Breaking Changes

- [x] Yes — `aasm proxy start --listen <non-loopback>` now refuses. It previously succeeded and produced an endpoint `aasm run` would not trust, so no working configuration is lost.

## Related Issues

- Related Jira ticket: AAASM-5348 (owner decision recorded at creation)
- Surfaced **AAASM-5370** — see below

## Testing

**12 mutations, all killed.** The load-bearing pair is **M6/M7**: the same refusal moved to *after* spawn and *after* the port binds. They keep the exit code non-zero and change **only** ordering, isolating "refused before anything binds or writes state" from "refused at all":

```
a refused start must not leave a state file at /var/folders/.../data/proxy.pid
a refused start must not have bound port 54661
```

Others: M1 drops the opt-in gate; M2 makes the opt-in sufficient (`--allow-remote-clients must not by itself authorize an unprotected listener`); M3/M10 claim each protection is implemented (`both protections are absent; a change here must be matched by one in the listener`); M4 breaks loopback acceptance; M5 lets an unparseable `--listen` silently become the default; M8 removes the guard entirely, reproducing pre-fix behaviour; M11 stops naming the address; M12 defaults the opt-in on.

`aa-proxy` + `aa-cli` **1207/1207**; `cli_proxy_remote_bind_refusal` + `cli_run_trusted_proxy` + `cli_run` **22/22**; `cli_proxy` + `cli_run_claude_launch_env` **21/21**. `cargo doc --workspace --no-deps` adds no new warnings. The integration test uses `proxy_trust_support::{aasm_binary, aa_proxy_binary}` — the unconditional-rebuild path, so no stale-binary risk.

## Checklist

- [ ] All CI checks passing

## Found, not fixed — one is a live exposure

**The `aa-proxy` process itself is still exposable, and closing it needs files outside this ticket.** The guard lives in `aasm proxy start`; `ProxyConfig::from_env()` still honours `AA_PROXY_ADDR`, which is exactly how the container runs. Wiring `check_bind_addr` into `from_env()` was deliberately **not** done here because it would break three shipped things at once:

1. `aa-proxy/Dockerfile:41` — `ENV AA_PROXY_ADDR="0.0.0.0:8080"`; the published image would refuse to start.
2. `ci/integration/bypass-permissions/docker-compose.yml`, driven by a workflow outside this lane's ownership; CI would go red.
3. `examples/docker-compose/docker-compose.yml` and `docs/src/usage-guide/self-hosting.md:148` document the exposure as supported.

That third item is a live security issue on its own merits and is now **AAASM-5370**: the shipped self-host example sets `AA_PROXY_ADDR: "0.0.0.0:8899"` **and** publishes `"8899:8899"` to all host interfaces, with `AA_PROXY_MCP_FAIL_OPEN: "1"`. `check_bind_addr` is public and ready for that caller.

**Port 0 was the same split reached by a different property of the address, and is now also refused** (second commit). `trust.rs:365` rejects a port-0 endpoint, so a proxy started on one could never be routed at — but it was the worse of the two outcomes, not a smaller one. `aa-proxy` parses `:0` fine and binds an OS-assigned port; `wait_for_port("127.0.0.1:0")` then fails, `start.rs:254` removes the pid file, and **the child keeps running** — an interception process holding CA material and provider credentials that `aasm proxy stop` can no longer name or reach. The check runs *before* the loopback branch, because it disqualifies every address and `--allow-remote-clients` states an intent about reachability that says nothing about a recordable port. Two tests added; the ordering was mutation-checked by moving the condition below the loopback `return Ok(())`, which lets `127.0.0.1:0` through and fails the test. `aa-proxy` + `aa-cli` now **1209/1209**.

**Orphaned proxy on any failed start — pre-existing, not fixed here, ticket filed.** Refusing port 0 removes one way to reach it, but the underlying defect is general: *any* `wait_for_port` timeout (port already in use, missing CA dir, a crash during startup) takes the same path — the spawned `aa-proxy` is never killed and the pid file is removed, so the CLI loses its handle on a running interception process. Fixing it means killing the child on the failure path and is a change to the spawn lifecycle, outside this ticket's contract.

**Pre-existing, unrelated:** `cli_proxy proxy_logs_follow_streams_new_entries` fails deterministically under `cargo nextest` without `AASM_BIN_PATH` — its fixture shells out to `cargo run`, which blocks on the build lock nextest already holds. Passes with `AASM_BIN_PATH` set, which is the CI path.

### DIFF
--- a/aa-cli/src/commands/proxy/start.rs
+++ b/aa-cli/src/commands/proxy/start.rs
@@ -28,8 +28,15 @@ pub struct StartArgs {
     /// File to redirect proxy stdout/stderr to (background mode only).
     #[arg(long)]
     pub log_file: Option<PathBuf>,
-    /// State that a non-loopback `--listen` address is intended. Still refused
-    /// until the proxy can protect such a listener — see `checked_listen`.
+    // The doc comment is the `--help` text clap shows, so it is written for the
+    // operator reading it; the reasoning behind the refusal lives on
+    // `checked_listen`.
+    /// State that a non-loopback `--listen` address is intended
+    ///
+    /// Intent is not authorization. A proxy reachable from other hosts also
+    /// needs TLS on its listener and client authentication, neither of which
+    /// aa-proxy implements — so this option currently changes only which
+    /// refusal you get, and never permits the bind.
     #[arg(long)]
     pub allow_remote_clients: bool,
 }
--- a/aa-cli/src/commands/proxy/start.rs
+++ b/aa-cli/src/commands/proxy/start.rs
@@ -325,6 +325,72 @@ mod tests {
         assert_eq!(w.inner.listen, "0.0.0.0:9000");
     }
 
+    /// The default is the address `aasm run` will trust, so the shipped
+    /// behaviour must survive the guard untouched.
+    #[test]
+    fn the_default_listen_address_is_accepted() {
+        let default = Wrapper::parse_from(["test"]).inner.listen;
+        assert_eq!(checked_listen(&default, false), Ok(()), "default {default} must start");
+    }
+
+    #[test]
+    fn an_explicit_loopback_listen_address_is_accepted() {
+        assert_eq!(checked_listen("127.0.0.1:9000", false), Ok(()));
+        assert_eq!(checked_listen("[::1]:9000", false), Ok(()));
+    }
+
+    /// AAASM-5348: the address that used to start a proxy `aasm run` would
+    /// never trust. The diagnostic has to carry the reason, since the operator
+    /// asked for this address on purpose.
+    #[test]
+    fn a_bare_non_loopback_listen_address_is_refused() {
+        let reason = checked_listen("0.0.0.0:9000", false)
+            .expect_err("a proxy reachable from other hosts must not start by default");
+        assert!(reason.contains("0.0.0.0:9000"), "must name the address, got: {reason}");
+        assert!(
+            reason.contains("not a loopback address"),
+            "must say what disqualified it, got: {reason}"
+        );
+        assert!(
+            reason.contains("--allow-remote-clients"),
+            "must point at the option that states the intent, got: {reason}"
+        );
+    }
+
+    /// The opt-in records intent; it cannot supply protections the proxy does
+    /// not have. Naming them is the point — "refused" alone would read as a bug.
+    #[test]
+    fn the_opt_in_does_not_authorize_an_unprotected_remote_listener() {
+        let reason = checked_listen("0.0.0.0:9000", true)
+            .expect_err("--allow-remote-clients must not open an unauthenticated listener");
+        assert!(
+            reason.contains("TLS on the proxy listener"),
+            "must name the missing transport protection, got: {reason}"
+        );
+        assert!(
+            reason.contains("client authentication and authorization"),
+            "must name the missing client-identity protection, got: {reason}"
+        );
+    }
+
+    /// A `--listen` the proxy cannot parse is refused here rather than becoming
+    /// an opaque "did not bind within 5s" after a process has been spawned.
+    #[test]
+    fn an_unparseable_listen_address_is_refused() {
+        let reason = checked_listen("localhost:9000", false).expect_err("a hostname is not an ip:port literal");
+        assert!(reason.contains("localhost:9000"), "must quote the input, got: {reason}");
+    }
+
+    #[test]
+    fn start_args_allow_remote_clients_defaults_false() {
+        assert!(!Wrapper::parse_from(["test"]).inner.allow_remote_clients);
+        assert!(
+            Wrapper::parse_from(["test", "--allow-remote-clients"])
+                .inner
+                .allow_remote_clients
+        );
+    }
+
     #[test]
     fn start_args_gateway_is_optional() {
         let w = Wrapper::parse_from(["test"]);
--- a/aa-cli/src/commands/proxy/start.rs
+++ b/aa-cli/src/commands/proxy/start.rs
@@ -29,7 +29,7 @@ pub struct StartArgs {
     #[arg(long)]
     pub log_file: Option<PathBuf>,
     /// State that a non-loopback `--listen` address is intended. Still refused
-    /// until the proxy can protect such a listener — see [`checked_listen`].
+    /// until the proxy can protect such a listener — see `checked_listen`.
     #[arg(long)]
     pub allow_remote_clients: bool,
 }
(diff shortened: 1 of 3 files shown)


Chisanan232:

Implements owner Decision 3 (Option A). Reviewed adversarially, since this touches a security claim, a CLI contract, and a shipped default.

### The split-brain is genuinely closed, and the two halves now share one predicate

`check_bind_addr` (`aa-proxy/src/config.rs`) and `trust::verify_endpoint` (`aa-cli/src/commands/proxy/trust.rs:359`) both test `socket.ip().is_loopback()` — the *same* call, not two hand-rolled equivalents. I specifically checked `[::1]`, the obvious place for a residual divergence: accepted by both. There is no address one command trusts and the other refuses.

### `--allow-remote-clients` always refuses, and that is the correct shape

Both required protections are absent, verified in source rather than assumed:

- **Listener TLS** — `aa-proxy/src/proxy/mod.rs:255` binds a bare `TcpListener` and speaks plain `CONNECT`. The three `rustls` configs nearby are *not* listener protection: `:1068` is the per-host MitM leaf presented **inside** an established tunnel (`.with_no_client_auth()`), and `:385`/`:394` are the **upstream client** config.
- **Client authentication** — no `Proxy-Authorization`, no `407`, no peer check anywhere in the crate.

So the flag is implemented as one whose preconditions cannot currently be met, driven by a `REMOTE_PROTECTIONS` table with both entries `available: false`. A flag that always refuses is honest; a flag that appeared to gate on protections which do not exist would be worse than the split-brain it replaces. Flipping either constant relaxes the guard automatically when the protection lands.

### `localhost:9000` is now refused — checked this is not a regression

`aa-proxy`'s own `parse_bind_addr` uses `parse::<SocketAddr>()` with no hostname resolution, so a hostname never worked. Refusing it at the CLI turns a delayed opaque "did not bind within 5s" into an immediate diagnostic. No working configuration is lost.

### Second commit — port 0, which the original PR body had understated

I filed port 0 as "a smaller version of the same split". It is the **larger** one. `aa-proxy` parses `:0` fine and binds an OS-assigned port; `wait_for_port("127.0.0.1:0")` then probes port 0 and fails; `start.rs:254` removes the pid file — and **the child is never killed**. The result is a live interception process holding CA material and provider credentials that `aasm proxy stop` and `aasm proxy status` can no longer name or reach.

`check_bind_addr` now refuses port 0 **before** the loopback branch, since it disqualifies every address and `--allow-remote-clients` states an intent about reachability that says nothing about a recordable port. That also makes the two commands agree exactly — `verify_endpoint:365` already rejected a port-0 endpoint.

Mutation-checked rather than asserted: moving the new condition below the loopback `return Ok(())` lets `127.0.0.1:0` through and fails `port_zero_is_refused_on_every_address`. Ordering is load-bearing and the test proves it.

### Found, not fixed — both disclosed rather than quietly carried

- **The `aa-proxy` process is still exposable via `AA_PROXY_ADDR`**, which is how the container runs. Wiring the guard into `from_env()` would break the published image (`Dockerfile:41`), a CI compose file outside this lane, and shipped self-host docs — so it is deliberately out of scope. `check_bind_addr` is public and ready for that caller. The docs/example exposure is **AAASM-5370**.
- **Orphaned proxy on *any* failed start** is the general defect behind the port-0 case — refusing port 0 removes one deterministic route into it, not the path itself. Filed as **AAASM-5372**, linked; fixing it means changing the spawn lifecycle, outside this ticket's contract.

### Verification

`aa-proxy` + `aa-cli` **1209/1209** (was 1207 before the two new tests). 26 checks passing on `d294a0c1`, 0 failing, 0 pending. 13 mutations killed across both commits.

Merging as a merge commit. Admin only for the solo-maintainer approving-review requirement — nothing else is bypassed.

PR #1887, opened 2026-08-01
[AAASM-5351] ✅ (conformance): Add CJK and full-width false-positive vectors

Chisanan232:
## Description

Adds 8 golden vectors to `conformance/vectors/credential_detection/`, closing the gap that let the zh-TW false-positive defect ship and survive a full conformance run: on `main` before this PR, `rg '[\x{4e00}-\x{9fff}]' conformance/` returned **nothing** — all 26 vectors were ASCII, and the suite's only false-positive guard (`entropy_false_positive_clean.json`) is English prose whose whitespace tokens never reach the entropy gate at all.

The vectors describe **shipped** behaviour. Both prerequisites are merged:

* **AAASM-5344** (`95dbafe7`) — ASCII-run segmentation in the entropy pass
* **AAASM-5345** (`c2a2a75e`) — full-width digit normalisation for Luhn / SSN

| Vector | Pins |
|---|---|
| `entropy_false_positive_zh_tw_clean.json` | Clean Traditional-Chinese prose → no findings |
| `entropy_false_positive_zh_tw_en_mixed_clean.json` | Clean mixed zh-TW / English agent traffic → no findings |
| `entropy_false_positive_ja_clean.json` | Clean Japanese prose → no findings |
| `entropy_false_positive_th_clean.json` | Clean Thai prose → no findings |
| `entropy_cjk_adjacent_secret_still_detected.json` | **Bypass guard** — CJK flush against a synthetic ASCII secret, secret still detected |
| `pii_credit_card_fullwidth.json` | Full-width digit card → detected, exact span |
| `pii_ssn_fullwidth.json` | Full-width digit SSN → detected, exact span |
| `pii_fullwidth_digits_fail_luhn.json` | Full-width digits failing Luhn → **not** detected |

Japanese and Thai are in scope because the defect class is space-less scripts generally, not Chinese specifically — see the mutation evidence below for what each one actually constrains.

**ADR 0015** (`docs/src/adr/0015-dlp-trust-boundary-and-redaction-semantics.md`) — *"Do not change any committed conformance golden vector to make a detector change pass."* No pre-existing vector was edited: `git diff --diff-filter=MDRT --name-only 95dbafe7..HEAD` is empty, and the diff is 8 new files, 54 insertions, 0 deletions.

All fixtures are synthetic: the card is the standard Visa test number already in `pii_credit_card.json`, the SSN is the canonical documentation SSN already in `pii_ssn.json` (both reused so each new vector differs from its ASCII twin **only** in digit width), and the "secret" is a hand-typed alphanumeric string with no provenance.

## Type of Change

- [x] ✅ Tests (golden conformance vectors)

## Breaking Changes

- [x] No — additive only. New files, no edits to existing vectors, no schema change. SDK polyrepos pin a `conformance/` revision; adding vectors is backward compatible and existing pins keep working.

## Related Issues

- Related Jira ticket: [AAASM-5351](https://lightning-dust-mite.atlassian.net/browse/AAASM-5351) (backlog item B-5; depends on AAASM-5344 / AAASM-5345, both merged)
- ADR 0015 — *DLP Trust Boundary, Redaction Fail-Safety & Heuristic Detection Limits* (golden vectors are immutable)

## Testing

- [x] Unit tests added / updated (golden vectors)

```
$ cargo nextest run -p conformance -p aa-security
    Summary [   0.185s] 206 tests run: 206 passed, 0 skipped
```

### Mutation evidence — every vector was proved to discriminate

A vector that passes for the wrong reason is worse than none, so each one was run against a deliberately broken scanner and confirmed to fail. Mutations were applied to a local scratch copy of `aa-security/src/scanner.rs` and reverted; that file is byte-identical to `95dbafe7` on this branch (SHA-256 verified) and is **not** in the diff.

Results are per-vector; `[nn]` is the index in sorted vector order.

| # | Mutation | Result |
|---|---|---|
| 1 | Pass 1 scores the whole whitespace token instead of each ASCII run (**AAASM-5344 reverted**) | **4 FAIL** — `[14]` bypass guard, `[16]` Japanese, `[18]` zh-TW, `[19]` mixed. All 26 pre-existing vectors still PASS, which is precisely why the defect shipped. |
| 2 | Pass 1 skips any token containing a non-ASCII byte (the **naive fix AAASM-5344 rejected**) | **1 FAIL** — `[14]` bypass guard *only*. The clean vectors pass, because this "fix" does remove the false positives — while handing over a universal script-prefix evasion. This is the vector's entire reason for existing. |
| 3 | `ascii_digit_of` no longer maps U+FF10–U+FF19 (**AAASM-5345 reverted**) | **2 FAIL** — `[23]` full-width card, `[27]` full-width SSN. |
| 4 | `luhn_valid` always returns `true` (normalisation widens what reaches the checksum *and* bypasses it) | **1 FAIL** — `[25]` fail-Luhn vector *only*. Nothing else in the entire suite objects to a bypassed checksum. |
| 5 | `digit_segment` advances by the normalised ASCII width instead of the original char width (the classic span bug) | Panics on `[23]` (`byte index 6 is not a char boundary`) after `[00]`–`[22]` all pass — no pre-existing vector reaches the bug. |
| 6a | Entropy gate retuned 4.5 → 2.5, **ASCII-run narrowing kept** | `[17]` Thai **PASSES** (plus 7 pre-existing vectors fail — `api_keys_aws`, `api_keys_azure`, `db_urls_mysql`, `pii_credit_card`, `pii_email`, `private_keys_generic`, `private_keys_openssh` — identically in 6a and 6b, so the Thai delta is the only difference between the two). |
| 6b | Entropy gate retuned 4.5 → 2.5 **and** narrowing removed | `[17]` Thai **FAILS** (same 7 pre-existing vectors as 6a). |
| 7 | **CJK-codepoint-range exemption** — skip any token holding U+3040–30FF / 4E00–9FFF / 3400–4DBF / FF00–FFEF / 3000–303F, score the rest whole | **1 FAIL** — `[14]` bypass guard only. Thai **PASSES**: it clears the gate on entropy, not on script, so script handling cannot expose it. |

Reading 1 and 2 together is the point: mutation 1 is the defect, mutation 2 is the tempting fix for it, and only the full set of vectors rejects both.

**Honest note on the Thai vector.** Thai's UTF-8 shape — a constant `E0` lead byte and only two distinct second bytes — caps its byte entropy near 3.2 (measured per clause: 2.987 / 3.125 / 3.130 / 3.226), so Thai prose does **not** trip today's 4.5 gate even with AAASM-5344 fully reverted (mutation 1 leaves it passing). It therefore does not reproduce the shipped defect the way the Han vectors do. What 6a/6b establish is the property it *does* pin, which is still worth having: its 45-byte clause is inside the 20–64 length window, and it is the ASCII-run narrowing — not the gate value — that keeps it out of the gate's reach. It is the vector that breaks first if the gate is ever retuned downward — which makes it the tripwire for AAASM-5368.

A CJK-codepoint-range exemption does not cover Thai *by construction*, but at today's gate that is latent rather than exploitable — mutation 7 confirms Thai passes under exactly that fix. The vector file's description originally overstated this as "would still leave this one exposed", contradicting its own next sentence; corrected in `e0035104` (new commit, no history rewrite).

### SDK harness

`conformance/runner/runner.py` globs `*.json`, opens with an explicit `encoding="utf-8"` and reads only `description` / `input_text` / `expected_findings` / `expected_redacted` — it consumes the new vectors with no change. `.github/workflows/ci.yml` already routes on `conformance/**/*.json`.

**Pre-existing latent defect found, not fixed here** — tracked as [AAASM-5371](https://lightning-dust-mite.atlassian.net/browse/AAASM-5371) (outside this ticket's file scope): `runner.py::_redact` slices a Python `str` by what the vector schema documents as **byte** offsets. Character index and byte index coincide for ASCII, so this has never mattered; these are the first non-ASCII vectors in the suite. Verified directly — feeding the runner the byte offsets a *correct* SDK would report for `pii_credit_card_fullwidth.json` produces the unredacted input (`end` exceeds `len(str)`, so the splice is skipped) and the vector reports a spurious FAIL:

```
runner._redact with byte offsets : 'card=４５３２０１５１１２８３０３６６'
vector expected_redacted         : 'card=[REDACTED:CreditCardLuhn]'
MATCH: False
ascii vector MATCH: True
```

This breaks nothing today: the runner short-circuits to a pass unless `AA_SDK_MODULE` is set, and the SDK conformance CI jobs are still no-ops with `continue-on-error: true`. It will need fixing (decode to bytes, or have the SDK shim report char offsets) before those jobs can gate merges.

Also noted, not changed: `conformance/README.md` documents `credential_detection/` as "23 vectors". That count was already stale at `95dbafe7` (26 actual) and is now 34. Also covered by AAASM-5371.

## Checklist

- [x] Code follows project style guidelines — JSON-only diff; `lefthook` fmt/clippy/deny skip (no matching staged files)
- [x] Documentation updated if behaviour changed — n/a, no behaviour change
- [x] Commits are small and follow the Gitmoji convention — one vector per commit, 8 commits
- [x] Commits are signed off (`git commit -s`)

[AAASM-5351]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Chisanan232:
## Maintainer review — approve

Reviewed adversarially at `31357c19`, re-checked at `e0035104`. Every mutation claim in the body was reproduced independently in an isolated scratch copy (its `.git` pointer renamed so no git command could reach the real repo), mutating `scanner.rs` from a pristine snapshot per run. **All seven reproduced exactly** — several were understated, none overstated.

### The central claim holds

Mutation 1 — revert AAASM-5344 — fails exactly the 4 new script vectors (`[14]` bypass, `[16]` ja, `[18]` zh-TW, `[19]` mixed) with **`old_fail=0/26`**. That is the mechanical proof of why the zh-TW defect shipped undetected: the pre-existing suite could not have caught it, because it contains no non-ASCII text at all.

The bypass guard is genuinely load-bearing rather than decorative. Its secret splits on an interior `.` into 15- and 16-char base64 sub-runs, both under `BASE64_RUN_MIN_LEN`, so pass 3 cannot reach it — only run-segmented pass 1 can. Mutation 2 (the naive "skip non-ASCII tokens" fix) proves it: that guard is the *sole* failure.

### One durable artifact stated something false — fixed

The Thai vector's description claimed a CJK-codepoint-range exemption "would still leave this one exposed." An eighth mutation implementing exactly that fix was run: **Thai passes**. It cannot be exposed, because its entropy is below the gate regardless of script handling — which the vector's own next sentence said.

Corrected in `e0035104` as a new commit, no history rewrite. Worth recording that the author re-ran the mutation rather than accepting the finding on report, and observed the claim was refutable from data already in the PR: mutation 1 removes script handling entirely and Thai passed there too.

The vector stays. It asserts a real user-facing guarantee, it is the suite's only non-Han multi-byte script, and it is the tripwire that fires first if the 4.5 gate is ever retuned for AAASM-5368.

### Verified independently

- **26 pre-existing vectors byte-identical** — SHA-256 of every file at `95dbafe7` vs HEAD: 26 identical, 0 differing; `git diff --diff-filter=MDRT` empty; none deleted. ADR 0015 respected.
- **`scanner.rs` untouched** — SHA-256 identical at base and HEAD. The only files in the diff are the 8 new vectors plus the description correction.
- **All fixtures synthetic** — the only digit runs are the standard Visa test number, its check-digit-altered twin, and the canonical documentation SSN, all reused from existing vectors. No personal data, no Taiwanese national ID, no phone number.
- **Every declared offset lands on a correct UTF-8 char boundary**; all 34 vectors parse; key order matches the existing schema.
- Quoted entropy figures spot-checked numerically (zh-TW 4.628/4.815/4.853; ja 63 B at 4.812; Thai max 3.226).

### Noted, not blocking

Mutation 6a is not a clean experiment — gate 2.5 also breaks 7 pre-existing vectors. The 6a/6b differential remains valid because the same 7 fail in both, so the Thai delta is the only difference; now disclosed in the table.

The `conformance/runner/_redact` byte-vs-codepoint defect found during this work is tracked as **AAASM-5371** and is being fixed separately. It is why AC4 passes in the letter today: the runner globs and parses the new vectors fine, but would mis-slice them. Deliberately not folded into this PR.

Merging on green.

PR #1888, opened 2026-08-01
[AAASM-5371] 🐛 (conformance): Slice vector offsets as bytes, not code points

Chisanan232:
## Description

`conformance/runner/runner.py::_redact` sliced a Python `str` with what the vector
schema defines as **byte** offsets. Python `str` indices are code points, so the
two units only coincide on ASCII — which every vector in the suite was until
AAASM-5351 landed the first non-ASCII ones. On any multi-byte input the redaction
lands in the wrong place or silently does not happen, and the harness reports a
spurious FAIL against a *correct* SDK.

`_redact` now splices the UTF-8 encoding and decodes once at the end. That form
was chosen over converting byte offsets to code-point offsets because it is the
same operation `ScanResult::redact` performs on the Rust side, so the harness
stays byte-for-byte comparable with the implementation it grades instead of
approximating it through a second unit. Spans that miss a character boundary are
now rejected rather than spliced, so a bad span cannot produce invalid UTF-8 or
leave a fragment of the flagged value in the output.

Also in this PR: the offset unit is now stated normatively in
`conformance/README.md` with the per-language consequence spelled out (Go indexes
bytes and is correct as-written; Node/TS indexes UTF-16 code units and is not),
and the stale `credential_detection/` vector count is corrected.

Scope is `conformance/runner/` plus those `conformance/README.md` edits. **No
vector file is touched** — ADR 0015 forbids editing a committed vector to make
anything pass, and that rule is exactly what this defect would have eroded.

## Type of Change

- [ ] ✨ New feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactoring
- [x] 📝 Documentation update
- [ ] ⬆️ Dependency upgrade
- [ ] 🚀 Release

## Breaking Changes

- [x] No
- [ ] Yes (describe below)

Test harness only — no production code path, no schema change to the vector files,
no SDK contract change. ASCII behaviour is provably unchanged (see below), so any
SDK currently pinning this `conformance/` revision is unaffected.

## Related Issues

- Related Jira ticket: [AAASM-5371](https://lightning-dust-mite.atlassian.net/browse/AAASM-5371)
- Parent Epic: [AAASM-5270](https://lightning-dust-mite.atlassian.net/browse/AAASM-5270) — Local-first sensitive-data intelligence and enforcement architecture
- Vectors that exposed this: [AAASM-5351](https://lightning-dust-mite.atlassian.net/browse/AAASM-5351) (#1887, merged — this branch is rebased on it)

## Testing

### 1. Mutation proof — the new regression tests fail without the fix

A test that passes both before and after proves nothing, so here is the same test
file run against both runners.

**Before** (`conformance/runner/runner.py` at `remote/main`):

```
test_finding_without_end_is_skipped ... ok
test_no_findings_returns_input_unchanged ... ok
test_single_ascii_span ... ok
test_multibyte_prefix_shifts_the_span ... FAIL
test_multibyte_value_body ... FAIL
test_multiple_spans_with_multibyte_separator ... FAIL
test_out_of_range_span_is_skipped ... ok
test_span_inside_a_character_is_skipped_not_mangled ... FAIL

FAIL: test_multibyte_prefix_shifts_the_span
AssertionError: 'メモ token=sk-DUMMY-NOT-A-REAL-KEY' != 'メモ token=[REDACTED:SyntheticKey]'

FAIL: test_multibyte_value_body
AssertionError: '番号=１２３４５６７８' != '番号=[REDACTED:SyntheticNumber]'

FAIL: test_multiple_spans_with_multibyte_separator
AssertionError: '一=DU[REDACTED:SyntheticKey]=DUMMY-VALUE-TWO' != '一=[REDACTED:SyntheticKey]、二=[REDACTED:SyntheticToken]'

FAIL: test_span_inside_a_character_is_skipped_not_mangled
AssertionError: '番[REDACTED:SyntheticNumber]4' != '番号=1234'

Ran 8 tests in 0.002s

FAILED (failures=4)
EXIT=1
```

**After** (this branch):

```
test_finding_without_end_is_skipped ... ok
test_no_findings_returns_input_unchanged ... ok
test_single_ascii_span ... ok
test_multibyte_prefix_shifts_the_span ... ok
test_multibyte_value_body ... ok
test_multiple_spans_with_multibyte_separator ... ok
test_out_of_range_span_is_skipped ... ok
test_span_inside_a_character_is_skipped_not_mangled ... ok

Ran 8 tests in 0.000s

OK
EXIT=0
```

The fixtures are synthetic and declared inline in the test file rather than loaded
from `vectors/` — the harness has to be provable independently of which vectors
happen to be committed, and no assertion diff can then print real credential
material. The three `ok`-before/`ok`-after cases are ASCII guards, not evidence;
they are labelled as such in the file.

### 2. The 8 AAASM-5351 vectors, end to end through the runner

Per the ticket's verification plan, driven with `AA_SDK_MODULE` set to an oracle
stub that reports correct byte offsets.

**Before** (`remote/main` runner) — 5/8, and the 3 failures are precisely the
vectors with a multi-byte span:

```
Results: 5/8 passed, 3 failed
  FAIL [AAASM-5344 bypass guard: ... ASCII high-entropy secret flush against Chinese text ...]: redact mismatch
  FAIL [AAASM-5345: ... Visa test number ... in full-width digits ...]: redact mismatch
  FAIL [AAASM-5345: ... documentation SSN ... in full-width digits with ASCII separators ...]: redact mismatch
EXIT=1
```

(The `got:`/`expected:` lines are omitted here deliberately; the failure identity
is the vector, not its payload.)

**After** (this branch):

```
Results: 8/8 passed  all passed
EXIT=0
```

### 3. The 26 pre-existing ASCII vectors: unchanged for every well-formed span

Not just "still pass" — the old code-point-slicing `_redact` and the new
byte-slicing one were run head-to-head over **every** span position in each of the
26 ASCII vector inputs (single spans at every `(offset, end)` pair, two-span
combinations, plus out-of-range / inverted / missing-`end` cases).

The sweep is committed as `conformance/runner/check_redact_equivalence.py` so the
numbers below are reproducible rather than something to take on trust:

```
$ python conformance/runner/check_redact_equivalence.py
ASCII vectors swept        : 26
non-ASCII vectors skipped  : 8
well-formed span cases     : 994838
  mismatches               : 0
malformed (offset < 0) now rejected instead of mangled: 6528
OK: identical for every well-formed span.
```

**Precise claim: unchanged for all well-formed spans; malformed negative offsets
are now rejected instead of silently mangled.** That second bucket is a real
behaviour change and it is a fix, not a regression — the old code let `offset: -1`
fall through to Python's negative indexing and spliced a placeholder into the
middle of the text, returning duplicated garbage; the guard added here rejects it.
An earlier draft of this PR said "byte-identical before and after" on the strength
of a sweep whose generator never emitted a negative offset. It is corrected above,
and the committed sweep now enumerates that bucket explicitly instead of leaving
it outside the search space.

A differential harness is the specific kind of test that passes for the wrong
reason — compare two things that are quietly the same object and you get zero
mismatches forever. So the sweep has to show it can still fail:

```
$ python conformance/runner/check_redact_equivalence.py --self-check
self-check: sabotaged implementation produced 994708 mismatches
OK: the sweep discriminates.
```

Driving the runner over those 26 vectors with the oracle stub gives `22/22 passed`
on both old and new runners — 22, not 26, because 4 vectors are excluded for an
unrelated pre-existing defect described below.

### 4. Suites

```
cargo nextest run -p conformance    → 49 tests run: 49 passed, 0 skipped   (exit 0)
python conformance/runner/test_runner_redact.py                            (exit 0)
python conformance/runner/runner.py    (CI's default invocation)           (exit 0)
```

Exit codes were captured from the command itself, not from a pipe.

### Node / Go harness audit (acceptance criterion)

**Neither harness exists.** `conformance/runner/` contains only `runner.py`; the
`conformance-node` and `conformance-go` CI jobs are `echo`-and-`exit 0`
placeholders that literally say "Add a `conformance/runner/runner.js` and wire it
up here." The `*conformance*` files in the `node-sdk` / `go-sdk` repos are
fail-open governance-gate tests, unrelated to credential-detection vectors.

So the defect cannot be present today, but it is pre-loaded for whoever writes
those harnesses, and the two languages are not equally exposed:

| Language | Native string index | Correct against this schema? |
|---|---|---|
| Go | bytes (`s[a:b]`) | yes, directly |
| Node/TS | UTF-16 code units (`s.slice(a, b)`) | **no** — must splice a `Buffer`/`Uint8Array` |

Node is the live hazard: `slice()` is wrong for *any* multi-byte character, and
like Python it will pass 100% of the ASCII vectors while doing so. That table is
now in `conformance/README.md` so the next harness author meets it before writing
the splice.

## Checklist

- [ ] All CI checks passing
- [x] Commits are signed off (`git commit -s`) — DCO sign-off is optional/advisory here, not enforced

No Rust source changed, so lefthook's `fmt`/`clippy`/`deny` hooks reported
"no matching staged files" on every commit rather than running.

## Out of scope — filed for follow-up, deliberately not fixed here

**`_redact` does not coalesce overlapping spans, so 4 committed vectors cannot
pass.** Rust's `ScanResult::redact` runs `coalesce_findings` first, merging
overlapping findings into a single label; the Python runner splices each finding
separately. For `db_urls_postgres`, `db_urls_mongodb`, `db_urls_mysql` and
`private_keys_ec_short_trailing_line`, `expected_findings` lists 2–3 findings
while `expected_redacted` contains exactly one label — so **the spans a correct
scanner reports cannot reproduce it**, whatever unit those spans are in.

(Stated precisely, because the stronger phrasing is falsifiable: for
`private_keys_ec_short_trailing_line` there do exist contrived `end` pairs that
happen to reconstruct the expected string, artefacts of the bound being checked
against the mutated buffer. They are not the spans the scanner emits. Three of
the four admit no solution at all.)

This is pre-existing, independent of the byte/code-point bug, and equally latent
(it also only bites once `AA_SDK_MODULE` is set and the jobs gate merges). It is
excluded from the 22/22 run above rather than papered over. Same blast radius as
this ticket: a correct SDK failing on a harness defect. Tracked as **AAASM-5373**.

Two smaller ones, same file:

- **`_redact` skips an unspliceable span; Rust fails closed** and returns
  `"[REDACTED]"` for the whole text rather than risk emitting the flagged region
  (`aa-security/src/scanner.rs:447-458`). This diff *adds* three such fail-open
  paths — the `offset < 0` guard and both `_is_utf8_boundary` checks — so the
  divergence widens here even though the offset unit is fixed. Deferred to
  AAASM-5373 rather than implemented, but deliberately not written down as
  correct: the two tests that pin it are named
  `..._pending_fail_closed`, carry the Rust line and the ticket in a comment, and
  the docstring no longer claims this function mirrors `ScanResult::redact`.
- **Nothing runs `test_runner_redact.py` or the sweep in CI**, and the
  `conformance-python` job that does run is a no-op twice over: beyond
  `continue-on-error: true`, `run()` short-circuits to `return True` at
  `runner.py:127-133` because `AA_SDK_MODULE` is unset, so it never loads a
  vector. The workflow comment at `ci.yml:1173-1174` saying the runner "exits 1
  until a real SDK implementation is wired up" is wrong — it exits 0. Wiring any
  of this up needs a `ci.yml` edit, outside this PR's scope. Tracked as
  **AAASM-5374**.

Separately: the repo configures **no Python linting at all** — no `ruff`/`mypy`
config, no `pyproject.toml`, no `.pre-commit-config.yaml`, and no Python linter in
any workflow. `conformance/runner/*.py` is unlinted by construction. All three
files were checked with `python -m py_compile`.

[AAASM-5371]: https://lightning-dust-mite.atlassian.net/browse/AAASM-5371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -20,7 +20,7 @@ conformance/
 │   ├── message_serialization/  # 10 vectors: one per proto message golden
 │   ├── proto/                  # *.bin golden files (prost encode_to_vec output)
 │   ├── policy_query/           # 10 vectors: ALLOW, DENY, PENDING, REDACT decisions
-│   ├── credential_detection/   # 23 vectors: API keys, auth tokens, DB URLs, PII, entropy
+│   ├── credential_detection/   # 34 vectors: API keys, auth tokens, DB URLs, PII, entropy
 │   └── session_lifecycle/      # 10 vectors: Register, Heartbeat, Deregister, ControlCommand
 └── runner/
     ├── requirements.txt        # Python runner dependencies (colorama)
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -25,7 +25,8 @@ conformance/
 └── runner/
     ├── requirements.txt        # Python runner dependencies (colorama)
     ├── runner.py               # Python SDK conformance runner
-    └── test_runner_redact.py   # Regression tests for the runner's own redaction logic
+    ├── test_runner_redact.py   # Regression tests for the runner's own redaction logic
+    └── check_redact_equivalence.py  # Differential sweep vs. the pre-AAASM-5371 redaction
 ```
 
 ## Test categories
@@ -162,6 +163,21 @@ without an SDK and without any vector file:
 ```bash
 python conformance/runner/test_runner_redact.py
 ```
+
+Changing how spans are spliced is the change most likely to break the 26 ASCII
+vectors without any of them failing, since a wrong unit still lands in the right
+place on ASCII. The differential sweep drives the current implementation and the
+pre-AAASM-5371 one over every span position in every ASCII vector and requires
+byte-identical output:
+
+```bash
+python conformance/runner/check_redact_equivalence.py
+python conformance/runner/check_redact_equivalence.py --self-check
+```
+
+`--self-check` sabotages the implementation by one character and asserts the
+sweep notices — a differential harness that silently compares one thing to itself
+reports zero mismatches forever and looks exactly like proof.
 
 ## Adding new vectors
 
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -24,7 +24,8 @@ conformance/
 │   └── session_lifecycle/      # 10 vectors: Register, Heartbeat, Deregister, ControlCommand
 └── runner/
     ├── requirements.txt        # Python runner dependencies (colorama)
-    └── runner.py               # Python SDK conformance runner
+    ├── runner.py               # Python SDK conformance runner
+    └── test_runner_redact.py   # Regression tests for the runner's own redaction logic
 ```
 
 ## Test categories
@@ -81,6 +82,27 @@ Vector schema:
 }
 ```
 
+#### Offset unit — normative
+
+`expected_findings[].offset` is a **byte offset into the UTF-8 encoding of
+`input_text`**, counting from zero. So is the `end` an SDK reports alongside it.
+This is not negotiable per language: it is the unit `CredentialScanner` emits,
+and the vectors are the same bytes for every SDK.
+
+It matters because each language's native string index means something different,
+and for ASCII input all three units coincide — so a harness that picks the wrong
+one still passes every ASCII vector and only breaks on the first multi-byte one:
+
+| Language | Native string index | Correct against this schema? |
+|---|---|---|
+| Rust | bytes (`&str[a..b]`) | yes, directly |
+| Go | bytes (`s[a:b]`) | yes, directly |
+| Python | code points (`s[a:b]`) | **no** — encode to `bytes`, splice, decode back |
+| Node/TS | UTF-16 code units (`s.slice(a, b)`) | **no** — splice a `Buffer`/`Uint8Array` instead |
+
+An offset that does not fall on a character boundary is not redactable; reject it
+rather than splicing, so no harness can emit invalid UTF-8 or a partial value.
+
 Categories: API keys (Anthropic, OpenAI, AWS, GCP, Azure), auth tokens (GitHub,
 Slack), database URLs (Postgres, MySQL, MongoDB), private keys (RSA, EC, OpenSSH,
 PKCS8, PGP), PII (credit card, SSN, email), high-entropy tokens.
@@ -126,6 +148,20 @@ The `scan()` function must return a list of dicts, each with:
 - `"kind"` (str) — credential kind string matching `CredentialKind.as_str()`
 - `"offset"` (int) — byte offset of the finding in the input text
 - `"end"` (int) — byte end of the matched region (used for redaction)
+
+Both positions are byte offsets into the UTF-8 encoding of the input, per
+[Offset unit — normative](#offset-unit--normative). A Python SDK that reports
+`str` indices will pass every ASCII vector and fail every non-ASCII one.
+
+### Testing the runner itself
+
+The runner reconstructs the redacted string from the spans an SDK reports, so it
+can be wrong in exactly the same way an SDK can. Its own regression tests run
+without an SDK and without any vector file:
+
+```bash
+python conformance/runner/test_runner_redact.py
+```
 
 ## Adding new vectors
 
--- a/conformance/runner/.gitignore
+++ b/conformance/runner/.gitignore
@@ -1,0 +1,2 @@
+# Byte-code caches from running the conformance runner and its tests locally.
+__pycache__/
--- a/conformance/runner/check_redact_equivalence.py
+++ b/conformance/runner/check_redact_equivalence.py
@@ -1,0 +1,227 @@
+#!/usr/bin/env python3
+"""
+Differential check: the byte-slicing `_redact` vs. the code-point-slicing one it replaced.
+
+Why this exists
+---------------
+AAASM-5371 changed how `runner._redact` interprets a finding's span. The claim
+that came with it — "ASCII behaviour is unchanged" — is the kind of claim that is
+easy to assert and easy to get wrong, because every committed ASCII vector passes
+either way. This script is the evidence for it, kept runnable so the number in the
+PR is reproducible rather than something you have to take on trust.
+
+It drives both implementations over every span position in every ASCII vector and
+compares their output byte for byte.
+
+    well-formed spans (0 <= offset <= end)   must be identical      → else exit 1
+    malformed spans   (offset < 0)           are expected to differ → reported only
+
+The second bucket is the one honest exception. The old code let a negative offset
+fall through to Python's negative indexing and spliced garbage into the middle of
+the text; the new code rejects it. That is a fix, not a regression, so it is
+counted and printed rather than asserted away.
+
+Self-check
+----------
+A differential harness that compares two things which are quietly the same object
+reports zero mismatches forever and looks like proof. `--self-check` sabotages the
+new implementation by one character and asserts this sweep *notices*, so a clean
+run means the sweep can discriminate rather than that it is inert.
+
+Run
+---
+    python conformance/runner/check_redact_equivalence.py
+    python conformance/runner/check_redact_equivalence.py --self-check
+
+Needs no SDK, sets no environment, and reads the vectors without writing to them.
+"""
+
+from __future__ import annotations
+
+import argparse
+import itertools
+import json
+import sys
+from pathlib import Path
+from typing import Any, Callable
+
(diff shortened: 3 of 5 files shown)


Chisanan232:
## Maintainer review — approve after changes, now applied

Reviewed adversarially in an isolated clone (origin remote removed, no path back to the checkout). Every claim re-derived from scratch rather than taken on report.

### The fix is correct and the evidence is unusually honest

- **Mutation proof exact.** 4 tests fail against `main`'s runner, and **all 4 fail as `AssertionError` on wrong output — zero are exceptions** (`grep -c '^ERROR:'` = 0). That distinction matters: a test that "fails" by crashing may not pin what it claims to pin. These do.
- **No vector edited — proven at the strongest available level.** The `conformance/vectors` **tree object is byte-identical** at base and head (`0c0aae69…` both sides), which is stronger than per-file hashing. `git diff --diff-filter=MDRT` over the whole tree returns only `README.md` and `runner.py`. ADR 0015 satisfied.
- **`_redact` really is the only affected site.** Independently swept: the only str/byte indexing in `runner.py` is there; `_findings_match` compares ints; `run()` never indexes; no other `.py` in the repo reads the vectors.
- **The equivalence sweep has real discriminating power.** Verified by negative control: sabotaging the new implementation by one character produced 69,002 / 69,080 mismatches in an independently-built harness. Not theatre.
- **`conformance::credential_detection all_vectors_redact_correctly` passes** in the `Test` job — which independently establishes that the vectors are right and the Python runner was the deviant, rather than the reverse.
- Fixtures synthetic throughout; no credential or PII material in the tests, PR body, or any of the commit messages.

### The one finding that had to be fixed before merge

The two tests asserting that an unspliceable span is **skipped** codified an ADR-0015 fail-open as expected behaviour — while this same diff *added* three new fail-open paths where Rust fails closed (`scanner.rs:447-458`: "never return the raw text with a secret intact"). Deferring fail-closed is fine and was stated; writing it down as correct is not. Anyone implementing fail-closed later would have had to delete two green tests to do it — the exact erosion dynamic AAASM-5373 was filed against.

Now fixed: both renamed to `..._pending_fail_closed`, each citing `scanner.rs:454-458` and AAASM-5373, stating they pin *current harness behaviour, not desired behaviour*. The self-contradicting `(mirrors Rust ScanResult::redact)` docstring is gone — **and the author found a second instance of the same false claim at `runner.py:160` that this review missed.**

### Two overstated numbers, corrected

- "0 mismatches / ASCII behaviour provably unchanged" was false at the edges: an independent 242,605-case sweep found **26 mismatches**, one per ASCII vector, all on `offset: -1`, where the old code applied Python negative indexing and produced mangled output. Every one is an improvement — but the `offset < 0` guard was added in this diff and the generator never produced a negative offset, so the claim was unfalsifiable by its own evidence. Now restated precisely, with that bucket enumerated (6,528 cases) rather than left outside the search space.
- "No per-finding splice can reproduce it, *whatever unit*" was falsifiable: `private_keys_ec_short_trailing_line` admits 70 contrived solutions, artefacts of the bounds check running against the partially-spliced buffer. Three of the four are genuinely impossible and the spans a *correct* scanner reports still don't work, so the conclusion stands. Now worded to match.

### The sweep is committed

The 994,838-case figure was the PR's strongest evidence and was unreproducible — the reviewer built their own and could not reconstruct the combination scheme. It now ships as `conformance/runner/check_redact_equivalence.py` with the scheme documented, plus a `--self-check` that sabotages the implementation and asserts the sweep notices (994,708 mismatches). That guard is the point: a differential harness that ends up comparing one implementation to itself reports zero mismatches forever and is indistinguishable from proof.

### What this PR's regression protection is actually worth today: nothing

Stated plainly because the PR states it plainly. `conformance/runner/` is triple-latent in CI — `conformance-python` carries `continue-on-error: true`, nothing invokes `test_runner_redact.py`, and `run()` short-circuits to `return True` at `runner.py:127-133` because `AA_SDK_MODULE` is unset, so it never loads a vector. The `ci.yml:1173-1174` comment claiming the runner "exits 1" is wrong; it exits 0.

So the validation behind this merge is the author's local runs and the committed sweep, not a pipeline. Tracked as **AAASM-5374**. The coalescing and fail-closed work is **AAASM-5373**. Neither folded in here.

Merging once checks settle.

PR #1891, opened 2026-08-01
[AAASM-5348] 📝 (docs): Document the proxy bind refusal and --allow-remote-clients

Chisanan232:
## Description

`aasm proxy start` changed a shipped CLI contract in #1886 (AAASM-5348) and added a flag, and the reference page was not touched. This documents both.

Three gaps, all in `docs/src/cli/proxy.md`:

- **`--allow-remote-clients` appeared nowhere in `docs/`.** A new flag, undocumented.
- **`--listen` was documented with no mention that a non-loopback value is now refused** — a breaking change an operator would meet as an unexplained error.
- **Port 0 is refused too**, and for a reason worth stating: the child binds a real port that nothing records, the five-second wait on port 0 fails, the pid file is removed, and the proxy keeps running with nothing able to name or stop it.

## The refusal text is verbatim, not paraphrased

The point of naming the two missing protections in the refusal is that the operator reads them, so the console block reproduces the `Display` impl exactly. Machine-diffed against `aa-proxy/src/config.rs:283-289` after normalising whitespace and substituting `{addr}` / `{}` — **exact match**. If that string is ever reworded, this page should be reworded with it.

## What this page now says that it did not

The guard lives in `aasm proxy start`. Running the `aa-proxy` binary directly with a non-loopback `AA_PROXY_ADDR` — which is exactly what `aa-proxy/Dockerfile:41` and `examples/docker-compose/docker-compose.yml` do — **still binds it**. Leaving that out would let the reference page imply a protection that is not there for the deployment shape most likely to be exposed. Called out with a pointer to **AAASM-5370**, which owns that exposure.

It also directs operators who genuinely need to reach another machine's proxy at SSH port-forwarding, rather than leaving "not supported" as the whole answer.

## Type of Change

- [x] 📝 Documentation

## Breaking Changes

- [ ] No — documents a breaking change made in #1886; changes no behaviour itself.

## Related Issues

- Jira ticket: AAASM-5348 (documentation follow-up to #1886, `80f8ef01`)
- References AAASM-5370

## Testing

`mdbook build` succeeds. `scripts/check-doc-orphans.sh` clean. `scripts/check-doc-links.sh docs/src/cli/proxy.md` — all internal links resolve. The refusal-string diff described above is the substantive check; `mdbook` would happily build a page that misquotes the binary.

## Checklist

- [ ] All CI checks passing

## Note on how this was found

Not by review of #1886 — I merged that PR without noticing its docs were missing. It surfaced afterwards, checking whether the two merges that landed minutes apart (`5906f3b4` docs, `80f8ef01` code) were mutually consistent. A three-code-file diff with no docs change is exactly the shape that passes a checks list and leaves a contract undocumented.

### DIFF
--- a/docs/src/cli/proxy.md
+++ b/docs/src/cli/proxy.md
@@ -32,7 +32,8 @@ attacker-planted `aa-proxy`.
 
 | Flag | Type | Default | Description |
 |---|---|---|---|
-| `--listen <LISTEN>` | string | `127.0.0.1:8899` (env `AA_PROXY_ADDR`) | Address the proxy listens on. |
+| `--listen <LISTEN>` | string | `127.0.0.1:8899` (env `AA_PROXY_ADDR`) | Address the proxy listens on. **Must be loopback with a named port** — see below. |
+| `--allow-remote-clients` | flag | off | State that a non-loopback `--listen` is intended. Does not currently permit one — see below. |
 | `--gateway <GATEWAY>` | string | env `AA_GATEWAY_URL` | Gateway URL to forward policy decisions to. |
 | `--ca-dir <CA_DIR>` | path | env `AA_CA_DIR` | Directory for CA certificate and key storage. |
 | `--no-detach` | flag | off | Run in the foreground instead of daemonizing. |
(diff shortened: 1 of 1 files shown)


Chisanan232:

Docs-only, but it makes a security claim about `aa-proxy`, so the claim is what I checked rather than the prose.

**The refusal block is verbatim, and verified so mechanically.** I extracted the `Display` impl string from `aa-proxy/src/config.rs:283-289`, joined its continuation lines, substituted `{addr}` and `{}`, normalised whitespace, and diffed it against the fenced block in the page — exact match. A paraphrase would have been the wrong choice here: the entire purpose of that message is to name the two absent protections, so the page has to carry them as the binary emits them.

**The "neither protection exists" claim was verified in source, not taken from the PR that introduced it.** Listener TLS: `aa-proxy/src/proxy/mod.rs:255` binds a bare `TcpListener` and speaks plain `CONNECT`; the `rustls` configs nearby are the per-host MitM leaf presented *inside* a tunnel and the upstream client config, neither of which protects the listener. Client auth: no `Proxy-Authorization`, no `407`, no peer check in the crate.

**The `AA_PROXY_ADDR` carve-out is the part that most needed saying.** Without it this page would read as "non-loopback binds are refused" full stop, while `aa-proxy/Dockerfile:41` and `examples/docker-compose/docker-compose.yml` bind `0.0.0.0` through the env path the guard does not cover — i.e. the reference would imply a protection that is absent in precisely the deployment shape most likely to be exposed. Pointed at AAASM-5370, which owns that exposure.

`mdbook build` clean, `check-doc-orphans.sh` clean, `check-doc-links.sh` run **with the changed file as an argument** — it takes files as arguments, so a bare invocation checks nothing and reports success, which is a non-result I have already been caught by once in this batch.

### Why this PR exists at all

I merged #1886 — a breaking `--listen` contract change plus a new flag — without noticing it touched no documentation. The checks list was green and complete about what it measured. This is the same failure shape the programme has been closing all along, and it went past me in my own review, which is worth stating in the record rather than fixing quietly.

Merging as a merge commit; admin solely for the solo-maintainer approving-review requirement.

PR #1896, opened 2026-08-02
[AAASM-5349] ♻️ (aa-policy): Extract policy document semantics into a leaf crate

Chisanan232:
## Description

First of three PRs for **AAASM-5349**, implementing the owner's chosen route (service authority): the developer-integration service must be able to resolve a policy, so the semantics that let it do so move below both crates that need them.

This PR is the move. It changes no behaviour.

## Why a new crate, rather than a dependency

AC 6 requires **audit, receipt, `status` and `verify`** to distinguish the four policy states. `status` and `verify` are rendered from what the service computed — `aa-cli/src/commands/integrations/status.rs` says so outright, *"Nothing is computed locally. `aasm` renders what the service derived (ADR 0030 forbidden design 10)"* — so the service in `aa-runtime` has to resolve the policy itself.

It could not. The resolver needs `aa_gateway::policy::PolicyValidator`, and **`aa-runtime` does not depend on `aa-gateway`**. Adding that dependency would invert the layering: the runtime is the enforcement point, the gateway is the brain above it, and the gateway is large (server, IAM, budget, storage) in a way the runtime has no business compiling.

So the semantics move down instead. Parsing, validation, the document model, the expression evaluator and version history reference nothing from `aa-gateway` — only `aa_core`, `aa_security` and third-party crates — so they become **`aa-policy`**, a leaf beside `aa-core` and `aa-security`.

## Where the split falls, and why there

`PolicyContext` — the contract behind topology-aware condition variables (`agent.depth`, `team.active_agents`) — moves **with** the evaluator that calls it, because it names nothing but `aa_core` types. Its production implementation, `ProductionPolicyContext`, **stays** in `aa-gateway`, which owns the `AgentRegistry` and `BudgetTracker` it reads.

That one boundary is the entire reason a leaf crate is possible: policy *semantics* do not depend on who holds the runtime state. `FakePolicyContext` moved with the trait it doubles.

## No caller changes

`aa-gateway/src/policy/mod.rs` is now a re-export shim (`pub use aa_policy::*;` plus the local `context`), so every existing `aa_gateway::policy::…` path resolves exactly as before. **Zero call sites edited.**

## Type of Change

- [x] ♻️ Refactor (no functional change)

## Breaking Changes

- [ ] No. Public paths through `aa_gateway::policy` are unchanged. `aa-policy` is a new published crate; three items widen visibility (`canonical`, `expr`, `references_live_context`) only because a crate boundary now sits where a module boundary did — the callers are the same callers.

## Related Issues

- Jira ticket: AAASM-5349 (PR 1 of 3)
- Follows: DI-API v3 carrying policy posture on `StatusView`/`VerificationView`; then the audit variant

## Testing

**The load-bearing check for a pure extraction is that no test was lost in the move**, since a move that silently drops a `mod tests` still compiles and still passes everything that remains:

```
baseline (remote/main), test fns under aa-gateway/src/policy/ : 305
branch,  aa-policy/src (299) + aa-gateway/src/policy (6)      : 305
```

Equal. My first attempt at this count returned `0` from a bad `git grep` pathspec — recorded here because a `0` that looks like "nothing to compare" is exactly the shape of a non-result.

- `cargo nextest run -p aa-policy -p aa-gateway` — **1482/1482 pass** (2 pre-existing `cascade_hot_reload_*` flakes, green on retry)
- `cargo build --workspace` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — passes
- `cargo fmt --all` — applied
- `git diff --find-renames` detects **17 renames**; `error.rs` and several `history/*` files move with a zero-line diff

## Checklist

- [x] Documentation updated if behaviour changed (crate/module docs state why the split exists)
- [ ] All CI checks passing

## Note on commit granularity

This is one commit for ~6,290 moved lines, against the usual preference for smaller units. A crate extraction cannot be split without leaving intermediate commits that do not build, and `.claude/CLAUDE.md` requires every commit to be bisectable. Bisectability wins; the move is atomic.


Chisanan232:
## Blocked — `Packaged-artifact gate` fails, and the gate is right

Do not merge. The failure is real, not flaky, and I reproduced it exactly rather than inferring it.

The gate's log only says `cargo package -p aa-gateway failed`, because `scripts/check-packaged-artifacts.sh:193` runs the command with `>/dev/null 2>&1` and reports the exit status alone. Rebuilding the gate's own tree locally (`git ls-files` copy + `.ci/strip-for-publish.sh`, exactly as the script does) surfaces the actual cause:

```
error: failed to prepare local package for uploading

Caused by:
  no matching package named `aa-policy` found
  location searched: crates.io index
  required by package `aa-gateway v0.0.1-rc.6`
```

`aa-policy` itself packages fine — the log shows `✓ packaged aa-policy`. What fails is every **already-published** crate that now depends on it: the gate packages at step 3 but injects `[patch.crates-io]` only at step 5, so `aa-policy 0.0.1-rc.6` is resolved against the crates.io index, where it does not exist and cannot until a release publishes it.

**This is the gate doing its job.** It exists (AAASM-5315/5316) to catch "a crate that could not compile for any consumer", and a dependency on an unpublished crate is precisely that — for anyone consuming the tarball rather than the workspace.

**The real publish is not affected.** `release.yml` publishes in dependency order, so `aa-policy` would land before `aa-gateway` and resolve normally. The gap is only in the gate, which packages everything from a tree without publishing anything.

### Why there is no in-PR fix

`aa-auth` is the only comparable precedent — a leaf crate extracted the same way — but it was introduced before this gate existed and was already on crates.io by rc.6. **This is the first new internal crate since the gate landed**, so nothing establishes how one is meant to be introduced.

The three ways forward are a release action, a change to release infrastructure, or a different crate topology. None of them belongs in this PR, and I am not choosing between them unilaterally — escalated to the owner.

The extraction itself is sound and stays green: workspace builds, `clippy -D warnings` passes, **1482/1482 tests**, and **305 = 305** test functions before and after, so nothing was lost in the move.

Chisanan232:

The blocker is resolved at the source: **`Packaged-artifact gate → pass`** on `ace3e47f`, the same gate that failed this PR at `10ef7274`. It was fixed in #1897 (AAASM-5432, `ffeb2955`) rather than worked around, and `Published-surface coherence gate` also passes. 45 checks pass, 0 fail, 0 pending.

Refreshed **without a force push**: `main` was merged into the pushed tip via a temporary branch rather than rebasing onto it, so the published commit stays an ancestor and the push fast-forwards. Nothing was reset or discarded.

### A defect in this PR that my own earlier verification missed

I reported **1482/1482 passing** for the extraction. That was true and proved less than I claimed. It came from `cargo nextest run -p aa-policy -p aa-gateway`, and cargo unifies features across the crates named in one invocation — `aa-gateway` declares `tokio = { features = ["full"] }`, which was silently supplying `time` to `aa-policy`.

Built alone, **`aa-policy`'s tests did not compile at all**: ten × `cannot find 'time' in 'tokio'`, because `history/fs_store.rs` sleeps between writes to force distinct mtimes and the dev-dependency never declared the feature.

A leaf crate whose entire purpose is to stand alone, verified only in company. Fixed by declaring `"time"`; **299/299 standalone**, 1482/1482 combined. The workspace root declares plain `tokio = "1"` with no features, so every crate must declare its own — which is what made this reachable.

A workspace-wide scan for the same shape across all 32 `aa-*` crates found **no other instance**. The near-miss is `aa-cache`, whose `src/testing.rs` calls `tokio::time::sleep` with only `sync` in `[dependencies]`; it survives solely because someone wrote `test-utils = ["tokio/time"]` deliberately.

### Extraction evidence, restated

* **305 = 305** test functions before and after the move (299 in `aa-policy`, 6 with the gateway-backed `ProductionPolicyContext`) — the check that a move which silently drops a `mod tests` would fail
* `cargo build --workspace` clean; `clippy --workspace --all-targets -- -D warnings` passes
* 17 renames detected by `git diff --find-renames`; `error.rs` and several `history/*` files move with zero-line diffs
* **Zero call sites edited** — `aa_gateway::policy` re-exports `aa_policy::*`

`PolicyContext` moved with the evaluator that calls it (it names only `aa_core` types); `ProductionPolicyContext` stayed with the `AgentRegistry` and `BudgetTracker` it reads. That boundary is what makes the leaf crate possible.

Merging as a merge commit. Admin only for the solo-maintainer approving-review requirement — no failing or pending check is bypassed.

PR #1900, opened 2026-08-02
[AAASM-5349] ♻️ (aa-policy): Move the launch resolver in beside the semantics it uses

Chisanan232:
## Description

Second of the **AAASM-5349** PRs. #1896 moved the policy *semantics* into the `aa-policy` leaf crate; this moves the **launch resolver** in beside them, so one implementation can serve every surface that has to answer "which of the four policy states did this launch resolve?".

It changes no behaviour.

## Why it has to move

`run_policy` lived in `aa-cli`. AAASM-5349 needs the same answer from the developer-integration service in `aa-runtime`, because ADR 0030 **forbidden design 10** makes `status` and `verify` render what the *service* computed rather than deriving a state client-side — `aa-cli/src/commands/integrations/status.rs` states the rule outright.

A resolver reachable only from the CLI cannot serve that. A second copy inside the service would be a second definition of what "governed" means, which is the defect this ticket exists to remove rather than one to introduce.

So it moves to `aa-policy`, beside the validator it already calls, where `aa-cli`, `aa-runtime` and `aa-gateway` can all reach one implementation. `aa-cli` imports it as `run_policy`, so its call sites read unchanged.

## Type of Change

- [x] ♻️ Refactor (no functional change)

## Breaking Changes

- [ ] No. Same four states, same refusal text, same search order, same tests. `aa_cli::commands::run_policy` is gone as a path; it was crate-internal apart from one integration test, updated here.

## Related Issues

- Jira ticket: AAASM-5349 (PR 2 of 4)
- Follows: #1896 (`ffeb2955` gate fix, `2d5dff58` extraction)
- Precedes: DI-API v3 carrying policy posture on `StatusView`/`VerificationView`; then the audit variant

## Testing

**Test-count parity, the check a move has to pass:** `main` has **16** `#[test]` in `aa-cli/src/commands/run_policy.rs`; this branch has **16** in `aa-policy/src/resolve.rs`. A move that silently dropped a `mod tests` would still compile and still pass everything remaining.

- **`aa-policy` alone: 315/315** — 299 from #1896 plus the 16 relocated
- **`aa-cli`: 946/946**
- `cargo clippy -p aa-policy -p aa-cli --all-targets -- -D warnings` passes

### Two things that only surfaced building the crate by itself

Both would have passed unnoticed in a workspace-wide run, which is exactly the trap that produced the `tokio/time` defect on #1896:

1. **`resolve.rs`'s tests called `crate::test_support::env_guard()`** — a lock that lives in `aa-cli`. `aa-policy` now has its own. A lock is per test binary, so the two crates' tests can never contend and sharing one would be meaningless.
2. **`aa-cli/tests/run_policy_fail_closed.rs` read `ALLOW_ALL_TEMPLATE` from the old path.** It now reads it from `aa_policy::resolve`, which is where the authority moved to — the test pointing at the shared crate is the point of the change, not an incidental fix.

## Checklist

- [x] Documentation updated if behaviour changed (module docs state why the resolver sits where it does)
- [ ] All CI checks passing

## One thing deliberately not done

`into_enforceable()` returns `anyhow::Result`, so `aa-policy` gains an `anyhow` dependency. A typed error per refusal is the better shape for a library crate whose consumers now include a service, but changing it would alter the API and every caller — that belongs in its own change, not in relocating code unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,6 +99,7 @@ dependencies = [
  "aa-core",
  "aa-devtool",
  "aa-gateway",
+ "aa-policy",
  "aa-proto",
  "aa-proxy",
  "aa-runtime",
@@ -451,6 +452,7 @@ version = "0.0.1-rc.6"
 dependencies = [
  "aa-core",
  "aa-security",
+ "anyhow",
  "async-trait",
  "chrono",
  "chrono-tz",
@@ -523,6 +525,7 @@ dependencies = [
  "aa-devtool-claude-code",
  "aa-ebpf",
  "aa-ebpf-common",
+ "aa-policy",
  "aa-proto",
  "aa-security",
  "aa-storage-sqlite-buffer",
--- a/aa-cli/Cargo.toml
+++ b/aa-cli/Cargo.toml
@@ -46,6 +46,7 @@ aa-security          = { path = "../aa-security", version = "0.0.1-rc.6", featur
 aa-devtool           = { path = "../aa-devtool", version = "0.0.1-rc.6" }
 # strip-for-publish:end devtool
 aa-gateway           = { path = "../aa-gateway", version = "0.0.1-rc.6" }
+aa-policy            = { path = "../aa-policy", version = "0.0.1-rc.6" }
 # AAASM-5280: `aasm integrations` is a Developer Integration API client. The
 # reference client, the socket convention and the local enrolment file all live
 # in aa-runtime, and reimplementing any of them here would be the second
--- a/aa-cli/src/commands/mod.rs
+++ b/aa-cli/src/commands/mod.rs
@@ -30,7 +30,6 @@ pub mod policy;
 pub mod proxy;
 // strip-for-publish:begin devtool
 pub mod run;
-pub mod run_policy;
 pub mod run_registration;
 // strip-for-publish:end devtool
 pub mod sandbox;
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -12,11 +12,13 @@ use tokio::signal::unix::SignalKind;
 
 use aa_core::{DevToolAdapter, DevToolInfo, DevToolKind, GovernanceLevel};
 
-use crate::commands::run_policy;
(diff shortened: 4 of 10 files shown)


Chisanan232:

Pure relocation, so the review question is whether anything was lost or quietly changed.

**Test-count parity holds exactly.** `main` carries **16** `#[test]` in `aa-cli/src/commands/run_policy.rs`; this branch carries **16** in `aa-policy/src/resolve.rs`. That is the check a move has to pass — dropping a `mod tests` still compiles and still passes everything that remains.

- **`aa-policy` alone: 315/315** — the 299 from #1896 plus these 16
- **`aa-cli`: 946/946**
- `clippy -p aa-policy -p aa-cli --all-targets -- -D warnings` clean
- `Packaged-artifact gate → pass`, `Published-surface coherence gate → pass`, 51 checks pass / 0 fail / 0 pending

**Two defects surfaced only by building the crate alone**, and both would have passed unnoticed in a workspace-wide run — the same trap that hid the missing `tokio/time` on #1896:

1. `resolve.rs`'s tests called `crate::test_support::env_guard()`, a lock living in `aa-cli`. `aa-policy` now has its own. A lock is per test binary, so these two crates' tests can never contend and sharing one would have been meaningless anyway.
2. `aa-cli/tests/run_policy_fail_closed.rs` read `ALLOW_ALL_TEMPLATE` from the old path. It now reads it from `aa_policy::resolve` — the test pointing at the shared authority is the point of the change.

**Why the move is required rather than tidy.** ADR 0030 forbidden design 10 makes `status` and `verify` render what the *service* computed; `aa-cli/src/commands/integrations/status.rs` states it outright. A resolver reachable only from the CLI cannot serve that, and a second copy inside the service would be a second definition of "governed" — the exact defect AAASM-5349 exists to remove.

**Carried deliberately:** `into_enforceable()` returns `anyhow::Result`, so `aa-policy` gains `anyhow`. A typed error per refusal is the better shape for a library crate whose consumers now include a service, but that changes the API and every caller; it belongs in its own change, not in relocating code unchanged.

PR #1903, opened 2026-08-02
[AAASM-5349] ✨ (aa-cli): Record a governed launch in the audit trail, and state what is missing

Chisanan232:
## Description

Last of the **AAASM-5349** PRs: `aasm run` records its governed launch in the gateway audit trail. It implements the authenticated half **and states plainly what it cannot record**, rather than describing a two-state surface as a four-state one.

## What is recorded

A `PROCESS_EXEC` event — the launch genuinely is a subprocess exec — carrying the resolved policy on labels:

```
AuditEvent {
  action_type: PROCESS_EXEC
  decision:    ALLOW
  detail: ProcessExecDetail { command: "claude", args: [...] }
  labels: {
    "aasm.launch":        "governed"     ← separates this from any other exec
    "aasm.policy_state":  "enforced"     ← same token as AA_POLICY_STATE
    "aasm.policy_source": "/…/policy.yaml"
  }
}
```

**No proto change.** `PROCESS_EXEC` and the `labels` map already exist, so the SDK-facing audit contract does not move, nothing is version-bumped, and the three SDKs are untouched.

The tokens come from `PolicyResolution::posture()`, which this PR also makes the single shared mapping — receipt, `status`, `verify` and audit now read **one** definition of the four states instead of four descriptions of them. `resolve_host_policy` in the devint service becomes a one-liner over it.

## What is not recorded, and why

| Policy state | Launch | In the trail |
|---|---|---|
| `enforced` | proceeds | ✅ |
| `permissive` | proceeds | ✅ |
| `unconfigured` | **refused** | ❌ |
| `load_failed` | **refused** | ❌ |

Two deliberate decisions meet, and each is right on its own terms:

1. **`aasm run` refuses before it registers** — `run.rs` calls `into_enforceable()` before `register_with_gateway`, because *"a registered session that never launched is a governed identity with no process behind it"*.
2. **The audit service is fail-closed on a registered principal** — `AuditServiceServer` is mounted with the `auth` interceptor, and `resolve_caller` resolves callers **only** via `registry.find_by_credential_token`. An API key does not authenticate there.

So a refusal has no principal and no authenticated path to the trail. The two states that cannot be recorded are the two that **refuse** — the security-relevant ones.

Recording only the permitting states while calling the surface complete would be the exact over-claim this ticket exists to remove. Instead: the module documents it at the point of use, `is_recordable()` asserts the limit **as a test**, and **AAASM-5435** owns designing a secure pre-registration refusal audit. If the ordering invariant ever changes so a refusal can register, that test fails and the documentation is revisited with it rather than quietly going stale.

### Three shortcuts rejected, not overlooked

- **Register purely to emit a record, then deregister** — manufactures the identity-without-a-process that decision 1 exists to prevent, and pollutes the registry with sessions that never ran.
- **A local file or socket side channel** — a second, unauthenticated write path into the governance record. ADR 0030 separately forbids audit traffic on the DI-API socket.
- **Relax the audit API to accept a pre-registration principal** — an authentication weakening on a governance surface, which needs AAASM-5435's threat analysis, not an expedient patch.

## Type of Change

- [x] ✨ New feature
- [x] 📝 Documentation (the partial-surface statement is part of the deliverable)

## Breaking Changes

- [ ] No. No proto change, no version bump, no SDK impact. The emission is best-effort: a launch that governed correctly is never aborted because the audit upload failed — the failure is printed, because a record that silently did not happen is worse than one that visibly did not.

## Related Issues

- Jira: AAASM-5349 (PR 4 of 4) — **does not close it**; AC 6 needs the revision proposed on the ticket
- Follow-up: **AAASM-5435** — secure pre-registration refusal audit
- Follows #1896, #1900, #1902

## Testing

`aa-policy` + `aa-cli` **1271/1271**. `clippy --all-targets -- -D warnings` clean. `aa-policy`, `aa-runtime` and `aa-cli` each compile **alone**.

Five tests on the new module, aimed at the claims rather than the mechanics:

| Test | Property |
|---|---|
| `the_state_label_is_the_shared_token` | an audit query and `AA_POLICY_STATE` select the same sessions without translation |
| `a_governed_launch_is_marked_as_one` | a governed launch is separable from any other `PROCESS_EXEC` |
| `an_absent_source_contributes_no_label` | absence of an artifact ≠ an artifact with a blank path |
| `only_the_two_permitting_states_are_recordable` | asserts the **documented limit**, so the docs cannot outlive the invariant |
| `an_unknown_posture_contributes_no_state_label` | an unresolved posture never acquires an invented token |

## Publish-strip correctness

`run_audit.rs` consumes `tonic`, which the publish strip removes, so it joins `run.rs` and `run_registration.rs` in the strip's `DELETED_FILES`, its module declaration sits in the existing `devtool` region, and its dependency sits inside the `sdkclient` region. A module left behind importing a dependency the published manifest no longer declares is the AAASM-5309 failure shape exactly.

Because this touches `aa-cli/src/commands/mod.rs` and `aa-cli/Cargo.toml`, the packaged-artifact gate **runs** on this PR rather than being path-filtered out.

## Checklist

- [x] Documentation updated (module docs carry the limitation where it is used)
- [ ] All CI checks passing

### DIFF
--- a/.ci/strip-for-publish.sh
+++ b/.ci/strip-for-publish.sh
@@ -89,6 +89,11 @@ DELETED_FILES=(
     # (publish = false), stripped from aa-cli/Cargo.toml by the `sdkclient`
     # region, so the module that imports it has to go with the dep.
     "${REPO_ROOT}/aa-cli/src/commands/run_registration.rs"
+    # AAASM-5349: the audit record a governed launch emits. Consumes `tonic`
+    # (stripped by the `sdkclient` region) and only ever runs on the `aasm run`
+    # path, so it goes with the two files above rather than being left behind
+    # importing a dependency the published manifest no longer declares.
+    "${REPO_ROOT}/aa-cli/src/commands/run_audit.rs"
     "${REPO_ROOT}/aa-cli/src/commands/tools.rs"
     "${REPO_ROOT}/aa-cli/tests/run_command.rs"
     "${REPO_ROOT}/aa-integration-tests/tests/cli_run.rs"
--- a/aa-cli/Cargo.toml
+++ b/aa-cli/Cargo.toml
@@ -68,6 +68,12 @@ aa-proxy             = { path = "../aa-proxy", version = "0.0.1-rc.6" }
 # `.ci/strip-for-publish.sh` along with `run.rs` and `run_registration.rs`, its
 # only consumers.
 aa-sdk-client        = { path = "../aa-sdk-client", version = "0.0.1-rc.6" }
+# AAASM-5349: `aasm run` reports its governed launch to the gateway's
+# `AuditService` over the same gRPC transport, authenticated with the credential
+# token registration returned. In this region because `run_audit.rs` is stripped
+# with the rest of the `aasm run` surface — a dependency left behind with no
+# consumer would enlarge the published crate for nothing.
+tonic                = { workspace = true }
 # strip-for-publish:end sdkclient
 aa-sandbox           = { path = "../aa-sandbox", version = "0.0.1-rc.6" }
 aa-storage           = { path = "../aa-storage", version = "0.0.1-rc.6" }
--- a/aa-cli/src/commands/mod.rs
+++ b/aa-cli/src/commands/mod.rs
@@ -30,6 +30,7 @@ pub mod policy;
 pub mod proxy;
 // strip-for-publish:begin devtool
 pub mod run;
+pub mod run_audit;
 pub mod run_registration;
 // strip-for-publish:end devtool
 pub mod sandbox;
--- a/aa-cli/src/commands/run.rs
+++ b/aa-cli/src/commands/run.rs
@@ -678,6 +678,21 @@ pub async fn execute_with_adapters(args: &RunArgs, adapters: &HashMap<&str, Box<
     // no identity is an ungoverned process wearing a governed launch's name.
     let registration = register_with_gateway(&info, args).await?;
     let handle = RegistrationHandle::of(&registration);
+
+    // Recorded now, not at exit: an audit trail that only learns about a session
+    // when it ends loses every session still running and every one that does not
+    // end cleanly. Reachable only for a policy that permitted the launch — the
+    // two refusing states are refused above, before any registration exists, so
+    // they cannot reach the trail at all (see `run_audit`'s module docs).
+    run_registration::report_launch(
+        &registration,
+        &handle.trace_id,
+        &handle.session_id,
+        &args.tool,
+        &args.tool_args,
+        &resolution.posture(),
+    )
+    .await;
     let mut child_env = build_child_env(&handle, proxy.as_deref(), args.no_proxy, mode);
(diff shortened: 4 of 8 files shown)


Chisanan232:

The deliverable here is partly a **refusal to claim something**, so that is what I checked hardest.

### The claim boundary is enforced, not just described

| Policy state | Launch | In the trail |
|---|---|---|
| `enforced` | proceeds | ✅ |
| `permissive` | proceeds | ✅ |
| `unconfigured` | **refused** | ❌ |
| `load_failed` | **refused** | ❌ |

Two deliberate decisions produce this, each correct alone: `aasm run` refuses **before** it registers (so a refused session never becomes a governed identity with no process behind it), and the audit service is fail-closed on a registered principal (`resolve_caller` resolves only via `find_by_credential_token`). A refusal therefore has no principal and no authenticated path.

The gap falls in the worst direction — the two unrecordable states are the two that **refuse** — so a trail logging every permitted launch and no refusal is exactly the shape that looks complete and is not. Three things keep it honest rather than hidden:

- the limitation is documented **at the point of use**, not in a changelog;
- `only_the_two_permitting_states_are_recordable` asserts it **as a test**, so if the ordering invariant ever changes the test fails and the docs are revisited with it;
- **AAASM-5435** owns the real fix with the full threat model.

Three shortcuts are rejected explicitly in both the code and the follow-up, so a later reader sees they were considered and refused rather than missed: fake registration, a local side channel, and relaxing the audit API to accept a pre-registration principal.

### One mapping, not four descriptions

`PolicyResolution::posture()` now lives in `aa-policy` and is the single source for receipt, `status`, `verify` **and** audit. `resolve_host_policy` collapses to a one-liner over it. Four surfaces reporting four independently-written descriptions of the same four states was the drift risk this removes.

### Publish-strip correctness

`run_audit.rs` consumes `tonic`, which the strip removes, so it joins `run.rs` and `run_registration.rs` in `DELETED_FILES`, its declaration sits in the existing `devtool` region, and its dependency sits in the `sdkclient` region. A module left behind importing a dependency the published manifest no longer declares is the AAASM-5309 shape exactly — and because this touches `mod.rs` and `Cargo.toml`, the **`Packaged-artifact gate` ran and passed**, as did `Published-surface coherence`, which is the check that would catch a dangling post-strip reference.

### Verification

46 checks pass / 0 fail / 0 pending. Locally `aa-policy` + `aa-cli` **1271/1271**; clippy `--all-targets -- -D warnings` clean; `aa-policy`, `aa-runtime`, `aa-cli` each compile **alone**. No proto change, no version bump, no SDK impact.

Emission is best-effort by design: a launch that governed correctly is never aborted because the upload failed, and the failure prints rather than being swallowed — a record that silently did not happen is worse than one that visibly did not.

### What this does not do

**It does not close AAASM-5349 and does not mark AC 6 complete.** The AC 6 revision splitting 6a (receipt/`status`/`verify`, all four states) from 6b (audit, two states, gap owned by AAASM-5435) is posted on the ticket awaiting the owner's decision.

PR #1921, opened 2026-08-03
[AAASM-5465] ✅ (ci): Make skipped Claude Code coverage visible instead of green

Chisanan232:
## Description

**AAASM-5465.** Live-Claude tests that could not run reported as **PASS** while asserting nothing. A green lane was indistinguishable from a lane that measured nothing.

## The inventory corrected the ticket

The ticket (and my own earlier summary) claimed *every* live-Claude suite was affected. **That was wrong**, and the correction matters because it changes how much of the existing evidence was ever in doubt:

| Suite | Actually guarded |
|---|---|
| `aa-cli/tests/integrations_claude_code.rs` | **5 of 5** — the whole file |
| `conformance_claude_code.rs` | **1 of 26** (the rest is hermetic and runs on Linux) |
| `spike_claude_code_lifecycle.rs` | **2 of 15** |
| `cli_run_claude_governed_launch.rs` | **1 of 5** |
| `claude_code_integration_lifecycle.rs` | **0** — fully hermetic, not guarded at all |

So the conformance evidence was largely real. The genuinely blind spot was the **aa-cli native suite**, which is 100% gated.

## A second defect found on the way

`spike_support::install()` — on the path of **every** scenario in that file — called the gating lookup merely to print a decorative version string. All 15 scenarios therefore emitted an unearned `SKIP:`, including the 13 that were about to run for real. Split into a non-declaring `locate_claude()`.

## What changed

**Three outcomes separated** (req 4). The single `Skipped` token became `UnsupportedPlatform` (a) vs `ToolAbsent` (b), alongside `Measured` (c) and `NotMeasured`. `require_macos` → (a), `require_claude` → (b). Records carry the platform.

**CI reporting** (req 2). `.ci/test-evidence-summary.sh` nets ledger declines against nextest's JUnit pass count and writes ran / passed / **substantive** / (a) / (b) / not-measured to `$GITHUB_STEP_SUMMARY` with a `::notice`/`::warning`/`::error` annotation. Taking the denominator from the runner means a newly-added test counts as substantive **by default** — the safe direction.

**A lane that measured nothing no longer passes** (req 3). `--require-evidence` exits 1 when `substantive == 0`, enabled on both required conformance lanes and both real-tool runs. Without it, the condition still prints **NO EVIDENCE PRODUCED**. A missing/unparseable JUnit report is a hard failure, and an *empty ledger* is called out separately — it could mean "the ledger was never enabled", which would silently restore the raw pass count.

**Skips no longer masquerade as evidence** (req 5). The two spike real-binary scenarios that cleared their opt-outs and then observed zero traffic now record `not_measured` **and panic**, matching what the conformance suite already did.

## Windows: absence made failing-visible, not implemented

`managed_settings_platform_integrity.rs` asserts on Unix that the model is present and enforcing (ownership + permission refusals through the public installer with `FakeAuthority` against a temp target — no prompt reachable). **On any other platform it fails**, naming the three properties with no implementation.

A Windows equivalent was deliberately **not** written: `MANAGED_SETTINGS_PATH` is a hardcoded macOS path, Windows has no uid, and "not writable by anyone but an administrator" is a DACL question needing a Win32 security-descriptor dependency and a decision about which SIDs are authoritative. That is expanding product semantics on a platform with **no CI lane at all** — there is no `windows-latest` anywhere in `.github/workflows/`. This cannot turn anything red today; the point is that adding that lane now reports the gap instead of going green having measured no integrity whatsoever.

## Type of Change

- [x] ✅ Tests · [x] 🔧 Configuration (CI)

## Breaking Changes

- [x] No product code changed. Test/CI only.

## Related Issues

- Jira: **AAASM-5465** · Epic: **AAASM-5453** · Same family as **AAASM-5375**, **AAASM-5444**

## Testing

**Gate proved in both directions**, which is the whole point:

* Fully-skipped lane (aa-cli suite, `claude` off `PATH`): runner reports `5 tests run: 5 passed`; the summary reports **substantive 0**, `tool_absent 5`, lists all five, and **exits 1** with `--require-evidence`.
* Required lane simulated on Linux (`AA_SPIKE_CLAUDE_BIN=/nonexistent`): 26 ran, **25 substantive**, 1 `tool_absent` → passes. So the gate is green today and only fires on genuine emptiness.

Suites run: `conformance_claude_code` 26/26 **twice** (tool present and absent), spike + integration-lifecycle + trusted-proxy **37/37**, aa-cli suite 5/5, new platform-integrity test 1/1. `fmt`, `clippy --all-targets --all-features -- -D warnings`, and `cargo doc` all clean.

## Two caveats stated rather than buried

1. **`ci.yml`'s `cargo nextest run --workspace` still counts the five aa-cli skips as passes.** Every skip is now printed and recorded there, but no evidence gate was added to that job — it is a large shared job outside the conformance workflow, and gating it is a CI-policy call left for the owner. The new non-required `native-cli` lane reports the truth about the same suite.
2. `cli_run_claude_launch_env.rs` and `cli_run_claude_governed_launch.rs` were **not run to completion locally** (each stands up a `TrustedProxy` that rebuilds `aa-proxy` per case). Their diffs are a `mod` declaration plus edits confined to `#[cfg(not(unix))]` bodies — inert on the Unix path — and both compile under `clippy --all-targets --all-features`.

## Checklist

- [ ] All CI checks passing


Chisanan232:
Superseded by #1950.

This PR's content was correct but its **commit order was not bisectable** — `Every commit in the range builds` failed with `error[E0432]: unresolved import \`crate::evidence\``. The first commit moved `outcome.rs` into the shared `evidence` module while the `mod evidence;` declarations landed in later commits, so the intermediate state did not compile. A module move is not atomic while its consumers still name the old path.

#1950 carries **byte-identical content** with the move and all its consumers in one commit, and all 8 commits verified to build individually before pushing.

Rebuilt on a fresh branch rather than corrected in place, because reordering pushed history would have required a force-push, which is not authorised here.

